PR #60, opened 2014-05-09
Initial iOS deployment docs

henrikhodne:
### WIP: do not merge right now, still has some TODOs

These are not quite finished. The largest TODO is the block missing on how to encrypt files. I was planning on just using this:

```
openssl aes-256-cbc -k a-passphrase -in infile -out outfile -a
```

However, I couldn't figure out how secure that was. The files that would probably need to be encrypted are `.travis/profile/$PROFILE_NAME.mobileprovision`, `.travis/certs/dist.cer` and `.travis/certs/dist.p12`. Not completely sure whether the first two need to be encrypted (what can they be used for?) and not completely sure if the last one (private key) is encrypted enough with the encryption that's done when exporting it from the keychain.

The entire doc could probably use some editing here and there too.

### DIFF
--- a/_layouts/en.html
+++ b/_layouts/en.html
@@ -155,6 +155,11 @@
               <li><a href="/user/deployment/#Other-Providers">Other Providers</a></li>
             </ul>
 
+            <h2>OS X &amp; iOS</h2>
+            <ul>
+              <li><a href="/osx-ios/ios-deployment/">iOS deployment</a></li>
+            </ul>
+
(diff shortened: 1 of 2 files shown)


plaindocs:
@henrikhodne is this still relevant?

MariadeAnton:
@henrikhodne I see, thanks for explaining - I'll wait for @cotsog thoughts on this to see if we could come up with a brand new thing including relevant information for this.

PR #115, opened 2014-09-07
Update tested Rubies

guilhermesimoes:
Changes include:

* Removing 1.8.7 and 1.9.2 from the default config since they were end-of-life'd.

* Prioritizing 2.2 over 1.9.3.

* Changing jruby-head to jruby, as per travis-ci/travis-ci/issues/3067.

* Specifying Ruby 2.x instead of 2.x.x.

  Since Ruby 2.1.0, [the MRI versioning policy changed to follow Semantic Versioning][1].

  Thus, it is better to leave the patch version unspecified as this way the latest patch release will be selected automatically.

  This probably deserves to be mentioned in the Ruby documentation but for now I think these changes will do.

[1]:
https://www.ruby-lang.org/en/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/

### DIFF
--- a/user/languages/ruby.md
+++ b/user/languages/ruby.md
@@ -39,19 +39,16 @@ To specify them, use `rvm:` key in your `.travis.yml` file, for example:
 
     language: ruby
     rvm:
-      - 1.9.3
-      - jruby-18mode # JRuby in 1.8 mode
-      - jruby-19mode # JRuby in 1.9 mode
-      - rbx-2.1.1
-      - 1.8.7
-      - ree
+      - 2.2
+      - jruby
+      - rbx-2
 
 > Note that the `rvm:` key is only available in Ruby Build Environments, not in other
 images containing a ruby implementation.
 
-As we upgrade both RVM and Rubies, aliases like `1.9.3` or `jruby` point to different exact versions, patch levels and so on.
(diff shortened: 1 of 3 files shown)


guilhermesimoes:
@BanzaiMan Is there any blocker for this getting merged? Even Ruby 1.9.3 has been EOL'd at this point :(

plaindocs:
I can't see any reason not to merge this in. @MariadeAnton any objections?

user/languages/ruby.md
```diff
@@ -39,19 +39,16 @@ To specify them, use `rvm:` key in your `.travis.yml` file, for example:
...
 images containing a ruby implementation.
 
-As we upgrade both RVM and Rubies, aliases like `1.9.3` or `jruby` point to different exact versions, patch levels and so on.
+As we upgrade both RVM and Rubies, aliases like `2.2` or `jruby` point to different exact versions, patch levels and so on.
 
-For full up-to-date list of provided Rubies, see our [CI
+For a full up-to-date list of provided Rubies, see our [CI
 environment guide](/user/ci-environment/).
 
 If you don't specify a version, Travis CI will use MRI 1.9.3 as the default.
```
guilhermesimoes: Is this still true? Travis uses 1.9.3 by default?

BanzaiMan: Yes.

PR #129, opened 2014-10-15
Note ambiguous message multiple branch deploy

behemphi:
travis-ci/travis-ci#2824

The message that appeears indicating the complementary branch(es) are not to be released is ambiguous enough to cause the user to think the deploy did not happen on the intended branch.  Made a note of this behavior.


gsf:
I am fully in support of adding the branch name to the message. It's really confounding the first time you see it and a headache to explain to each member of my team.

PR #144, opened 2014-11-11
Add a page about a git options (depth, submodules, strategy)

drogus:


plaindocs:
Shall I look at this as something we want to add?

MariadeAnton:
@plaindocs definitely! Do you need any more info about this? We could add a note about setting a smaller depth too.

joshk:
For informations sake, the `git strategy` section can be removed.

PR #170, opened 2014-12-26
add docs on pushover

jantman:
This is a companion to https://github.com/travis-ci/travis-tasks/pull/28


jantman:
Oops. That ``push origin --delete pushover`` was *definitely* on the wrong clone...

PR #176, opened 2015-01-05
Fixing #175

mnapoli:
I totally messed up in #175, this change fixes the docs.

Sorry, here is a cat :cat2: 

ping @joshk


joshk:
:dog2: :dog2: :dog2:

PR #178, opened 2015-01-07
update instructions for getting api key

linclark:
When you generate a new .npmrc, the _auth key no longer exists. Instead there is a _password key. The _auth property had contained the base64 encoded string composed of username:password. The _password property only contains the base64 encoded password, without the username.

This updates the instructions in a way that works for both versions.

### DIFF
--- a/user/deployment/npm.md
+++ b/user/deployment/npm.md
@@ -25,8 +25,11 @@ commits, like so:
 
 If you tag a commit locally, remember to run `git push --tags` to ensure that your tags are uploaded to Github.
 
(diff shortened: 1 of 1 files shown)


joshk:
Do the updated instructions work for users who are still using NPM 1.x?

linclark:
Yes, it is independent of npm version.

joshk:
:heart: :heart: :heart:

joshk:
Thank you very much for this!

linclark:
and thank you for the quick merge :grinning:

PR #188, opened 2015-02-02
Add docs for the apt_packages addon

meatballhat:


user/apt-packages.md
```diff
@@ -0,0 +1,19 @@
+---
+title: APT Package Installation
+layout: en
+permalink: /user/apt-packages/
+---
+<div id="toc">
+</div>
+
+When using the [container based infrastructure](/user/workers/container-based-infrastructure/), there is a limitation on
+the use of `sudo` without a password, as is required to install packages with `apt-get`.  In order to have Travis CI do
```
BanzaiMan: "there is a limitation on the use of `sudo` without a password…" doesn't read well. How about something like:

> `sudo` is disabled in user-defined build phases such as `before_install`. This prohibits installation of APT packages via `apt-get`.

user/apt-packages.md
```diff
@@ -0,0 +1,19 @@
...
+this on your behalf prior to `sudo` privileges being disabled, use the `addons.apt_packages` key, e.g.:
```
BanzaiMan: "prior to `sudo` privileges being disabled" → "before `sudo` privilege is removed`

user/apt-packages.md
meatballhat: :+1: love it

PR #192, opened 2015-02-06
Fix broken link

b4hand:
Fixed a broken link on this page:

http://docs.travis-ci.com/user/private-dependencies/#Deploy-Key

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ In order to run a local Web server that will serve documentation site, run:
 
     ./bin/jekyll serve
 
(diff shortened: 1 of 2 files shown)


joshk:
Thanks for the PR

I would personally recommend using the 'settings' panel to add your SSH key. I'll talk to the team about this as we may change the documentation completely.

MariadeAnton:
Thank you so much for the PR, @b4hand!

I was just catching up on this and I think the page is now pointing to https://docs.travis-ci.com/user/encrypting-files correctly. Thank you again for spotting it out!

PR #193, opened 2015-02-11
Deploy S3 - HTTP cache headers

Halama:
I've tried to figure out if it is possible to set `Cache-Control` for S3 deploy, finally I've found that it is supported by deploy tool https://github.com/travis-ci/dpl#s3 but not documented.


BanzaiMan:
Thank you!

PR #204, opened 2015-03-02
you -> your

bbodenmiller:
From:

> If you are using a workspace instead of a project, use the `xcode_workspace` key in you .travis.yml instead of `xcode_project`.

to

> If you are using a workspace instead of a project, use the `xcode_workspace` key in **your** .travis.yml instead of `xcode_project`.


joshk:
Thank you very much!

PR #208, opened 2015-03-10
Ruby 2.0.c  ->  2.0.0

amatsuda:
Maybe you meant 2.0.x?
But actually we're not gonna make any other 2.0 release than 2.0.0, so this doesn't have to be 2.0.x.
Let it just be 2.0.0 here.

### DIFF
--- a/user/languages/ruby.md
+++ b/user/languages/ruby.md
@@ -17,7 +17,7 @@ configuration](/user/build-configuration/) guides first.
 
 The Ruby VM provides recent versions of:
 
-- Ruby 2.2.x, 2.1.x, 2.0.c, 1.9.3, 1.9.2 and 1.8.7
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Indeed, it should've been `2.0.x`. But we'll accept `2.0.0` and be done with it.

Thanks for your eagle eyes! <3

amatsuda:
Actually I was reading this document because I realised that giving 
`    - rvm: 2.2`
picks version 2.2.0, not 2.2.1, where it supposed to work as "2.1".
Should I better raise another issue, or could you look at it?

BanzaiMan:
Indeed, `2.2` is currently an alias to `2.2.0`. This will change with the build environment update next month.

amatsuda:
Understood. I'll do `2.2.1` then. Thanks for clarifying this!

PR #210, opened 2015-03-13
C#: docs about Mono version switching and a few minor fixes

akoeplinger:
Mainly about https://github.com/travis-ci/travis-build/pull/394

If you have any other improvements to the docs in mind, we should get them done as part of this PR :)

/cc @Joshua-Anderson @nterry

### DIFF
--- a/user/languages/csharp.md
+++ b/user/languages/csharp.md
@@ -30,12 +30,12 @@ language: csharp
 solution: solution-name.sln
 {% endhighlight %}
 
-When the `solution` key is present, Travis will run NuGet package restore and build the given solution. You can also specify your own scripts, as shown in the next section.
+When the optional `solution` key is present, Travis will run NuGet package restore and build the given solution. You can also specify your own scripts, as shown in the next section.
 
 ### Script
 
 By default Travis will run `xbuild /p:Configuration=Release solution-name.sln`. Xbuild is a build tool designed to be an implementation for Microsoft's MSBuild (the tool that Visual Studio uses to build your projects).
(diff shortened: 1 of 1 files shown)


user/languages/csharp.md
```diff
@@ -59,6 +59,29 @@ install:
...
+{% highlight yaml %}
+language: csharp
+mono:
+  - latest
+  - 3.12.0
+  - 3.10.0
+...
+{% endhighlight %}
+
+You can choose from the versions available from Mono's [snapshot repositories](http://www.mono-project.com/docs/getting-started/install/linux/#accessing-older-releases), i.e. 3.8.0 and later. To make it easier for you to test against the versions that ship with Ubuntu 12.04 and 14.04, we also support `2.10.8` and `3.2.8`. Other earlier versions are not supported. Note that those special cases also don't provide NuGet and PCL reference assemblies.
```
nterry: "... i.e. 3.8.0 and later." should be fixed to "... i.e. 3.8.0 and *earlier*

akoeplinger: No, the snapshot repos have 3.8 and later. How can we make this more clear?

Joshua-Anderson:
@akoeplinger It looks good but I would add an section documenting all versions of mono we are supporting.

akoeplinger:
@Joshua-Anderson I added a table that should hopefully make the versions more clear. I don't want to explicitly list versions after 3.10 since those should "just work" without us having to update the table each time a new Mono is released.

Joshua-Anderson:
Thanks! LGTM

PR #213, opened 2015-03-20
Update notifications.md

brixen:
According to travis-lint, on_start should be always, never, or change:

https://github.com/travis-ci/travis-yaml/blob/44e072dd42a69dc11a9a6cd0f57899b6b4c95d5d/SPEC.md#notificationswebhookson_start

I'm not certain that the default is always, but I assumed it probably is based on the previous two lines.


joshk:
@BanzaiMan can you please confirm the default then merge?

Thanks for the awesome PR @brixen !!!

PR #228, opened 2015-04-14
Formally document process to add a new community-supported language

BanzaiMan:
To formally document what we have been doing so far when a new community language is added.


BanzaiMan:
Paging @joshk

BanzaiMan:
Paging @joshk again.

joshk:
Looks great to me
This definitely shows some room for improvement on our end, primary with how we can make it simpler. 
:shipit:

PR #230, opened 2015-04-16
add hack for building all branches

afeld:
The use case is Pages-only repositories, where the maintainers run the build with Travis, and want all branches and pull requests to be run.

### DIFF
--- a/user/build-configuration.md
+++ b/user/build-configuration.md
@@ -343,7 +343,7 @@ You can either white- or blacklist branches that you want to be built:
         - legacy
         - experimental
 
-    # whitelist
+    # safelist
     branches:
       only:
         - master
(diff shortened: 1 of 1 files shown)


user/build-configuration.md
```diff
@@ -351,7 +351,11 @@ You can either white- or blacklist branches that you want to be built:
 
 If you specify both, "except" will be ignored. Please note that currently (for historical reasons), `.travis.yml` needs to be present *on all active branches* of your project.
 
-Note that the `gh-pages` branch will not be built unless you add it to the whitelist (`branches.only`).
+Note that the `gh-pages` branch will not be built unless you add it to the whitelist (`branches.only`), or force all branches to build:
```
MariadeAnton: Please use _safelist_ instead of _whitelist_ :)

MariadeAnton:
This is super helpful, thank you Aidan!

Do you think you could update the PR accordingly? 

I know it's taken us a while to reply, so I could also cherry-pick your commit and add the proposed change to it, just let us know what would be easier for you!

afeld:
Updated! _However_, I think this behavior may have changed, where `gh-pages` needs to be included explicitly. I don't have time to double-check this right now...would someone be able to verify on your end?

MariadeAnton:
@afeld thanks for that! You're so very much right. What would work for people expecting to build all branches (including `gh-pages`) is:

```yaml
branches:
  only:
    - gh-pages
    - /.*/
```

I personally think it's worth mentioning, what do you think?

MariadeAnton:
awesome, thanks for the updates Aidan! 

I think it's ready to be updated with the `gh-pages` branch so that we can merge it here :)

PR #268, opened 2015-06-19
sudo is required

mfrasca:
R support in Travis CI currently requires `sudo`, so let's help first-time-users with a more complete basic configuration file.


BanzaiMan:
Thank you!

PR #270, opened 2015-06-20
Fix typo

edunham:
If this output was copy/pasted directly from the output of a tool, it might be worth fixing it in that tool's source as well. But I grepped for `scirpt` in the ruby CLI source, I didn't find anything.


BanzaiMan:
Thank you!

PR #271, opened 2015-06-23
Gh pages fix

JohnWong:
Fix style of multi-level list. Indent the 2nd level list.
Fix issue of "settings-env-vars.png" do not shown in figure tag.


BanzaiMan:
Thank you!

PR #284, opened 2015-07-06
Use https:// (not git://) URLs

jdennes:
GitHub doesn't advertise `git://` URLs anywhere, and we don't encourage using them.

The preferred protocol for "Public URLs" is HTTPS.

Thanks! :v:


BanzaiMan:
Thank you!

PR #289, opened 2015-07-11
fix: migrating-from-legacy use of compound bash in yaml

grooverdan:
https://travis-ci.org/openquery/mariadb-server/jobs/70511922

as per https://github.com/travis-ci/travis-ci/issues/3772


BanzaiMan:
Thank you!

PR #291, opened 2015-07-13
Add support of OpenJDK 8

gildegoma:
To be merged when OpenJDK8 will be indeed added to build environment.

See https://github.com/travis-ci/travis-cookbooks/pull/494

### DIFF
--- a/user/ci-environment.md
+++ b/user/ci-environment.md
@@ -232,9 +232,10 @@ in order to minimize frictions when images are updated:
 ### JDK
 
 * Oracle JDK 7 (oraclejdk7)
-* OpenJDK 7 (openjdk7)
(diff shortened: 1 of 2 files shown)


plaindocs:
Thanks!

gildegoma:
@plaindocs I just want to make things the clearest possible (in case this PR was merged too early):

* OpenJDK 8 is part of the new Trusty Beta fully virtualised environment **with , but
* OpenJDK 8 is not present on Precise Legacy OpenVZ-based environment, and
* OpenJDK 8 is *probably* not present on the Precise Docker-based environment (I've not verified so far)

/cc @BanzaiMan @meatballhat

PR #294, opened 2015-07-15
Document commit_subject for IRC notifications

jhass:
Using commit_message breaks the notification if it contains a newline, so commit_subject is the far more useful key. Document it so people don't have to dive to https://github.com/travis-ci/travis-tasks/blob/master/lib/travis/addons/util/template.rb#L30


joshk:
Thank you!

PR #304, opened 2015-07-23
Documenting skip_cleanup

plaindocs:


user/customizing-the-build.md
```diff
@@ -114,7 +114,15 @@ Currently, neither the `after_success` nor `after_failure` have any influence on
 
 ## Deploying your Code
 
-An optional phase in the build lifecycle is deployment. This step can't be overridden, but is defined by using one of our continuous deployment providers to deploy code to Heroku, Engine Yard, or a different supported platform.
+An optional phase in the build lifecycle is deployment. This step can't be
+overridden, but is defined by using one of our continuous deployment providers
+to deploy code to Heroku, Engine Yard, or a different supported platform.
+
+When deploying files to a provider, prevent Travis CI from deleting all files
```
svenfuchs: I'd make this more explicit. Something like: "By default Travis CI resets your working directory (by running `git stash --all`), removing all changes that potentially have been made while running your tests or other build steps. In order to prevent this, and for example keep files generated as part of you build you can add the `skip_cleanup` option to your configuration:

plaindocs: Can't we assume that if people are deploying they are using build artefacts in most cases? 

I can talk about changes instead of files, and if you think mentioned `git stash` is important that also.

plaindocs:
@svenfuchs I've updated to address your comments. Just merge if you're happy.

PR #312, opened 2015-07-28
Add Screensaver Ninja as a Mac OS X tool to display Travis CI.

pupeno:


user/apps.md
```diff
@@ -292,6 +292,16 @@ By ThoughtWorks Inc.
 * [app store](https://itunes.apple.com/us/app/ccmenu/id603117688?mt=12&ign-mpt=uo%3D4)
 * [tutorial](/user/cc-menu/)
 
+![Travis CI in Screensaver Ninja with Custom CSS](/images/apps/screensaver-ninja.gif){:.app}
```
BanzaiMan: This screenshot should appear below the "Screensaver Ninja" heading.

pupeno: I can change it, but that's not true for CCMenu, not sure about others, but the next entry, which is for Linux, seems to also have the screenshot above the heading. Do you want me to still change it?

BanzaiMan: Oh! You are right about that.

pupeno: Do you need me to do anything else here?

plaindocs:
@BanzaiMan @lislis is looking into the formatting on this page. If this is the same as the others I'd say just merge it in.

PR #313, opened 2015-07-28
Note which infrastructures databases are available on

henrikhodne:


MariadeAnton:
This does not longer apply. Although it would be worth revisiting the idea of having a note in https://docs.travis-ci.com/user/database-setup/ the availability in the different infras. @plaindocs thoughts?

plaindocs:
Yeah, I think close it, and we'll refactor ALL of the infrastructure and docs when we hit that. Do we have an easy summary for what is available on what infrastructure for the meantime? Or do we need to make one.

MariadeAnton:
I think we'll need to make one. Although basics should be the same for everyone but Trusty, afaik.

PR #314, opened 2015-07-29
Julia versions doc update

tkelman:
for https://github.com/travis-ci/travis-build/pull/487

cc @staticfloat @ninjin

(p.s. Elliot I think it should be possible to go back as far as 0.2.0 on osx workers, is it worth copying 0.2.1 to 0.2-latest in case anyone wants to try that? It won't have `Pkg.test` so the default script won't work, but a few people might find it useful)

### DIFF
--- a/user/languages/julia.md
+++ b/user/languages/julia.md
@@ -20,15 +20,22 @@ and cc @tkelman @ninjin @staticfloat.
 
 ## Choosing Julia versions to test against
 
-Julia workers on travis-ci.org download and install the latest binary of Julia,
-either the most recent release version or the latest nightly build, from
-https://status.julialang.org. To select one or more versions, use the `julia:`
-key in your `.travis.yml` file, for example:
+Julia workers on travis-ci.org download and install a binary of Julia.
+You can select the most recent release version, the latest nightly build
+(downloaded from https://status.julialang.org), or a specific version number
+(downloaded from https://s3.amazonaws.com/julialang). To select one or more
+versions, use the `julia:` key in your `.travis.yml` file, for example:
 
     language: julia
     julia:
       - release
       - nightly
+      - 0.3
+      - 0.3.10
+
+If the version number contains one `.`, then the latest release for that minor version
+is downloaded. The oldest versions for which binaries are available is 0.3.1 for Linux,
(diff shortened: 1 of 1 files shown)


staticfloat:
Done.  If someone really wants to hurt themselves, they now are able to.  :)

ninjin:
Looks great!  @tkelman: Did the Travis guys let you try the PR out on their staging server?

tkelman:
Nope, I haven't heard anything other than https://github.com/travis-ci/travis-build/pull/487#issuecomment-125662783 and the "self-assignment" on that issue. The PR shouldn't be breaking, but I really don't understand (and didn't touch) rspec so there's always a chance I may have unintentionally broken something so it might be worth testing on staging if that's not a lot more trouble.

ninjin:
Unless things have changed since December, rspec does not execute the shell script that is generated from your Ruby code.

BanzaiMan:
It's deployed now. You can test it on production.

tkelman:
Awesome, thank. release, nightly, and x.y.z look like they're working. I'm getting "Unknown Julia version 0.3" though so I did something wrong there. At least no one's existing build should be broken right now, so we can take our time to figure it out and get the new feature working.

tkelman:
Judging by the syntax highlighting, is this some ridiculous yaml thing where un-quoted `0.3` is parsed as a float, but un-quoted `0.3.5` is parsed as a string? I'll try with quotes on all version numbers, may want to adjust the docs to recommend that.

tkelman:
All better. With quotes it works fine. Can we put a `to_string` somewhere to make that automatic?

tkelman:
Thanks @BanzaiMan for merging https://github.com/travis-ci/travis-build/pull/487 and https://github.com/travis-ci/travis-build/pull/496. With the latter (I tested it, works great), I think the docs are fine as-is, and we don't need to worry about recommending quotes.

Let me know if you have any feedback, otherwise it would be good to get this documented so I can announce the feature to the Julia mailing list.

tkelman:
Bump? I announced this anyway.

tkelman:
Bump again? The feature's working nicely, shame to not have it documented.

PR #344, opened 2015-08-28
Provide clearer documentation on how to encrypt slack tokens while using on_success, etc.

tim-evans:
I thought this would make it more helpful for other people like me who put the secure field at the top level alongside the `on_success` / `on_error` / `on_start` flags.


roidrage:
Thank you!

:heart: :raised_hands:

PR #345, opened 2015-08-28
Improve panda usage examples

softmoth:
This fixes a wrong `rakudobrew` command, and provides a more realistic example use case for installing dependencies.


BanzaiMan:
Thank you.

PR #346, opened 2015-09-02
Fix "go_build" vs "gobuild_args" disparity

tianon:
See https://github.com/travis-ci/travis-build/blob/305921ecd5eaf1feb4801cc0ed0b8d1561ab741c/lib/travis/build/script/go.rb#L78 for where it's actually used. :+1:


solarce:
:boom:

PR #353, opened 2015-09-08
Add some simple documentation about "go_import_path"

tianon:
I wasn't sure at what column I should flow the text here -- some of this file appears to be flowed somewhere around 122 characters, but then other parts are around 54 or so, so I'm really confused, but happy to update/reflow to whatever! :+1:


tianon:
(This is the companion to https://github.com/travis-ci/travis-build/pull/508.)

PR #369, opened 2015-09-28
Corrected port for vhost in php documentation

danrot:
This caused me some trouble, it seems like the given port in the vhost configuration file is 8080 instead of 80.

### DIFF
--- a/user/languages/php.md
+++ b/user/languages/php.md
@@ -250,7 +250,7 @@ Note that <code>sudo</code> is not available for builds that are running on the 
 You will need to have ``build/travis-ci-apache`` file that will configure your virtual host as usual, the important part for php-fpm is this:
 
 ```
-<VirtualHost *:80>
+<VirtualHost *:8080>
   # [...]
 
   DocumentRoot %TRAVIS_BUILD_DIR%


BanzaiMan:
Could you point to the build URLs (one that doesn't work, and one that does) that explains this issue? Thank you.

danrot:
Sry, I can't find the build URLs anymore... however, I am talking about this line: https://github.com/sulu-io/sulu-standard/pull/517/files#diff-7bde54db60a6e933518d8b61b929edceR117

It has to be 8080 in order to work, and even if I output the configuration file using `cat` it displays a file listening to port 8080 instead of 80.

danrot:
ping

BanzaiMan:
What do you mean by "in order to work"? How does it fail?

danrot:
I've already explained that previously, what's not clear about it? You can also check the given file i have edited in the documentation. You'll see that the actual value of the port in the vhost config file is 8080 instead of 80.

danrot:
I've already explained that previously, what's not clear about it? You can also check the given file i have edited in the documentation. You'll see that the actual value of the port in the vhost config file is 8080 instead of 80.

PR #372, opened 2015-09-29
Per suggestion from support email, remove unwanted space in mailto.

emdantrim:
Per suggestion from support email, remove unwanted space in mailto: link.


cotsog:
Thanks @emdantrim & @solarce! :+1:

PR #380, opened 2015-10-11
Perl 6 language documentation: more generic dependency installation

moritz:
Use the "installdeps" trick to install all declared dependencies,
instead of a list of hard-coded dependencies


plaindocs:
Thanks!

PR #383, opened 2015-10-14
Point to apt plug-in for customizing containers

tripleee:
The notice that using `apt` with `sudo` is not suitable for container-based
environments needs a pointer to what to use instead.

I searched literally for days and finally gave up on using containers.  This needs to be easier to find, especially from the page where you are directed by the dialog box which recommends to use containers instead.


tripleee:
Should I add something here to clear the failure now that the root cause for it has been fied?

plaindocs:
No need. Thanks!

PR #389, opened 2015-10-17
The Docker enabled environment now comes with docker-compose preinstalled

gildegoma:
After [Ubuntu Trusty Beta official roll-out](http://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta/), docker-compose is now preinstalled.

PS: This new Build Env is awesome, thanks :heart:

### DIFF
--- a/user/docker.md
+++ b/user/docker.md
@@ -148,13 +148,17 @@ docker login -e="$DOCKER_EMAIL" -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD" regi
 
 ### Using Docker Compose
 
-[Docker Compose](https://docs.docker.com/compose/) is not preinstalled, so
-install it manually by adding the following `before_install` step to
-`.travis.yml`:
+The [Docker Compose](https://docs.docker.com/compose/) tool is also [installed in the Docker enabled environment](/user/trusty-ci-environment/#Docker).
+
+If needed, you can easily replace this preinstalled version of `docker-compose` by adding the following `before_install` step to your `.travis.yml`:
 
 ```yaml
+env:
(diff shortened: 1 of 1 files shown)


plaindocs:
Is there a reason to talk about installing other versions of compose? Is this something people are likely to want to do? If not I'd be tempted to keep it pretty simple and remove the mention entirely.

gildegoma:
Thanks @plaindocs for the review! I understand your KISS point (and also began at first to write this change in this way), but the reasons why I think it might be good to give some hints about how `docker-compose` is installed (and how it could be replaced) are the following:

- By giving these instructions, it explains how `docker-compose` is preinstalled in the Travis CI environment ([the `curl` way is used, not the `pip` way](http://docs.docker.com/compose/install/)).
- Docker and related tools are still moving targets, and Travis users might run some `apt-get update` commands to get latest docker engine installed. This will not upgrade `docker-compose` as it is not part of the Docker Apt packaging and might lead to incompatibilities with existing `docker-compose` setup.
- It seems that - sooner or later - the [original Python-based implementation](https://github.com/docker/compose) of `docker-compose` will be replaced by https://github.com/docker/libcompose programmed in Go. I can figure out that some users might want to use this new tool and would appreciate to quickly find how to uninstall the preinstalled stuff.

For all these reasons, I think that keeping this information can help to save some support effort.
For better readability, we maybe could put them in a more specific sub-section (e.g. `#### Custom Installation`).

But I am totally okay if you and Travis team do prefer to remove these parts.

plaindocs:
Thanks for the extra explanation! I don't think it is a huge deal either way.

PR #390, opened 2015-10-20
Add notes about APT source requests are not currently processed

BanzaiMan:


BanzaiMan:
travis-build needs to change the message, and apt-source-whitelist README needs to be updated.

plaindocs:
@BanzaiMan I think this is no longer relevant, is that correct?

BanzaiMan:
@plaindocs It still is, but needs update.

remram44:
The fact that the whitelist process is stuck by lack of manpower to review PRs should be listed clearly IMHO. It is not tragic, there are sometimes workarounds, but users should know not to hold their breath over this.

PR #412, opened 2015-10-30
clarify automatically bundler caching

MariadeAnton:
A customer thought explicit bundler caching was not yet available due to the line *(not yet enabled)*
 http://i.nikwakelin.com/823m19p6dj-m29.png

@plaindocs, what do you think?


plaindocs:
I can see how that could be misinterpreted, will take a look.

PR #415, opened 2015-11-02
[user/artifacts] Add a note about file names and delimiter

Kami:
Right now the example which explains how to upload specific files uses colon (":") as a delimiter meaning it won't work with files which contain colon in the name.

This minor PR adds a note which clarifies this.

On a related note - I didn't test it out, but it seems like Travis assumes colon as a delimiter. Is this true? Or can you use arbitrary character (\0 or \n seems like a better default delimiter)? If you can't, can you escape paths which contain with a delimited (e.g. with double quotes)?


plaindocs:
Thanks!

PR #424, opened 2015-11-11
minor oversight of image name

emdantrim:


emdantrim:
Merging this since it's a one-character fix.

PR #425, opened 2015-11-12
user/notifications.md: use irc.freenode.net

Mikaela:
While irc.freenode.org is CNAME for irc.freenode.net, it should never be
used or documented anywhere, because the SSL certificates are valid for
*.freenode.net.

In case someone followed Travis documentation and configured their IRC
client that way, they would get strange errors on freenode's valid SSL
certificates being invalid or worse, think they should be invalid and
blindly accept them.

### DIFF
--- a/user/notifications.md
+++ b/user/notifications.md
@@ -125,22 +125,22 @@ to [your verified email addresses](https://github.com/settings/emails) on GitHub
 You can also specify notifications sent to an IRC channel:
 
     notifications:
-      irc: "irc.freenode.org#my-channel"
+      irc: "chat.freenode.net#my-channel"
 
 Or multiple channels:
 
     notifications:
       irc:
-        - "irc.freenode.org#my-channel"
-        - "irc.freenode.org#some-other-channel"
+        - "chat.freenode.net#my-channel"
+        - "chat.freenode.net#some-other-channel"
 
 As with other notification types you can specify when IRC notifications will be sent:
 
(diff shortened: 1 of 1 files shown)


BanzaiMan:
https://freenode.net/irc_servers.shtml says:

> Our main server rotation is chat.freenode.net.

Where is irc.freenode.net documented?

(I can't remember how I came up with the current documented value `irc.freenode.org`.)

Mikaela:
irc.domain.tld is the de-facto subdomain for IRC and also works on freenode due to being CNAME to chat.freenode.net and *.freenode.net

```
% host irc.freenode.net|head -n1
irc.freenode.net is an alias for chat.freenode.net.
```

Where the chat. comes from depends whom you ask, but usually they say that there was possibility long time ago of freenode moving to another protocol.

However I can change irc. to chat. if you wish, but people could be confused if they tried to connect the bot to another IRC network that doesn't have chat as CNAME. I will wait for confirmation before doing that.

Mikaela:
The Travis build failure doesn't seem to have anything to do with me(?).

BanzaiMan:
@Mikaela I don't think the build failure has anything to do with your PR. Let us go with `chat.freenode.net`, and I can merge this.

Thank you!

Mikaela:
Changed to chat.freenode.net.

We seem to be on weird timezones so I am always sleeping or something when you comment.

PR #426, opened 2015-11-12
Add syntax highlighting

lislis:
screen shot 2015-11-12 at 18 11 01
Looks like this right now. Question is how many languages we need. Right now there are styles for JS and Ruby


plaindocs:
Step by step towards better docs! Thanks @lislis!

PR #435, opened 2015-11-20
Allow standalone deployment to Heroku

BanzaiMan:
This PR modifies the app so that it can be deployed to Heroku.

The change in the server is somewhat arbitrary, so I'm up to suggestions if anyone has strong opinions.


BanzaiMan:
The build is failing because some sites are under maintenance and whatnot.

PR #441, opened 2015-12-04
Explain deployment is done via Ruby 1.9.3

BanzaiMan:
Closes https://github.com/travis-ci/travis-ci/issues/5205

### DIFF
--- a/user/deployment/script.md
+++ b/user/deployment/script.md
@@ -14,7 +14,7 @@ deploy:
   provider: script
   script: scripts/deploy.sh
   on:
-  	branch: develop
+    branch: develop
 {% endhighlight %}
 
 This example executes `scripts/deploy.sh` on the `develop` branch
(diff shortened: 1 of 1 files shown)


user/deployment/script.md
```diff
@@ -27,3 +27,19 @@ Be sure to make the script executable.
...
 a failure, and the build will be marked as "errored".
+
+## Deployment is executed by Ruby 1.9.3
+
+In order to ensure that deployments are executed reliably, we use a version of Ruby that
+is pre-installed on all of our build images.
+Currently, this is Ruby 1.9.3.
+
+A side effect of this is that, if you need to execute a command
+that requires a different Ruby, you need to execute it explicitly.
```
pabloh: @BanzaiMan, may be is also worth to document that if you execute some script at `after_deploy:` the rvm environment will be the same as the one on the build.

BanzaiMan: It shouldn't be. There is no default action for `after_deploy`, so this is not a problem. See https://travis-ci.org/BanzaiMan/travis_production_test/builds/94971838

pabloh: And the same happens for `before_deploy:` the rvm environment will be the same as the one on the build.

BanzaiMan: Do you have an example of the problems with `before_deploy` and `after_deploy`?

pabloh: @BanzaiMan , I think is worth mentioning because I you happen to install a gem during `before_deploy` it won't be immediately available at your deploy script until you change your rvm environment.

BanzaiMan: Ah! I understand what you are saying now. Thank you for the input!

PR #442, opened 2015-12-06
Clarify which Ruby versions are installed

waits:
The "Supported Ruby Versions" section claims that "recent" versions of ruby-2.2.x are installed, when in reality only 2.2.0 is and not 2.2.1, 2.2.2, or 2.2.3. I changed that line to clarify that just 2.2.0 is installed.

### DIFF
--- a/user/languages/ruby.md
+++ b/user/languages/ruby.md
@@ -18,7 +18,7 @@ configuration](/user/customizing-the-build/) guides first.
 
 The Ruby VM provides recent versions of:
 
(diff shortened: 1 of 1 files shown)


waits:
Here's [an issue] with more context, although that has more to do with the build information fold giving bad information.

waits:
I have no idea why the build is failing; seems like an unrelated issue with a link to http://travalizit.org/ which is now down.

plaindocs:
Thanks!

PR #444, opened 2015-12-07
Proof of concept of version tables for lang pages

plaindocs:
WIP

This is the kind of info I'm looking at. More relevant for some languages than others.

At a minimum we need to use the same info here and on the environment docs so we're not repeating ourselves. Then we can look at actually generating the info.

### DIFF
--- a/user/languages/php.md
+++ b/user/languages/php.md
@@ -10,6 +10,15 @@ This guide covers build environment and configuration topics specific to PHP pro
 
 PHP builds are not available on the OSX environment.
 
+|---
+|   | Legacy | Docker | OSX
+|---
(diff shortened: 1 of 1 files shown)


plaindocs:
@MariadeAnton

waits:
I can help out on this if needed. I've been looking at inconsistencies in the listed Ruby versions, and it would be nice if there was a standard format for all languages that could be shared between pages.
https://github.com/travis-ci/travis-ci/issues/5190
https://github.com/travis-ci/docs-travis-ci-com/pull/442

plaindocs:
Hi @waits! That sounds great. I'm going to talk to the team, to see how much effort generating a yaml for each CI environment would be on the one hand.

If you want drop any suggestions or thoughts onto this thread in the meantime that would be awesome. Then of course there will be some significant manual effort implementing all the changes.

PR #448, opened 2015-12-10
Fix a YAML indent issue with a caching.md example

aprescott:


BanzaiMan:
This is valid YAML either way. Did you try it and fail? If so, do you have a build log URL that shows it?

aprescott:
Is it not conventional to indent to show the level of nesting? For instance, the CocoaPods example further down has this:

```yaml
 cache:
   directories:
     - path/to/Pods
```

I know it would lead to the same Ruby hash as

```yaml
 cache:
   directories:
   - path/to/Pods
```

But I think the extra indentation on the `- vendor/bundle` line is idiomatic, no? Maybe I made that up. :) Either way, it's inconsistent with the CocoaPods indentation, so maybe one of them should be fixed. I don't know which is more prevalent on the docs site.

plaindocs:
I don't know enough about the hard details of YAML to make the call - but knowing what I do I'd probably pick the second. @BanzaiMan if you want to make the call, I'll add it to a style guide and we'll slowly to converge to consistency.

plaindocs:
@MariadeAnton do you have input on this?

MariadeAnton:
@plaindocs: I'm all in for consistency, could you look into doing so for https://docs.travis-ci.com/user/caching ? 

@aprescott, sorry for the delay hearing back from us and thanks for pointing out the consistency issue! Our docs have been updated since you opened the PR (as our caching has recently improved) and the lines:

```
cache:
    directories:
     - vendor/bundle
```

Are no longer there so I'm closing this PR. But we'll look into making the indentation consistent. Would you be interested in helping with it? Thank you!

plaindocs:
@MariadeAnton are we saying we definitely want to converge on the extra space for every level of indent? Roger that.

PR #451, opened 2015-12-11
add yaml syntax highlighting styles

lislis:
important: the language must be specified as `yaml`. `yml` does NOT work


plaindocs:
Awesome. :-) Is this good to merge?

We have quite a few random langs in the docs, not sure how much sense it makes to strip loads of them out.

PR #455, opened 2015-12-15
add MySQL 5.6 info

MariadeAnton:


user/database-setup.md
```diff
@@ -131,6 +131,24 @@ before_install:
...
+sudo: required
+addons:
+  apt:
+    packages:
+    - mysql-server-5.6
+    - mysql-client-core-5.6
+    - mysql-client-5.6
+```
+
+It's worth noting that **the default mysql root password** will be different, and that you can use sudo to set it to what you expect, which is `travis` everywhere else.
```
BanzaiMan: This reads a bit awkward. Perhaps we can just add a snippet to set the password to the expected value elsewhere.

user/database-setup.md
```diff
@@ -131,6 +131,26 @@ before_install:
...
+  apt:
+    packages:
+    - mysql-server-5.6
+    - mysql-client-core-5.6
+    - mysql-client-5.6
+```
+
+Make sure to use `sudo` to set the default **mysql root password** to the expected value elsewhere, `travis`.
+
+It's worth noting that **the default mysql root password** will be different, and that you can use sudo to set it to what you expect, which is `travis` everywhere else.
```
BanzaiMan: You can remove this line, right?

BanzaiMan: Perhaps something like:

> Note that the default `root` password will be set by the package. To set it to `travis` (which is the value used everywhere else), do: …

user/database-setup.md
```diff
@@ -131,6 +131,23 @@ before_install:
...
+addons:
+  apt:
+    packages:
+    - mysql-server-5.6
+    - mysql-client-core-5.6
+    - mysql-client-5.6
+```
+
+Note that the default root password will be set by the package. You can set it to travis (which is the value used everywhere else), do: ...
+
```
plaindocs: As well as adding info on how to set the root password, are we also telling people what the default is set to? Or is it generated each time?

PR #465, opened 2016-01-04
Add notes to optimise PHP runs by disabling XDebug

andrewnicols:
XDebug is required for code coverage generation, and has a high impact upon
test performance. Composer recommend disabling XDebug when running composer
(https://getcomposer.org/doc/articles/troubleshooting.md#xdebug-impact-on-composer),
and huge benefits can be seen by running phpunit without xdebug enabled.

These changes:
* add information to this effect to the Speeding up the build page; and
* improve organisation of the PHP page to enable direct linking for
  disabling PHP extensions in travis.yml.

### DIFF
--- a/user/speeding-up-the-build.md
+++ b/user/speeding-up-the-build.md
@@ -136,3 +136,33 @@ order to make it faster, you may try caching the dependencies.
 You can either use our [built-in caching](/user/caching/) or roll your own on S3. If you
 want to roll your own and you use Ruby with Bundler, check out [the great WAD project](https://github.com/Fingertips/WAD).
 For other languages, you can use s3 tools directly to upload and download the dependencies.
+
+## Environment-specific ways to speed up your build
+
+In addition to the optimizations implemented by Travis, there are also
+several environment-specific ways you may consider increasing the speed of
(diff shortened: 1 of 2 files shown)


user/speeding-up-the-build.md
```diff
@@ -136,3 +136,33 @@ order to make it faster, you may try caching the dependencies.
...
+which may be faster.
+
+#### Using phpdbg example
+
+    before_script:
+      - phpenv config-rm xdebug.ini
+      - composer install
+
+    script:
+      - phpdbg -qrr vendor/bin/phpunit
```
BanzaiMan: `phpunit` should be pre-installed. I understand that it might be common to put it in `vendor/bin` as a part of the cache, but I don't think we should mention this in our docs.

andrewnicols:
Thanks @BanzaiMan. I've updated the docs accordingly.

For our purposes we wish to test with a known version of phpunit, and not whatever happens to be on the system - largely because different options are available between 4.8, and 5.X and we don't know for certain which version will be installed on travis.

BanzaiMan:
Thank you!

PR #466, opened 2016-01-04
`-no-skin` is removed from Android docs

tasomaniac:
`-no-skin` parameter removes the skin. This means that the emulator will be 320x480 pixel widthxheight and 320 density which messes up with most of the tests.
The default skin is 480x800 which is fine for most of the cases. 

`--skin WXGA720` can be used like below to make tests more trusted. This is the best emulator config that is found by default. 

    - echo no | android create avd --force -n test -t android-19 --abi armeabi-v7a --skin WXGA720

### DIFF
--- a/user/languages/android.md
+++ b/user/languages/android.md
@@ -125,7 +125,7 @@ If you feel adventurous, you may use the script [`/usr/local/bin/android-wait-fo
     # Emulator Management: Create, Start and Wait
     before_script:
       - echo no | android create avd --force -n test -t android-19 --abi armeabi-v7a
-      - emulator -avd test -no-skin -no-audio -no-window &
+      - emulator -avd test -no-audio -no-window &
       - android-wait-for-emulator
       - adb shell input keyevent 82 &
 


BanzaiMan:
This is an example, though. Is it lacking in some way?

tasomaniac:
`320x480` device with 320 density for Android means that the device is actually `160x240`. With this nothing fits in the screen. Non UI related tests in the emulator passes but UI related Espresso tests will fail. 

Circle CI also defaults to only `-no-audio -no-window`

screen shot 2016-01-04 at 23 47 56

BanzaiMan:
Thanks for the information.

ardock:
@tasomaniac Thanks for the explanation, I added your comments on this stackoverflow [response](http://stackoverflow.com/a/33667117/1009132).

tasomaniac:
Thank you for merging it. 
I only fixed the problem there. It would be really good to have additional information about screen sizes and available skins there. 

I found that `--skin WXGA720` is really close to real devices sizes. 

I didn't write any because I am not good at technical writing. :)

PR #490, opened 2016-01-25
Haskell docs - better .travis.yml examples

lwm:
Well, the current examples aren't that great, so here are some better ones IMHO. The first, a complicated cabal example, the second, a lot of GHC versions covered, and the final, a stack X docker one. 

EDIT: Also, the `stack` Travis file would probably be useful here too.


lwm:
Closing in favour of #500.

PR #492, opened 2016-01-25
Adding Private GitHub OAuth scopes

plaindocs:
We should have private repo scopes as well as open source as per travis-pro/team-sapphire/issues/214

@sinthetix could you take a look at this?


sinthetix:
Hey @plaindocs I think overall this looks great but maybe we need to specify WHY we need `repo` access (since Github doesn't have a read-only for private repositories). This question comes up a LOT so it would be nice to have in the docs :)

plaindocs:
Good catch. Is this sufficient do you think?

sinthetix:
I think so! Merging :)

PR #493, opened 2016-01-26
Add a docs page detailing the IP addresses each build environment uses

henrikhodne:
I'm not quite happy with the wording here, for example almost every paragraph starts with "For". So I'd love some feedback on how to better word this.

### DIFF
--- a/user/ip-addresses.md
+++ b/user/ip-addresses.md
@@ -13,8 +13,6 @@ Container-based (travis-ci.org) | `nat-org.aws-us-east-1.travisci.net` (`52.0.24
 OS X | `208.78.110.192/27`
 Sudo-enabled Linux | N/A (see below)
 
-If we need to add more capacity to the network stack of our container-based infrastructure, we might add more IP addresses
-
 We do not have static public IP addresses available for jobs running on the sudo-enabled Linux infrastructure at this time. We recommend using one of the other infrastructures if you need static IP addresses.
 
 Note that these ranges can change in the future.
--- a/user/ip-addresses.md
+++ b/user/ip-addresses.md
@@ -4,12 +4,17 @@ layout: en
 permalink: /user/ip-addresses/
 ---
 
-For some it might be necessary to know the public IP addresses of build machines in order to whitelist them to access some internal resources. We run builds in a variety of different infrastructures, so which one you use determines which IP range you need to whitelist.
+For some it might be necessary to know the public IP addresses of build machines in order to whitelist them to access some internal resources. We run builds in a variety of different infrastructures, so which ones you use determines which IP ranges you need to whitelist.
 
-For jobs running on the container-based infrastructure, the ranges are `54.172.141.90/32` and `52.3.133.20/32` for travis-ci.com (private repositories), and `52.0.240.122/32` and `52.22.60.255/32` for travis-ci.org (public repositories).
+Infrastructure | IP ranges
+-------------- | ---------
+Container-based (travis-ci.com) | `nat-com.aws-us-east-1.travisci.net` (`54.172.141.90/32`, `52.3.133.20/32` as of March 14th, 2016)
+Container-based (travis-ci.org) | `nat-org.aws-us-east-1.travisci.net` (`52.0.240.122/32`, `52.22.60.255/32` as of March 14th, 2016)
+OS X | `208.78.110.192/27`
+Sudo-enabled Linux | N/A (see below)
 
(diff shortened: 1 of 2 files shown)


roidrage:
This looks good to me. Is it worth adding the hostnames that would resolve to all IP addresses too, at least for the EC2 bits?

user/ip-addresses.md
```diff
@@ -0,0 +1,15 @@
+---
+title: IP Addresses
+layout: en
+permalink: /user/ip-addresses/
+---
+
+For some it might be necessary to know the public IP addresses of build machines in order to whitelist them to access some internal resources. We run builds in a variety of different infrastructures, so which one you use determines which IP range you need to whitelist.
```
plaindocs: You could simply this to "If you need to know the public..."

user/ip-addresses.md
```diff
@@ -0,0 +1,15 @@
...
+
+For jobs running on the container-based infrastructure, the ranges are `54.172.141.90/32` and `52.3.133.20/32` for travis-ci.com (private repositories), and `52.0.240.122/32` and `52.22.60.255/32` for travis-ci.org (public repositories).
```
plaindocs: And instead of running text I'd present these IP ranges as either a table or or a list for ease of lookup:

Infrastructure | IP range
------------ | -------------
container based | xxxx-yyyy

OR
* Container based: xxxx-yyyy
* Private Repositories: xxxxx-yyyy

user/ip-addresses.md
```diff
@@ -0,0 +1,15 @@
...
+
+For jobs running on our Mac infrastructure, the ranges are 208.78.110.192/27.
+
+For jobs runing on the Linux infrastructure with sudo enabled, we currently do not have static public IP addresses available. We recommend using one of the other infrastructures if you need static IP addresses.
```
plaindocs: If you are converting to table/list, add this entry with an N/A or something, as well as the explanation.

user/ip-addresses.md
henrikhodne: I'm not quite sure what part I should replace with that. It seems like replacing "For some it might be necessary to know" with "If you need to know", but there's no "…then xyz" to correspond with the "if", and I couldn't think of what to write for that.

MariadeAnton:
@henrikhodne I really like the table! Is this ready to merge or are you still working in it? Thanks!

henrikhodne:
Ready to merge if you think the text reads okay. I'm not 100% sure about it myself, but I'm not sure what I would change.

MariadeAnton:
@henrikhodne I see what you mean, I'll add some comments with a proposed rewording in a bit. Let me know what you think.

user/ip-addresses.md
```diff
@@ -0,0 +1,18 @@
...
+Infrastructure | IP ranges
+-------------- | ---------
+Container-based (travis-ci.com) | `nat-com.aws-us-east-1.travisci.net` (`54.172.141.90/32`, `52.3.133.20/32` as of March 14th, 2016)
+Container-based (travis-ci.org) | `nat-org.aws-us-east-1.travisci.net` (`52.0.240.122/32`, `52.22.60.255/32` as of March 14th, 2016)
+OS X | `208.78.110.192/27`
+Sudo-enabled Linux | N/A (see below)
+
+We do not have static public IP addresses available for jobs running on the sudo-enabled Linux infrastructure at this time. We recommend using one of the other infrastructures if you need static IP addresses.
+
+Note that these ranges can change in the future.
```
MariadeAnton: I may also add a note at the end linking to the infrastructures: 

"More details about our different infrastructures are available on the [virtualization environments page](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments/)"

user/ip-addresses.md
```diff
@@ -0,0 +1,18 @@
+---
+title: IP Addresses
+layout: en
+permalink: /user/ip-addresses/
+---
+
+For some it might be necessary to know the public IP addresses of build machines in order to whitelist them to access some internal resources. We run builds in a variety of different infrastructures, so which ones you use determines which IP ranges you need to whitelist.
```
MariadeAnton: Knowing the IP addresses of the build machines Travis CI uses can be helpful when you need them whitelisted to access your internal resources. As builds are run in a variety of different infrastructures, the IP ranges to whitelist depend on the infrastructure your builds are running at:

MariadeAnton:
@henrikhodne thank you!

PR #500, opened 2016-02-06
Add better haskell examples

lwm:
Following from #490, I put this on the `gh-pages` branch, so it should be a quick merge.


lwm:
:rocket:

PR #506, opened 2016-02-13
fix redirect_to -> redirect_uri

JCBarry:
Based on the code here: https://github.com/travis-ci/travis-api/blob/master/lib/travis/api/app/endpoint/authorization.rb#L97

The actual parameter is `redirect_uri` instead of `redirect_to`.  This pull request adjusts the documentation to accurately reflect the parameter.


plaindocs:
Thanks!

PR #509, opened 2016-02-16
Update README.md for the bintray provider

uriahl:
Hello, I am with JFrog (home of Bintray). There was a small change that was needed to be made for the documentation of the Bintray provider section:

"Removed small documentation mistake about having to use brackets for the excludePattern param in the bintray descriptor example."

We have verified this to be incorrect. There is no need to use brackets for the 'excludePattern' param.

Note - an equivalent PR was submitted to the travis-ci/dpl repository readme file.

Thank you.


plaindocs:
Thanks!

PR #516, opened 2016-02-29
Update travis_wait prefix options

MariadeAnton:


MariadeAnton:
@plaindocs: what do you think?

user/common-build-problems.md
```diff
@@ -296,6 +296,10 @@ If you have a command that doesn't produce output for more than 10 minutes, you
 
 `travis_wait` writes a short line to the build log every minute for 20 minutes, extending the amount of time your command has to finish.
 
+If you need more than 20 minutes, you can prefix `travis_wait` with a custom value. For example, this would extend the time to 30 minutes:
```
plaindocs: Looks good. You could clarify a little if you wanted:

If you expect the command to take than 20 minutes, prefix `travis_wait` with a greater number. For example, to extend the time to 30 minutes:

Also, we get better syntax highlighting with backtick syntax:

    ````yaml

MariadeAnton:
@plaindocs updated!

plaindocs:
Two nitpicking comments:

* " a function that's exported by our build environment." adds almost nothing 
* I did not notice before, but we use prefixing twice, once for the command, and once for the custom number, which one could say is ugly. ;-

MariadeAnton:
@plaindocs: thank you for nitpicking!

I was the one adding the prefixing on both commands per your suggestion, could you update this or let me know how would we do this the "pretty" way? Thanks much.

PR #518, opened 2016-03-01
Fix for off-by-one error.

emdantrim:


plaindocs:
:+1: Even better is to remove the number entirely to prevent re occurrence - which I've done in the branch I'm working on at the mo (and have not merged :/).

PR #528, opened 2016-03-23
YAML-ify Xcode information.

emdantrim:
In hopes of removing some multiplicity of data, making it easier for us to programmatically update this in the future. :ok_hand:


MariadeAnton:
This has changed a bit the indentation, but looks good to me.
Before:

image


After:

image

PR #533, opened 2016-03-25
Fix example script for excluding tests that rely on secure vars

svenfuchs:
The example, as given in our docs, will always fail the build when run on a pull request. I think that's not the intention behind the example script.

```
bash-3.2$ TRAVIS_PULL_REQUEST="true"; [ "${TRAVIS_PULL_REQUEST}" = "false" ] && echo "Secure vars available" || true; echo $?
0
bash-3.2$ TRAVIS_PULL_REQUEST="false"; [ "${TRAVIS_PULL_REQUEST}" = "false" ] && echo "Secure vars available" || true; echo $?
Secure vars available
0
```

In the case when `$TRAVIS_PULL_REQUEST` is `true` (or anything else) one still would want the tests to pass, I guess, so it should default to `true`, not `false`.

### DIFF
--- a/user/pull-requests.md
+++ b/user/pull-requests.md
@@ -50,7 +50,7 @@ environment variables are available, or disable them for pull requests entirely.
 Here's an example of how to structure a build command for this purpose:
 
     script:
-      - '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && bundle exec rake tests:integration || false'
+      - '[ "${TRAVIS_PULL_REQUEST}" = "false" ] && bundle exec rake tests:integration || true'
 
 The environment variable `${TRAVIS_PULL_REQUEST}` is set to `"false"`
 when the build is for a normal branch commit. When the build is for a


plaindocs:
@svenfuchs I'm having trouble parsing this one. 

The intention of the example afaik is to only run the test if we're not running on a pull request (ie if `TRAVIS_PULL_REQUEST="false"`.

So I think the example is correct.

```
TRAVIS_PULL_REQUEST="true"; [ "${TRAVIS_PULL_REQUEST}" = "false" ] && echo "RUN IT" || false
$ TRAVIS_PULL_REQUEST="false"; [ "${TRAVIS_PULL_REQUEST}" = "false" ] && echo "RUN IT" || false
```

Does this make sense?

MariadeAnton:
I think both of you might be right.
If the expected behaviour would be run the tests if this is not a PR, otherwise do nothing, what do you think about using this line instead?

    - if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then bundle exec rake tests:integration; fi

plaindocs:
I like @MariadeAnton s version. I'll reword the surrounding text while I'm there.

@svenfuchs ?

svenfuchs:
@plaindocs your example would have the same issue, ... it would fail the build because the exit code would be `1` when `$TRAVIS_PULL_REQUEST` is not `false`.

You can check that by outputting the exit code with `echo $?`, as in:

```
$ TRAVIS_PULL_REQUEST="true"; [ "${TRAVIS_PULL_REQUEST}" = "false" ] && echo "RUN IT" || false; echo $?
1
$ TRAVIS_PULL_REQUEST="false"; [ "${TRAVIS_PULL_REQUEST}" = "false" ] && echo "RUN IT" || false; echo $?
RUN IT
0
```

@MariadeAnton's version works. That's because an `if` statement does not have an exit code on its own in bash:

```
$ if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then echo "RUN IT"; fi; echo $?
RUN IT
0
```

Whatever expression we recommend, the exit code of the full statement should not be `1` unless the exit code of the wrapped statement (e.g. `rake tests:integration`) is `1`.

Does that make sense?

svenfuchs:
I think an `if` statement maybe makes most sense because it might be the least confusing one for most users:

> In order to run commands on pull request builds:

```
if [ "$TRAVIS_PULL_REQUEST" = "true" ]; then bash ./travis/run_on_pull_requests; fi
```

> In order to run commands on non-pull-request builds:

```
if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then bash ./travis/run_on_pull_requests; fi
```

plaindocs:
OK, I've pushed two commits to https://github.com/travis-ci/docs-travis-ci-com/pull/613 one using your latest examples as is, and another tightening up some of the surrounding text.

If it looks good I'll merge it in.

PR #534, opened 2016-03-26
nodejs: stable keyword is deprecated

paulirish:
using `stable` for your node version is deprecated now: https://github.com/creationix/nvm#usage

ref: https://github.com/GoogleChrome/lighthouse/pull/85


plaindocs:
Thanks!

PR #537, opened 2016-03-29
APT add-on installs also in container-based builds

krlmlr:
See e.g. https://travis-ci.org/rstats-db/RMySQL/builds/119360508#L426

Not sure if I got the links right. Perhaps this information should be placed more prominently.

CC @jimhester @hadley @craigcitro

### DIFF
--- a/user/languages/r.md
+++ b/user/languages/r.md
@@ -115,6 +115,32 @@ language: r
 pandoc_version: 1.16
 ```
 
+### APT packages
+
+Use the [APT addon][apt-addon] (currently in beta)
+to install APT packages on both container-based (`sudo: false`)
+and standard (`sudo: required`) infrastructures.
+The snippet below installs a prerequisite for the R package `xml2`:
+
+```yaml
+addons:
+  apt:
+    packages:
+      - libxml2-dev
+```
+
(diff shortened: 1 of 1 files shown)


hadley:
LGTM, although I think we could add even more about the apt-addon, include an small example, e.g.

```yaml
addons:
  apt:
    packages:
      - libxml2-dev
```

We also need to mention that these dependencies are transitive - i.e. if your package depends on xml2, you'll also need to install `libxml2-dev`

krlmlr:
Could `apt_packages` eventually forward to the `apt-addon`?

jimhester:
@krlmlr Maybe they could be forwarded but not all packages are in the whitelist for the addon so it may result in confusing error messages.

krlmlr:
@jimhester: c55d54d creates an own section for this information, let me know if you prefer the previous version.

user/languages/r.md
```diff
@@ -115,6 +115,32 @@ language: r
 pandoc_version: 1.16
 ```
 
+### APT packages
+
+Use the [APT addon][apt-addon] (currently in beta)
```
plaindocs: I *think* we're out of beta?

plaindocs:
Thanks!

PR #545, opened 2016-04-07
Remove extra dash in the command line switch.

kwilczynski:
Fix the following `gpg` invocation error:

```
gpg: Invalid option "--pass-phrase-fd"
```

As per the https://www.gnupg.org/documentation/manpage.html, there is no dash (- present in the command line switch name.


plaindocs:
Good catch! Thanks

PR #546, opened 2016-04-07
Command to install specific ElasticSearch version

veatch:
I think this line was accidentally deleted. Restored it, along with command to restart ElasticSearch to start using new version.


plaindocs:
Thanks for the catch!

PR #568, opened 2016-04-27
Sidebar toggle

lislis:
change some colors, add sidebar toggle, persist open section.

The tricky part is knowing exactly which page is active. In theory we could also store the name of the clicked link. But this is a good first step

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -10,7 +10,7 @@
         (function() {
           Swiftype.key = 'tXvDDzd4fNTBnLvxfEyx';
 
-		  Swiftype.searchSearchFields = {
+      Swiftype.searchSearchFields = {
             "page": ["title, body, tags"]
           };
           Swiftype.autocompleteSearchFields = {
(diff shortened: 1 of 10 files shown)


MariadeAnton:
Thank you @lislis!!

plaindocs:
Hey, that is awesome, esp persisting the open section. 

Currently the Menu and the subheadings are both H2, I'll change the subs to H3. But can you make close the gaps up, and provide some indication that there is a fold? I can try to find some examples if that helps, but I'm thinking more a single menu than sections of a side bar.

lislis:
I added a little arrow to indicate the fold. 
Please ping me once you changed the h2 to h3, then I'll rebase and adjust the styles.

plaindocs:
Done.

lislis:
Updated and rebased and squashed. Let me know if you need something else

plaindocs:
Looking better and better.

Can we collapse the spacing so that it looks like a single list? I want it to look like a single menu, not various sections of a "sidebar". Does that makes sense?

A few examples

http://gallium.readthedocs.io/en/stable/tgsi.html
http://docs.biicode.com/c++/gettingstarted.html
http://docs.biicode.com/c++/advanced/toolchain.html
https://docs.djangoproject.com/en/1.9/topics/forms/#django-s-role-in-forms (although right side)

lislis:
@plaindocs like so?
screen shot 2016-05-04 at 12 15 47

plaindocs:
I think the gap between "Languages" and "Deployments and Uploads" is more what I had in mind (although that one is different because it it is currently empty, obviously won't be)

lislis:
aaaaaaah, more like this:
screen shot 2016-05-04 at 12 24 48
Would you want to keep the indented list?

plaindocs:
That looks great! Thanks @lislis - I'll fill in the extra two empty sections and make sure that works, then I'm happy to ship this as is.

@MariadeAnton has some comments on folding persistence, but we can discuss those separately I think.

MariadeAnton:
party-gif

PR #571, opened 2016-04-29
Clarifies lead of environment variables docs

jfinkels:
This commit removes a top-level list for information that is not
appropriate for a list. This also helps to prevent rendering
weirdness with multiple unordered lists where one was
intended. It also adds an initial global sentence describing the
main idea of the lead section.

### DIFF
--- a/user/environment-variables.md
+++ b/user/environment-variables.md
@@ -8,21 +8,22 @@ A common way to customize the build process is to use environment variables, whi
 
 <div id="toc"></div>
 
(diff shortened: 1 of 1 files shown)


BanzaiMan:
@plaindocs Could you review this? Thanks.

plaindocs:
@jfinkels thanks very much for that! Comment here if you've any comments or questions on the other changes I made.

jfinkels:
Looks good, thanks!

PR #576, opened 2016-05-05
Adding 7.3.1

solarce:
- Added 7.3.1 version
- Tweaked so that we can display the Xcode minor version but the
  major version links still work

### DIFF
--- a/_data/xcodes.yml
+++ b/_data/xcodes.yml
@@ -1,7 +1,6 @@
 osx_images:
   - image: xcode7.3
-    xcode: "7.3.1"
-    xcode_alias: "7.3"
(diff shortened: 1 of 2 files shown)


emdantrim:
>Tweaked so that we can display the Xcode minor version but the major version links still work

That's what `xcode_full_version` is for. `xcode` is for the "X.Y", whereas `xcode_full_version` is "X.Y.Z Beta/GM/GA", since we'd previously decided to overload the `xcode7` image with Xcode 7.0.1, `xcode7.1` with 7.1.1, etc.

solarce:
Updated to use `xcode_full_version`

PR #577, opened 2016-05-05
Remove whitespace around list items

backspace:
As [reported by @solarce](https://travisci.slack.com/archives/teal/p1462462487001036
), Firefox was showing the text of each line item beneath its bullet. For example:

image

This unsightly template formation seems required to prevent the generated whitespace being interpreted as paragraphs in Markdown. See this discussion:
https://github.com/Shopify/liquid/issues/216

Haml has `<` and `>` to handle this kind of thing, hopefully eventually Liquid will get something like it.


solarce:
:shipit:

PR #579, opened 2016-05-06
Correct Node.js versions

okuryu:
There're some release versions for Node.js v4, v5, and v6.

See: https://nodejs.org/en/download/releases/


plaindocs:
Thanks for this, I'll work the updated values into a rewrite of the whole node page soon.

PR #584, opened 2016-05-10
Sidebar tweaks

lislis:
hey there @plaindocs 
I did a little scroll thing for sidebar. Is that how you meant it?

Also, I toyed around with some styles. What do you think about this for the navigation? The blue gets quite heavy imo. But this is just an idea:
screen shot 2016-05-10 at 11 38 35


lislis:
Wrong base branch ...

PR #587, opened 2016-05-10
Add instructions for testing with Speclj in Clojure

beccanelson:
I love testing my Clojure projects with [Speclj](https://github.com/slagyr/speclj), and I noticed that it's not included here yet. I've had success using it with Travis, so I've added a few lines to the Clojure documentation.

### DIFF
--- a/user/languages/clojure.md
+++ b/user/languages/clojure.md
@@ -41,7 +41,7 @@ Projects that find this sufficient can use a very minimalistic .travis.yml file:
 
     language: clojure
 
-### Using Midje or Speclj on travis-ci.org
+### Using Midje on travis-ci.org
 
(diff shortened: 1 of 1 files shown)


user/languages/clojure.md
```diff
@@ -41,7 +41,7 @@ Projects that find this sufficient can use a very minimalistic .travis.yml file:
 
     language: clojure
 
-### Using Midje on travis-ci.org
+### Using Midje or Speclj on travis-ci.org
```
BanzaiMan: Is there a reason that this is an "or" heading here? Is Speclj any way related to Midje? If it is not, I feel that is is much cleaner to have a separate section instead.

beccanelson:
Good call. I originally tried to avoid redundancy, as the instructions for using Speclj and Midje are nearly identical. However, including it as a separate heading does make more sense.

BanzaiMan:
Thank you!

PR #604, opened 2016-06-10
Clarifying Cron 

plaindocs:


carlad:
Hey @plaindocs thanks for this it looks good.

The images I included were actually just base images from which to make all the graphics on that page from. I didn't explain that sorry. I assumed you were the person adding red circles and boxes to the images.

We actually need to replace all the images. I've added a comment to the issue, including a new screen grab for the base image of the first image.

aakritigupta:
@carlad What is the time zone of the scheduled time of the cron job?

I see that in other parts of the UI, we do add time zone info, but not for this feature.

A user asked about it too.

plaindocs:
Hey,

Sure I can fix these. The screenshots you took look pretty small on my screen, I've taken new ones, and used ugly orange instead of really ugly red.

I also suspect we could easily reduce the amount of shots, probably just include one of the Cron Settings tab, no need for the next three...

@aakritigupta they look local to me, so probably from the location in the github profile? But you're right we should confirm and specify.

carlad:
Hey @aakriti can you ask in the #hpi channel. The students who built the cron feature lurk there, and know more about it than I do. 

As far as i understand the created_at time (when a cron is added by the user) is UTC.

MariadeAnton:
The time displayed will vary depending on the user's system local timezone. It's stored in UTC and then transformed to be displayed. Makes sense?

carlad:
Thanks for this @plaindocs . I've merged this PR but the auto deployment has failed, and the build log on Heroku is empty. I'm not sure what's up as there are no tests (so I can't check anything on travis-ci), and am not sure how to redeploy. Can you take a look?

plaindocs:
Looking now.

plaindocs:
I restarted the build, and it looks ok now. But we should probably have a look and see if there is an underlying issue.

carlad:
thanks @plaindocs

PR #608, opened 2016-06-10
added note on jdk_switcher for osx

acnagy:


plaindocs:
So you can't test against multiple JDKs on OS X, and you can't use `jdk_switcher` either. So probably switching Xcode won't help?

acnagy:
@plaindocs Ah... Looks like that note might be best somewhere else... The issue is that `jdk_switcher` doesn't work with OS X builds yet, so you can't switch within a single job - or at all, for that matter on OS X. But because different version of Xcode have to different JDKs by default, we're recommending switching Xcode images entirely if you need to test against a different JDK - for example, it's been successful to try Xcode7.3 for JDK 8.

PR #617, opened 2016-06-15
Describe workaround for Coverity's new certificate

flatcap:
Coverity's new site certificate, for scan.coverity.com isn't in the system's
CA Certs bundle.  This means Coverity scans cannot be performed.

Offer the user two ways to work around this problem:
  - Update the CA Certs using apt
  - Download the certificate manually using openssl

Fixes: https://github.com/travis-ci/travis-ci/issues/6142


BanzaiMan:
@flatcap Hello! Thanks for the PR. As mentioned in the other issue, I have opted to handle this with the addon itself.

PR #618, opened 2016-06-15
added section to osx build environment only on jdk_switcher

acnagy:
Created a short section on the osx environment page regarding jdk_switcher not being available, to serve in addition to the note on the java languages page


plaindocs:
:-)

PR #636, opened 2016-06-22
include node_modules in the caching page

afeld:
This is a minor change, but I always end up Googling how to cache NPM packages in Travis, but this page doesn't come up. Basically including it here for better SEO :wink: since I figure it's a fairly common thing to want to cache.


plaindocs:
Thanks!

PR #638, opened 2016-06-24
footer.html - Missing <ul>

stonio:


plaindocs:
@lislis I suspect I broke this one. Any preference if we want to add the `UL` or remove the `LI`?

plaindocs:
Thanks!

PR #634, opened 2016-06-22
First version of the SQ add-on doc

bellingard:
@BanzaiMan Here's a first version of the documentation page for the SonarQube add-on. Feel free to give feedback on it.

@dbmeneses and @henryju: as this is markdown, you can just go on the sonarqube.md file and click on "View" to have something readable and give feedback. Thanks.

PR #644, opened 2016-06-27
Sonar source sq addon

BanzaiMan:
To supersede #634.


plaindocs:
The original author manually picked the changes, and updated - so we can close this.

PR #652, opened 2016-06-29
Add -y to 'sudo apt-get install varnish' command

benhutchins:
This fixes the example to make it clear that Travis would need to have the `-y` argument passed when running `apt-get install`, or it'll hang for a response.


plaindocs:
Thanks!

PR #660, opened 2016-07-06
Document how BUNDLE_GEMFILE is defined

BanzaiMan:
Resolves https://github.com/travis-ci/travis-ci/issues/6266

### DIFF
--- a/user/languages/ruby.md
+++ b/user/languages/ruby.md
@@ -251,8 +251,8 @@ engines, hosted service providers and so on.
 ### `$BUNDLE_GEMFILE` environment variable
 
 When `gemfile` is thus defined *and* the file exists in the repository,
-we define the environment variable `$BUNDLE_GEMFILE`, so that `bundle install`
-respects this.
+we define the environment variable `$BUNDLE_GEMFILE`, which `bundle install`
+uses to resolve dependencies.
 
 If you need to work with multiple Gemfiles within a single job, some care is
 required for the alternate `bunlde` invocation.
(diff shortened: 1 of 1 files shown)


BanzaiMan:
@plaindocs Review, please.

user/languages/ruby.md
```diff
@@ -248,6 +248,25 @@ versions](https://github.com/acrmp/chefspec/blob/master/.travis.yml):
 The same technique is often applied to test against multiple databases, templating
 engines, hosted service providers and so on.
 
+### `$BUNLDE_GEMFILE` environment variable
```
plaindocs: A few BUNLDE typos here and there

user/languages/ruby.md
```diff
@@ -248,6 +248,25 @@ versions](https://github.com/acrmp/chefspec/blob/master/.travis.yml):
...
+
+When `gemfile` is thus defined *and* the file exists in the repository,
+we define the environment variable `$BUNLDE_GEMFILE`, so that `bundle install`
+respects this.
+
```
plaindocs: I try to avoid using 'thus'. In this case, you're saying if `gemfile` is defined *and* exists we set $GEMFILE to that value, right?

user/languages/ruby.md
```diff
@@ -248,6 +248,25 @@ versions](https://github.com/acrmp/chefspec/blob/master/.travis.yml):
...
+Either:
+
+1. Define `$BUNLDE_GEMFILE` while invoking `bundle` explicitly
+
+    env BUNDLE_GEMFILE=my_gemfile bundle install
+
+1. Pass `--gemfile=` flag
+
+    bundle install --gemfile=my_gemfile
+
```
plaindocs: Numbered lists usually indicate a sequence, so in a case where you pick on or the other I'd use a bullet list.

Is there any difference in these two methods? If not we could just suggest one?

BanzaiMan:
@plaindocs Thanks for the feedback. I've updated the document.

PR #662, opened 2016-07-11
Fixing errors in network description

mrichmon:
Current docs are incorrect with regard to IPv6 networking.
- When running in legacy VM (e.g. using "sudo: required"), the following
  command output indicates that IPv6 is not available:
  
    $ cat /proc/net/if_inet6
- When running in the container-based environment (both Precise and Trusty),
  the following command output indicates that IPv6 is available:
  
    $ cat /proc/net/if_inet6
    00000000000000000000000000000001 01 80 10 80       lo
    fe800000000000009c859ffffe42cbae 1309 40 20 80     eth0

### DIFF
--- a/user/ci-environment.md
+++ b/user/ci-environment.md
@@ -110,8 +110,8 @@ All [Education Pack](https://education.travis-ci.com/) builds use Container-base
 
 ## Networking
 
-The virtual machines in the Legacy environment running the tests have IPv6 enabled. They do not have any external IPv4 address but are fully able to communicate with any external IPv4 service.
-The container-based, OSX, and GCE (both Precise and Trusty) builds do not currently have IPv6 connectivity.
+The virtual machines in the Legacy environment running the tests do not have IPv6 enabled. They do not have any external IPv4 address but are fully able to communicate with any external IPv4 service.
+The container-based, OSX, and GCE (both Precise and Trusty) builds have IPv6 connectivity.
 
 The IPv6 stack can have some impact on Java services in particular, where one might need to set the flag `java.net.preferIPv4Stack` to force the JVM to resort to the IPv4 stack should services show issues of not booting up or not being reachable via the network: `-Djava.net.preferIPv4Stack=true`.
 


MariadeAnton:
@mrichmon thank you for opening the PR - I really appreciate your help clarifying the ipv6 availability! 

Would it be possible for you to point us to the build logs in which you tested this? That would help us extend the information you fixed in the PR.

Just to add some more info, the legacy infrastructure `sudo: required`  now points to Precise GCE VMs, https://blog.travis-ci.com/2016-04-07-migration-update which should be similar to the VMs available in Trusty build environment, https://docs.travis-ci.com/user/trusty-ci-environment/, does that match with your tests?

Thank you again, Michael!

plaindocs:
@mrichmon @MariadeAnton 

Container environments "have" ipv6 in the sense that the kernel supports it and it is enabled on the loopback interface. but since we do not have a publicly routable ipv6 address/gateway you can't get to the internet over ipv6.

mrichmon:
There are two issues here:
1) The container environments have ipv6. So tests that use the IPv6
loopback address work in the container environment. However, when "sudo:
required" is set tests that use an IPv6 loopback address fail. This is
because requiring sudo switches to a VM environment. From the user
perspective it looks like enabling sudo has disabled IPv6. If IPv6 support
for loopback is enabled in the VM environment this would be fixed.
2) Your docs are wrong. The docs state that IPv6 *is not* available in the
container environment and that IPv6 *is* available in the VM  environment.
In this regard, the docs are exactly reversed from reality. The PR I
submitted corrects the documentation.

I would prefer to have the environment fixed, but at least fixing the docs
may help others avoid needing to debug this issue.

I will send pointers to Travis logs that demonstrate the issue shortly.

Regards,
Michael Richmond

mrichmon:
It is a difficult to follow the various builds I used while debugging this
issue. I have created a separate project to illustrate the network settings
in different travis environments.

The project is: https://github.com/mrichmon/validate-travis-network


   - Build 1:
   - https://travis-ci.org/mrichmon/validate-travis-network/builds/144202338
      - Shows a "trusty" execution environment without sudo.
      - The output of "cat /proc/net/if_inet6" is on line 150.
      - This output indicates that IPv6 is available. ("container"
      environment)
   - Build 2:
      -
      https://travis-ci.org/mrichmon/validate-travis-network/builds/144204863
      - Shows a "trusty" execution environment with sudo.
      - The output of "cat /proc/net/if_inet6" is on line 178.
      - This output indicates that IPv6 is not available. ("VM" environment)
   - Build 3:
      -
      https://travis-ci.org/mrichmon/validate-travis-network/builds/144205785
      - Shows a "precise" execution environment without sudo.
      - The output of "cat /proc/net/if_inet6" is on line 150.
      - This output indicates that IPv6 is not available. ("VM" environment)
   - Build 4:
      -
      https://travis-ci.org/mrichmon/validate-travis-network/builds/144207070
      - Shows a "precise" execution environment with sudo.
      - The output of "cat /proc/net/if_inet6" is on line 171.
      - This output indicates that IPv6 is not available. ("VM" environment)

The PR I submitted is to correct your docs to match the above results.
Specifically to indicate that IPv6 is available in the container
environments, and IPv6 is not available in the VM environments.

A preferable fix is to update your VM environments to enable IPv6. This
would allow IPv6 loopback addresses to work even if the VMs are not
externally routable via IPv6.

Regards,
Michael Richmond

mrichmon:
Hello,
Any further update to this IPv6 network issue? (Specifically, can IPv6 be provided in your VM image to support a localhost IPv6 stack?)

PR #669, opened 2016-07-15
Update indentation to match travis gem output

backspace:
I attended a workshop at Wicked Good Ember where we added Heroku deployments via Travis, which produced a [diff that was mostly full of trivial indentation and quotation changes](https://github.com/backspace/wge-advanced-2016/commit/7f2bd8c6c23255d680c156606a1a3394a82ba120#diff-354f30a63fb0907d4ad57269548329e3).

I submitted a [pull request](https://github.com/ember-cli/ember-cli/pull/6082) to Ember CLI to have the blueprint `.travis.yml` files match the indentation of the output of the Travis gem, but it was rejected because the documentation shows indentation that doesn’t match the gem output.

I suggest that we either update the documentation to show `.travis.yml` examples that match the output of the gem or change the gem to match the indentation of the documentation. If the latter seems like a preferable solution, I can close this and work on that instead.

### DIFF
--- a/user/customizing-the-build.md
+++ b/user/customizing-the-build.md
@@ -60,8 +60,8 @@ You can specify your own script to run to install whatever dependencies your pro
 You can also provide multiple steps, for instance to install both ruby and node dependencies:
 
     install:
-      - bundle install --path vendor/bundle
-      - npm install
+    - bundle install --path vendor/bundle
+    - npm install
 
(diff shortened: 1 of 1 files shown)


backspace:
See some discussion [here](https://travisci.slack.com/archives/dev/p1468609867000655).

PR #676, opened 2016-07-21
fix current page highlighting

lislis:
The old implementation did not consider being directly linked to a page which
led to inconsistent behaviour


plaindocs:
This does not seem to fix the problem for me, in Chrome on Linux. 

screenshot from 2016-07-29 17 08 11

plaindocs:
Also - would it make sense to be able to click the open title to close it? or should that be a separate bug? :-p

lislis:
Oh wow. Seems like there is another component to this that I haven't though about :/
Thanks for properly testing 💃 I'll have another look next week

lislis:
@plaindocs I think I got the cause for the bug with the double highlighting. It should work now, but let me know (as usual) if you encounter another bug

plaindocs:
Thought I'd merged this already, oops. Looks awesome though, thanks!

PR #677, opened 2016-07-23
Added details about when the artifacts are uploaded

cotsog:
i.e. between the `after_success` and `after_script`.

### DIFF
--- a/user/uploading-artifacts.md
+++ b/user/uploading-artifacts.md
@@ -10,6 +10,8 @@ Travis CI can automatically upload your build artifacts to S3. Unless you
 programatically generate unique filenames and folders, artifacts are
 overwritten every build.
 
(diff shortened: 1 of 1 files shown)


user/uploading-artifacts.md
```diff
@@ -10,6 +10,8 @@ Travis CI can automatically upload your build artifacts to S3. Unless you
 programatically generate unique filenames and folders, artifacts are
 overwritten every build.
 
+> Note: Artifacts uploading happens between the `after_success` and `after_script` steps of [the build lifecycle](https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle).
+
```
plaindocs: Is it relevant to say whether this is before or after the optional deploy steps? 

Also might be more direct to say

> Artifacts are uploaded between ... 

or if you prefer

> Travis CI uploads artifacts between...

PR #692, opened 2016-08-11
Added missing _ to before_install

asbjornu:
The `before_install` step was missing an underscore.


plaindocs:
Good catch! Thanks!

PR #693, opened 2016-08-15
add a comment about who gets notified on failing cron jobs

soenkehahn:


user/cron-jobs.md
```diff
@@ -9,7 +9,7 @@ permalink: /user/cron-jobs/
 
 <div id="toc"></div>
 
-Cron jobs run builds at regular scheduled intervals independently of whether any commits were pushed to the repository. When they run, cron jobs always fetch the most recent commit on a particular branch and build the project at that state. Cron jobs can run `daily`, `weekly` or `monthly`, which in practice means up to an hour after the selected time span, and they can also be [skipped](#Skipping-Cron-Jobs). Cron jobs cannot be set to run at specific times. 
+Cron jobs run builds at regular scheduled intervals independently of whether any commits were pushed to the repository. When they run, cron jobs always fetch the most recent commit on a particular branch and build the project at that state. If the build fails, the same notifications get sent that would've been sent on a normal, push-triggered failing job. Cron jobs can run `daily`, `weekly` or `monthly`, which in practice means up to an hour after the selected time span, and they can also be [skipped](#Skipping-Cron-Jobs). Cron jobs cannot be set to run at specific times.
```
plaindocs: Good catch, we should definitely add this info.

This is already a pretty convoluted paragraph. How about we break out the information on notifications to a new paragraph, and say something like

> Cron job builds use the same notification settings as normal push builds.

soenkehahn: Sounds good. I've force-pushed a new version.

plaindocs:
Thanks!

PR #697, opened 2016-08-17
Prefer to use atoum through Composer

Hywan:
atoum PHAR is no longer the best way to use it. This is better to grab it through Composer instead.


plaindocs:
Thanks!

Hywan:
@plaindocs Thank **you** 😃. Do you have an idea when the documentation will be updated?

plaindocs:
Just takes a couple of minutes to propagate.

Hywan:
Thanks :+1:.

PR #699, opened 2016-08-17
Disk size and availability added

joepvd:
The available space was taken with `language: ruby`.

Maybe this should be noted somehow. Would not know how to make this
per-language info clear.


user/ci-environment.md
```diff
@@ -103,6 +103,13 @@ The following table summarizes the differences between the virtual environments:
 <td>1</td>
 <td>~2, bursted</td>
 </tr>
+<tr>
+<td>Disk</td>
+<td>30 Gb (Available: 22 Gb)</td>
```
igorwwwwwwwwwwwwwwwwwwww: Should be uppercase `GB`. `Gb` usually refers to `Gigabit`.

igorwwwwwwwwwwwwwwwwwwww:
Agree that the per-language space should be more directly explained. Even if as a more general statement of "language runtimes will take up some of this space".

igorwwwwwwwwwwwwwwwwwwww:
👍

plaindocs:
Some picky comments:

* would Disk Space be better than Disk? 
* If we guesstimated which runtime uses *most* space, and gave those figures, people are unlikely to care if they have less space. And then would not need to mention Ruby.
* Probably we're not just varying according to runtime, but according to language image (ie, other pre installed dependencies)

BanzaiMan:
I am not a fan of committing these numbers to documents when the build images can be updated at any time and render this document stale and misleading.

Perhaps we can add a default run of `df -h` or some such for this purpose.

plaindocs:
I'm closing this for now, as the numbers are probably out of date, and hope to include some more automated disk space info in the *near* future.

PR #700, opened 2016-08-19
Fixed incorrect use of TRAVIS_PULL_REQUEST.

lennartstolz:
Due to the documentation of environment variables `TRAVIS_PULL_REQUEST` returns either a number of the pull request or `false` if it's not a pull request. Therefore the check `= "true"` won't work as expected.


plaindocs:
Good catch! Thanks!

lennartstolz:
@plaindocs you're welcome.

PR #705, opened 2016-08-23
How do environment variables actually work. WIP

### LINKED ISSUE #691: Consider adding a note about using the -r/--repository option when encrypting environment variables
We often tell the customers to add `-r repo` or `--repository repo` to their `travis encrypt` command when they are having troubles encrypting their environment variables. It might be a bug in our CLI but it could also be caused by a user mistake e.g. not running it in the right directory locally, `--org` vs `--pro`.

To mitigate these issues and perhaps reduce support a bit, I suggest we mention the usage of `-r`/`--repository` option when calling `travis encrypt`.

This information could be mentioned on [the environment variables page](https://docs.travis-ci.com/user/environment-variables/) for example.

Open to discuss further.

Thanks!

### LINKED ISSUE #698: Clarify sensitive information in environment variables
"In one place on the page, it seems like Repository Settings should not be used to set sensitive environment variables, and farther down that page, it looks like you can use Repository Settings to set sensitive environment variables."

https://docs.travis-ci.com/user/environment-variables/

plaindocs:
Fixes 

#698 

Does not yet fix 

#691

### DIFF
--- a/user/environment-variables.md
+++ b/user/environment-variables.md
@@ -4,41 +4,41 @@ layout: en
 permalink: /user/environment-variables/
 ---
 
-A common way to customize the build process is to use environment variables, which can be accessed from any stage in your build process.
+A common way to customize the build process is to define environment variables, which can be accessed from any stage in your build process.
 
 <div id="toc"></div>
 
-When you need to define environment variables you can do so in:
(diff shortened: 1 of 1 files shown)


plaindocs:
OK, I'm gonna merge this for now, and will reopen for 691 if needed.

PR #714, opened 2016-08-31
Add note about %2F requirement in request URL

backspace:
Since %2F is the escaped form of a slash, people are assuming
it can just be converted to a slash. That results in failures,
though, because the slug needs to be treated as a single URL
segment.

### DIFF
--- a/user/triggering-builds.md
+++ b/user/triggering-builds.md
@@ -34,6 +34,8 @@ curl -s -X POST \
   https://api.travis-ci.org/repo/travis-ci%2Ftravis-core/requests
 ```
 
+> The %2F in the request URL is required so that the owner and repository name in the repository slug are interpreted as a single URL segment.
+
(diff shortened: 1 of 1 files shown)


user/triggering-builds.md
```diff
@@ -34,6 +34,8 @@ curl -s -X POST \
   https://api.travis-ci.org/repo/travis-ci%2Ftravis-core/requests
 ```
 
+> The `%2F` in the request URL is required to prevent the owner and repository name in the repository slug from being interpreted as different URL segments.
```
plaindocs: Is it clearer to say

> The `%2F` in the request URL is required so that the owner and repository name in the repository slug are interpreted as a single URL segment.

do you think?

backspace:
I agree, @plaindocs, thanks so much for the wordsmithing. Awkward phrasing is one of my areas of expertise 😆

plaindocs:
I think we're ALL good at that. I just have a lot of practice in catching myself doing it. I'm gonna :ship: :it: !

PR #720, opened 2016-09-06
Add link to Django Webhook Signature Verifier

andrewgross:
Adds a link that shows a sample Python implementation inside a Django view, similar to the Sinatra app.  Feel free to take the code and stick it in a repo like the Sinatra app as well.


plaindocs:
Nice! Thanks @andrewgross ! 

CC @joshk

PR #721, opened 2016-09-06
remove warning text

pestrada:
The script is already fixed.
See: https://github.com/pestrada/android-tdd-playground/blob/master/.travis.yml


plaindocs:
Thanks!

PR #736, opened 2016-09-21
Rename "standard" environment to "sudo-enabled"

igorwwwwwwwwwwwwwwwwwwww:
The naming of "standard" is very misleading, especially because it is not the _default_ environment.


igorwwwwwwwwwwwwwwwwwwww:
Bump

PR #740, opened 2016-09-28
Replace 'rbx-2' with 'rbx'.

brixen:
Rubinius no longer supports any version of the 2.x epoch. Use of 'rbx' is
encouraged so that the current stable release is tested.

Rubinius releases as frequently as once or more a day and any attempt to
'lock' to a Rubinius version significantly disadvantages a project by not
using Rubinius fixes and enhancements.

It also disadvantages Rubinius because time is wasted dealing with outdated
issues for which fixes have already been released.

### DIFF
--- a/user/languages/ruby.md
+++ b/user/languages/ruby.md
@@ -53,10 +53,6 @@ environment guide](/user/ci-environment/).
 
 If you don't specify a version, Travis CI will use MRI 1.9.3 as the default.
 
-When using Rubinius, there's currently an issue with selecting the correct version
-in RVM in our build environment, but only when specifying `rbx` as your version. As
-a workaround, specify `rbx` instead.
-
 ### Using `.ruby-version`
 
 If the ruby version is not specified by the `rvm` key as described above, Travis CI


iainbeeston:
Do these instructions work for the default travis environment? (Doesn't rubinius require "dist: trusty" too?)

brixen:
@iainbeeston that's a good point. They *could* work on Precise if Travis pre-installs the required libraries to make LLVM > 3.5 available. Otherwise, I have not been able to build Rubinius binaries for Precise that will statically link the required libc++ and supporting libraries.

Precise will EOL in a short few months, and given the time it's taken for Trusty to be the default, the process cannot start soon enough. So perhaps this could be an impetus for that.

I'd be open to adding caveats on Rubinius requiring Trusty if that's desirable.

user/languages/ruby.md
```diff
@@ -55,7 +55,7 @@ If you don't specify a version, Travis CI will use MRI 1.9.3 as the default.
 
 When using Rubinius, there's currently an issue with selecting the correct version
 in RVM in our build environment, but only when specifying `rbx` as your version. As
-a workaround, specify `rbx-2` instead.
+a workaround, specify `rbx` instead.
```
brixen: @randycoulman ah, good catch. :) I'll update.

pnomolos:
These instructions currently aren't working for us on `dist: trusty` - https://travis-ci.org/seamusabshere/upsert/jobs/166126394

brixen:
@pnomolos yep, there's a bug in RVM that has been long-standing (years) and intermittent. You can see our Travis Canary results here https://travis-ci.org/rubinius/travis-canary/builds/163446822

pnomolos:
@brixen Has any though been given to moving from RVM over to something like rbenv?  Maybe ref: https://github.com/rbenv/ruby-build/issues/42#issuecomment-205511021

karinohka66:
https://github.com/rubinius/docs-travis-ci-com/commit/c6c444445b8468422f0c49e327faefdc965ff6c7

plaindocs:
Closing this as the issue seems fixed by #1534

PR #743, opened 2016-09-29
Document workaround for slow JRuby installation on GCE

BanzaiMan:
One should install HAVEGED.

### DIFF
--- a/user/languages/ruby.md
+++ b/user/languages/ruby.md
@@ -104,6 +104,28 @@ JRuby into account. Most of popular C extensions these days also have Java
 implementations (json gem, nokogiri, eventmachine, bson gem) or Java
 alternatives (like JDBC-based drivers for MySQL, PostgreSQL and so on).
 
+### JRuby: Installation takes a long time on Standard and Trusty environments
+
+In the [Standard and Trusty
+environments](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments),
+installing JRuby can take several minutes.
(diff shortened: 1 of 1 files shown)


BanzaiMan:
One could also argue that this should be automatic on JRuby installation.

perlun:
Thanks for that. I tried this now in one of our repos. But it still takes about 2 minutes to install JRuby with this fix, and the `sudo`-based infrastructure means that the build starts slower as well.

I noted that this list (from the build output):

```
Pre-installed Ruby versions
jruby-9.1.2.0
ruby-1.9.3-p551
ruby-2.0.0-p648
ruby-2.1.9
ruby-2.2.5
ruby-2.3.1
```

...doesn't contain `jruby-9.1.5.0`. Maybe that would be the quick and easy fix, that you add `jruby-9.1.5.0` to the list of preinstalled ruby versions? Then it should be instant to `rvm use` that version in the builds.

perlun:
Any progress on this? JRuby 9.1.5 builds are still, regretfully, taking _minutes_ longer than builds on other (supported) jrubies.

perlun:
Still nothing? This is **really really** frustrating, since it slows _all_ our builds down with 2 minutes. JRuby 9.1.5.0 was released a long time ago, it's really sad that Travis don't seem to care about the JRuby community.

perlun:
9.1.6.0 is out now, so 9.1.5.0 is no longer so interesting. But good support for 9.1.6.0 is really, really important.

PR #746, opened 2016-10-03
add documentation for upgrading gcc and clang

sxlijin:
Getting started with a modern C++ project is _painful_. The nature of Ubuntu means that the default toolchain is ancient and you have to hunt through documentation, discover that none of this is actually documented, and then end up at blog posts that sort of describe how to fix this. _Then_ you learn about the source and package whitelists and cross your fingers that whatever it is that you're looking to bring in is whitelisted, and the rabbit hole just gets deeper and deeper...

Travis can do better!


user/languages/cpp.md
```diff
@@ -93,3 +93,134 @@ before_install:
...
+  include:
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-4.9
+      env:
+         - MATRIX_EVAL="CXX=g++-4.9"
```
CaseyCarter: Why not `CC=gcc-4.9`? (and on 126 and 136.)

sxlijin: I didn't test this, but since `g++-4.9` has to be explicitly installed as a package, I'd be surprised if `gcc-N` was available by just installing `g++-N`. Perhaps it should be the case that `CC` should be explicitly unset in `MATRIX_EVAL` though.

CaseyCarter: The g++-N package depends on gcc-N, e.g. `apt show g++-4.9`:

<pre>
Package: g++-4.9
Version: 4.9.4-2ubuntu1~16.04
Priority: optional
Section: devel
Source: gcc-4.9
Maintainer: Ubuntu Core developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Debian GCC Maintainers <debian-gcc@lists.debian.org>
Installed-Size: 33.6 MB
Provides: c++-compiler, c++abi2-dev
Depends: gcc-4.9-base (= 4.9.4-2ubuntu1~16.04), gcc-4.9 (= 4.9.4-2ubuntu1~16.04), libstdc++-4.9-dev (= 4.9.4-2ubuntu1~16.04), libc6 (>= 2.14), libcloog-isl4 (>= 0.17), libgmp10 (>= 2:5.0.1~), libisl15 (>= 0.15), libmpc3, libmpfr4 (>= 3.1.3), zlib1g (>= 1:1.1.4)
Suggests: g++-4.9-multilib, gcc-4.9-doc (>= 4.9), libstdc++6-4.9-dbg (>= 4.9.4-2ubuntu1~16.04)
Download-Size: 21.3 MB
APT-Manual-Installed: yes
APT-Sources: http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu xenial/main amd64 Packages
Description: GNU C++ compiler
 This is the GNU C++ compiler, a fairly portable optimizing compiler for C++.
</pre>

user/languages/_includes/modern-c-cpp-toolchain.md
```diff
@@ -0,0 +1,130 @@
...
+  include:
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-4.9
+      env:
+         - MATRIX_EVAL="CC=gcc-4.9 && CXX=g++-4.9"
```
sxlijin: @CaseyCarter added `CC` to each GCC `MATRIX_EVAL` statement. Also moved the entire section that I'm adding out to another file so it can be included in both `c.md` and `cpp.md`.

user/languages/_includes/modern-c-cpp-toolchain.md
```diff
@@ -0,0 +1,130 @@
+### Modern C/C++ and Toolchain Versioning
+
+If you need a more modern toolchain or more granular control over your tools, you will have to handle that yourself. Fortunately, the community has already whitelisted the [repositories](https://github.com/travis-ci/apt-source-whitelist/blob/master/ubuntu.json) and [packages](https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise) necessary for most such upgrades.
```
plaindocs: > for most such upgrades

PICKY, I'd delete the "most"

user/languages/_includes/modern-c-cpp-toolchain.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### GCC on Linux
+
+Ubuntu 12.04 shipped with GCC 4.6.3 and Ubuntu 14.04 shipped with GCC 4.8.2. If you need a more recent version (if you are working with C++11, you *must* do this; support for C++11 was not finished until GCC 4.9), you will want to follow these directions:
```
plaindocs: >  (if you are working with C++11, you *must* do this; support for C++11 was not finished until GCC 4.9)

I'm not 100% clear on what *this* is that you *must* do:

plaindocs:
@Lyoness this looks mostly fine to me, couple of comments excepted, could you let me know if it is ok to merge from a technical persepctive?

user/languages/_includes/modern-c-cpp-toolchain.md
sxlijin: That's a fair point, particularly since 4.8.1 is marked as the first feature-complete implementation of the C++11 standard. Notably, however, [support for `<regex>` does not exist until 4.9](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53631).

Perhaps something like this?

```
(note: C++11 is feature-complete in 4.8.1, but [support for `<regex>` does not exist until 4.9](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53631))
```

user/languages/_includes/modern-c-cpp-toolchain.md
sxlijin: That was actually intentional, and it remains true today: clang-3.9 is not available, and you also can't version cmake (the latest `cmake` that you can upgrade to is 3.2.3, which was [released in June 2015](https://blog.kitware.com/cmake-3-2-3-available-for-download/)).

Perhaps instead a note about how to request that packages be be whitelisted?

user/languages/_includes/modern-c-cpp-toolchain.md
agauniyal: @plaindocs  c++11 dates back 2011, while we're living in 2017 which means we should be using c++14 at least. However, c++11 is a definite must here.

user/languages/_includes/modern-c-cpp-toolchain.md
BanzaiMan: > Fortunately, the community has already whitelisted…

I don't think this works here. Who listed the packages is probably not important to the reader, but just the fact that the sources and packages are available for use. (Who is "the community"?)

The instructions to request new sources/packages are given when you run jobs on containers, but I now see that the documentation https://docs.travis-ci.com/user/installing-dependencies does not contain that. We should fix it.

user/languages/_includes/modern-c-cpp-toolchain.md
BanzaiMan: A few comments:

1. I think we should use present tense here: Ubuntu 12.04 _ships_ with … While highly unlikely, these versions *could* change if Ubuntu team updates the packages.
1. While I agree with the sentiment, I don't think this parenthetical remark belongs here.
1. This YAML fragment is not *instructions* per se. And you don't have to have multiple jobs via `matrix.include` to realize C++-11 support, either. I agree with @plaindocs here; this is not a *must*.

I am guessing that, here, _this_ means "realizing the C++-11 support". I think this subject should be treated in a separate section.

user/languages/_includes/modern-c-cpp-toolchain.md
BanzaiMan: Having an auxiliary environment variable `MATRIX_EVAL` gives a false impression that this is mandatory when it is not. You are trying to consolidate `script` sections that have different values depending on the OS, but that is a secondary concern of this document.

user/languages/_includes/modern-c-cpp-toolchain.md
```diff
@@ -0,0 +1,130 @@
...
+          packages:
+            - g++-6
+      env:
+        - MATRIX_EVAL="CC=gcc-6 && CXX=g++-6"
+
+before_install:
+    - eval "${MATRIX_EVAL}"
+```
+
+#### GCC on OSX
```
BanzaiMan: There is no such thing as "OSX". Use "OS X" and/or "macOS". I am not sure what Apple's guideline says about the term that means "all OS versions up to and including 10.12.x".

user/languages/_includes/modern-c-cpp-toolchain.md
```diff
@@ -0,0 +1,130 @@
...
+      env:
+        - MATRIX_EVAL="brew install gcc && CC=gcc-6 && CXX=g++-6"
+
+before_install:
+    - eval "${MATRIX_EVAL}"
+```
+
+#### LLVM
+
+Be warned that the LLVM-Travis ecosystem [has faced some stability problems for over a year now](https://github.com/travis-ci/apt-source-whitelist/issues/156).
```
BanzaiMan: I don't think this passage belongs here. The issue does not help the reader on their attempt to set up builds with LLVM.

user/languages/_includes/modern-c-cpp-toolchain.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### CMake
+
+To upgrade `cmake` (note that the most recent verison of `cmake` that has been backported is 3.2), add the following to each `apt` addon which requires it:
+
+```yaml
+addons:
+  apt:
+    sources:
+      - george-edison55-precise-backports
```
BanzaiMan: Q: Does this source have Trusty packages, too?

user/languages/_includes/modern-c-cpp-toolchain.md
agauniyal: I get it how must could translate into 'no pre c++11 support' but what I and @sxlijin (probably) meant was - you 'should definitely be using' c++11 now. I strongly support this argument because a lot of students I know have fallen back to not using newer(and better) language constructs just because their support was lacking somewhere in their build cycle which is unacceptable.

BanzaiMan: @agauniyal Such nuances of language usage might be better addressed elsewhere. This is what I am suggesting with a "C++-11 support" section (which is yet to be written).

user/languages/_includes/modern-c-cpp-toolchain.md
sxlijin: To the best of my knowledge, [no](https://github.com/travis-ci/apt-source-whitelist/blob/master/ubuntu.json#L147-L151).

sxlijin: Did some more digging into this. It seems that Nathan Osman (george-edison55 over at Launchpad) has [a separate PPA](https://launchpad.net/~george-edison55/+archive/ubuntu/cmake-3.x) for more recent backports of `cmake`, including one to Trusty.

user/languages/_includes/modern-c-cpp-toolchain.md
sxlijin: The motivation was to, in some shape or form, standardize how the C++ compiler should be specified. Ideally I would've done

```
env:
  - CC=gcc-4.9
  - CXX=g++-4.9
```

but Travis overrides `CC` and `CXX` after the environment is configured per the build matrix, which forces the user to override them *again*, and the earliest point at which they can do so is in `before_install`. I'm reluctant to PR into Travis itself to support additional values for `compiler`, e.g. `gcc-4.9`, `gcc-5`, `gcc-6`, because (in my mind) it would cause a plethora of issues with ecosystem and images.

That all being said, I'm not exactly sure what you mean by "secondary concern" - could you clarify that, and what a preferable alternative would look like?

user/languages/_includes/modern-c-cpp-toolchain.md
sxlijin: @BanzaiMan I take it that you prefer that this instead point the reader to `installing-dependencies` then?

user/languages/_includes/modern-c-cpp-toolchain.md
sxlijin: 1. Is that actually possible? The 12.04 images have already been released and mirrored thoroughly extensively. Even in the scenario where the base packages are bumped, they wouldn't just drop them into 12.04 itself, but rather release a new version, [like they did with 12.04.3](https://wiki.ubuntu.com/PrecisePangolin/ReleaseNotes/ChangeSummary/12.04.3).

1. I'm loath to add another section in its entirety to address C++11 support specifically, since (1) this entire PR's purpose is to add a section titled "Modern C/C++ and Toolchain Versioning" and (2) there really isn't much to say about supporting C++11 *specifically* besides what's here in this parenthetical. Supporting C++14 and C++17 are different stories entirely.

    Also, @BanzaiMan I suggested above in response to @plaindocs that the parenthetical be replaced with the following - thoughts (this also addresses the lack of clarity w.r.t. "this")?

    > (note: C++11 is feature-complete in 4.8.1, but [support for `<regex>` does not exist until 4.9](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53631)

1. The "follow these directions" language can be adjusted, yes.

user/languages/_includes/modern-c-cpp-toolchain.md
BanzaiMan: Thanks for your research. If you open a PR and ping me, I'll look into it.

user/languages/_includes/modern-c-cpp-toolchain.md
BanzaiMan: 1. We use Ubuntu minor updates when we provision new images. For example, the current container image is 12.04.5.
1. I see what you are saying. How about eschewing the use of vague "modern" in favor of more explicit phrases, say "Adding C++-11 support"?

user/languages/_includes/modern-c-cpp-toolchain.md
sxlijin: Opened at travis-ci/apt-source-whitelist#353.

sxlijin:
@BanzaiMan language has been revised - let me know if you have any objections.

user/languages/_includes/modern-c-cpp-toolchain.md
sxlijin: I see the OS X vs. macOS question is still open... (#766).

plaindocs:
Revisiting this. Could you remove the `Gemfile.lock` from the PR unless there is a reason I'm missing to include it? Thanks.

plaindocs:
@BanzaiMan this look ok to you
@sxlijin I think we're just waiting on the conflict to merge, and I don't think there is any need for the Gemfile.

sxlijin:
Whoops, sorry for the delay! Completely missed that I accidentally committed Gemfile changes.

Let me know if you want me to squash.

sxlijin:
@BanzaiMan @plaindocs bump

joepvd:
This text has been unclear for at least one human: https://stackoverflow.com/questions/44213719/how-to-use-matrix-eval-to-enable-a-build-matrix-row-based-on-a-compiler

plaindocs:
Interesting. This is a thorny subject -- I'll add taking another look to my todo list. Thanks for the ping!

user/languages/_includes/c11-cpp11-and-beyond-and-toolchains.md
```diff
@@ -0,0 +1,179 @@
...
+
+    - os: osx
+      osx_image: xcode8
+      env:
+        - MATRIX_EVAL="brew install gcc5 && CC=gcc-5 && CXX=g++-5"
+
+    - os: osx
+      osx_image: xcode8
+      env:
+        - MATRIX_EVAL="brew install gcc && CC=gcc-6 && CXX=g++-6"
```
AraHaan: This fails with brew now as it will install gcc 7.2.0 Do you mean gcc6?

In this case the docs need another updating.
```
/Users/travis/.travis/job_stages: line 57: g++-6: command not found

The command "$CXX -O3 -std=c++14 bin2c.cpp -lstdc++fs -o "bin2c"" exited with 127.
```

AraHaan:
@sxlijin @BanzaiMan Just 17 hours ago I spotted these issues. and because of these issues I had to modify my .travis.yml to bypass them in some ways. Also I had to disable all of my clang steps because they would end up pointing to gcc 4.8's stlib but things like C++14/17 will not work for it. This might mean also upgrading gcc in the case of clang to bypass it to have it temporarily uninstall gcc and then reinstall it to bypass the clang issues. Otherwise idek.

user/languages/_includes/c11-cpp11-and-beyond-and-toolchains.md
BanzaiMan: I'm guessing that Homebrew formula is updated to install a new major version.

sxlijin: I would presume that yes, you probably have to `brew install gcc-6` or something. Unfortunately I have stuff on my plate right now and don't have the time to come around and fix this - @AraHaan, maybe you could PR a fix?

I think I did try doing `brew install gcc-6` when I submitted this PR, and unfortunately it failed because the latest version was not specifically tagged by homebrew. Perhaps this is also an issue to raise with Homebrew?

AraHaan: I think it's ``brew install gcc@6``.

AraHaan:
https://github.com/travis-ci/docs-travis-ci-com/pull/1432

user/languages/_includes/c11-cpp11-and-beyond-and-toolchains.md
sxlijin: Awesome!

Unfortunately I just *know* that I'm going to get notifications in the future when GCC 8 is released and Homebrew points `gcc` to GCC 8 and not 7...

PR #750, opened 2016-10-06
add navigation overview class

lislis:
Looks like this:
screen shot 2016-10-06 at 14 10 14

add `is-overview` to either the `li` or `a` tag in the nav sidebar works


plaindocs:
Thanks @lislis !

PR #751, opened 2016-10-06
Clarify absolute and relative paths in caching

joepvd:


user/caching.md
```diff
@@ -147,7 +147,7 @@ cache:
   - $HOME/.m2
 ```
 
-As you can see, it is also possible to use environment variables in the directories.
```
plaindocs: I know this isn't part of your change, but while we're here, can we change this to something like:

> You can use environment variables as part of the directory path

So it does not sound like something you can do inside the directory.

user/caching.md
```diff
@@ -147,7 +147,7 @@ cache:
...
+As you can see, it is also possible to use environment variables in the directories. Paths that do not start with a `/` are relative to `$TRAVIS_BUILD_DIR`. When, after possible variable expansion, the path starts with a `/`, that absolute path is used. Please be aware that the `travis` user needs to have write permissions in this directory.
```
plaindocs: And here maybe something like:

After variable expansion, paths that

> * do *not* start with a `/` are relative to `$TRAVIS_BUILD_DIR`
> * do start with a `/` are absolute

To create the cache, the `travis` user needs to have write permission to the directory.

(I'd say *to* not *in* -- what do you think?)

plaindocs:
Looks good to me. :-)

PR #762, opened 2016-10-18
Added MySQL to the list of database NOT pre-installed on Trusty

cotsog:
@meatballhat: can you confirm? Thanks!


joepvd:
MySQL 5.6 is installed by default in the current sudo enabled Trusty image.

PR #772, opened 2016-10-22
Add note that users must use the MySQL 5.6 instructions for Trusty

mmb:
I added Docker to my builds which switched me to the Trusty image. It took me a while to figure out why MySQL stopped working with the default MySQL instructions. This note tells users that they must install MySQL themselves on Trusty.


plaindocs:
Thanks!

PR #773, opened 2016-10-25
First steps towards adding a style guide for the docs

plaindocs:
WIP


plaindocs:
@acnagy want to take a look and see if you think anything is missing?

STYLE.md
```diff
@@ -0,0 +1,90 @@
...
+or
+
+```markdown
+The [link](http://www.example.com "Example URL") in the text
+
+```
+
+### Code Inline
+
+All function names, filenames, etc should be marked with `back-ticks`.
```
joepvd: Maybe this can be more precise.

With that the database PostgreSQL should not be backticked, but the `psql` application should.

plaindocs:
I think there is enough here to be useful, @MariadeAnton wanna :ship: :it: from :portugal: ?

PR #783, opened 2016-11-02
Include JDK 8 Build workaround in public docs

Vogel612:
As requested [here](https://github.com/travis-ci/travis-ci/issues/5081#issuecomment-257889136) I'm just adding this to the public docs for other users


Vogel612:
This also shows how to fix issues with JavaFX applications, especially since the discussed other fix from travis-ci/#3259 is not working anymore because it now causes a prompt during installation of the jdk

PR #786, opened 2016-11-04
Updated docs to match current firefox addons

jrbenny35:
This updates the firefox addons doc to match my patch adding firefox developer and nightly versions.


BanzaiMan:
Thank you!

PR #787, opened 2016-11-06
Update Code Climate ruby-test-reporter steps

maxjacobson:
The setup is a little more involved now.

I removed the part about `WebMock` because the coverage data is now sent outside of the test context.

### DIFF
--- a/user/code-climate.md
+++ b/user/code-climate.md
@@ -27,13 +27,23 @@ free for open source.
 For Ruby projects you need to add a library to your Gemfile called
 [`code-climate-reporter`](https://github.com/codeclimate/ruby-test-reporter):
 
-    gem "codeclimate-test-reporter", group: :test, require: nil
+    group :test do
+      gem "simplecov"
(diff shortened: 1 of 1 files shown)


user/code-climate.md
```diff
@@ -104,22 +114,6 @@ You can find additional setup, installation, and troubleshooting information her
 
 ## Common Problems
 
-#### I get an error transmitting coverage results
```
BanzaiMan: I am curious to know why this question is removed.

maxjacobson: Of course. We used to run our reporting within the test context, which meant that [WebMock](https://github.com/bblimke/webmock) users would inadvertently prevent us from making the network requests that submits the coverage data to our services. The workaround was to configure WebMock to allow network requests to our domain.

Now that we submit test coverage in a separate step, WebMock won't interfere with our network requests.

maxjacobson:
Btw, I'm going to close and re-open this a little later today. I noticed one more change I think I'm going to need to make, but I want to double check before you ship this change. I appreciate your patience.

PR #790, opened 2016-11-08
Update NAT IP addresses

meatballhat:
<del>A potential problem with this update is that, with the rollout of replacement container-based ubuntu precise capacity, it may not make sense to call out a precise/trusty split.</del>

(already collapsed 'em)


plaindocs:
LOL at "Appease the Coala" they should definitely make tshirts.

PR #794, opened 2016-11-10
Update Code Climate Docs

GordonDiggs:
We've released a new version of our Ruby test reporter, that works differently than it used to. Additionally, we are converging on keeping all of our test coverage documentation up to date at our documentation site, so this moves other languages there as well.

This replaces #787


GordonDiggs:
Thank you @plaindocs!

PR #808, opened 2016-11-17
Update trusty-ci-environment.md

grooverdan:
As tested, and according to https://blog.travis-ci.com/2016-11-08-trusty-container-public-beta/, group: beta isn't needed.


meatballhat:
@grooverdan Thank you! :heart:

PR #817, opened 2016-11-22
Prepare build env update for GCE group: stable promotion

meatballhat:


user/build-environment-updates/2016-12-01.md
```diff
@@ -0,0 +1,58 @@
...
+
+## 2016-12-01
+
+### Schedule
+
+[2016-12-01 15:00 UTC](http://everytimezone.com/#2016-12-1,180,cn3)
+
+### Platforms
+
+This update applies to fully-virtualized Ubuntu Trusty and Precise.
```
BanzaiMan: The term we are using in https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments is "sudo-enabled".

user/build-environment-updates/2016-12-01.md
```diff
@@ -0,0 +1,58 @@
...
+
+- Include at least one JDK and the `jdk_switcher` tool in all Trusty stacks.
+- Preinstall ruby 2.2.5 for use with build caching and deployment tools.
+- Force removal of `~travis/.pearrc` at the end of image baking to ensure PHP is
+  less confused.
+
+### Details
+
+This update is a promotion of `group: edge` to `group: stable`.  We will retain
+the previous `group: stable` as `group: legacy` for at least a month following
```
joepvd: Awesome to keep the current stable images around. Afraid that [the word _legacy_](https://docs.travis-ci.com/user/migrating-from-legacy/) might cause a confusion or two. `oldstable` or `deprecated` would be other options.

user/build-environment-updates/2016-12-01.md
```diff
@@ -0,0 +1,58 @@
...
+this change.  **NOTE**: When targeting Trusty, some `language:` declarations may
+result in fewer features than are listed here, as they may be routed to the
+"minimal" stack image.
```
joepvd: Think this could be more concrete: Which languages are affected?
And less concrete: Some way to avoid mentioning image stacks? Or should this be elaborated on someplace else?

user/build-environment-updates/2016-12-01.md
meatballhat: I'm good with `deprecated` :+1:

user/build-environment-updates/2016-12-01.md
meatballhat: @joepvd How about I drop the note entirely? 😺

BanzaiMan:
@meatballhat Is it too much to get yarn on Precise? I know we are trying not to mess with Precise (for a good reason), but I do wonder how much cognitive burden we'd put on our users and support team when we have to remember that it is only available for Trusty.

jasonmp85:
> I don't know potential negative effects to our users

I can't speak for others, but it broke our build: citusdata/citus@74d89c1ba991e8652bf4a93cc820e107cc11e0ca

This is largely due to the fact that we had some code to use a version-specific `psql` binary but didn't bother to upgrade the `libpq` package to the latest version. PostgreSQL 9.6 changed logic around `CONTEXT` display, and we used a `psql` setting to reinstate the old behavior; however, using a 9.5 `psql` does not understand that new setting, but the underlying `libpq` still needs to be notified to engage the old behavior.

PR #824, opened 2016-11-25
Quotes around version numbers

mienaikage:
Got caught out when I put in `- 2016.10` and the trailing 0 went missing!


BanzaiMan:
Thanks

Issue #821, opened 2016-11-23
Notifications: always|never|change


svenfuchs:
The documentation says that:

> Note: Items in brackets are placeholders. Brackets should be omitted.
-- https://docs.travis-ci.com/user/notifications/#Email-notifications

So people remove the brackets, and keep `always|never|change`

We should add that they're also supposed to pick one of the values.

PR #835, opened 2016-11-29
Clarify notification options

plaindocs:
Fix for #821

We only actually mention the options in the email notifications though, so I'll take a look at rest of the file, for Slack etc as well when I have a minute.

FAO @svenfuchs


plaindocs:
@svenfuchs I've simplified this down a bit - take a :eyes:

PR #838, opened 2016-11-29
Fix code climate Python link

samuela:


MariadeAnton:
Thank you, Samuel!

samuela:
@MariadeAnton no problem!

PR #852, opened 2016-12-05
Clarify how multiple caching mechanism works

BanzaiMan:
The previous example with

```yaml
cache:
  - bundler
  - pip
```

does not work.


user/caching.md
```diff
@@ -243,23 +243,39 @@ Use one of the following ways to access your cache and delete it if necessary:
...
+```
+
+This is another example; a Rust repository caching cargo and Ruby gems:
+
+```yaml
+language: rust
+cache:
+  cargo: true
+  directories:
+    - vendor/bunlde
```
BanzaiMan: Good catch!

PR #864, opened 2016-12-09
Add guides on how to disable yarn

nazieb:
The example case is: if we're building a NodeJS app for AWS Lambda (which only supports up to v4.3).

On development machines we're using node v6 and yarn, but to deploy the app into Lambda we need to use node v4.3 in the `.travis.yml` file and force the usage of `npm install` since some of the dependencies will fail if installed using yarn.

This is one of those libraries:
```
error sol-redis-pool@0.3.2: The engine "node" is incompatible with this module. Expected version ">= 4.5.0".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```


plaindocs:
Nice one!

PR #879, opened 2016-12-16
Removed -e flag from docker login command

remojansen:
If you add the -e flag you get the following warning:

> Flag --email has been deprecated, will be removed in 1.13.


plaindocs:
Thanks for that! https://docs.docker.com/engine/deprecated/#e-and-email-flags-on-docker-login

PR #884, opened 2016-12-17
Fix installation step documentation

neoascetic:
I guess you must use `install: false` instead of `install: true` to skip it

### DIFF
--- a/user/customizing-the-build.md
+++ b/user/customizing-the-build.md
@@ -73,7 +73,7 @@ When one of the steps fails, the build stops immediately and is marked as [error
 You can skip the installation step entirely by adding the following to your `.travis.yml`:
 
 ```yaml
-install: true
+install: false
 ```
 
 ## Customizing the Build Step


neoascetic:
Hmm, have I chosen a wrong base branch?

grooverdan:
looks like it. everything else targets the gh-pages branch (the default gh branch)

BanzaiMan:
The correct branch is `gh-pages`. We want to move it to `default` eventually (and update the `default` branch). However, the switch will invalidate the outstanding PRs, so we have not done anything about that yet.

neoascetic:
I've fixed it

BanzaiMan:
Has this configuration been tested anywhere? I don't think this works.

If `install` or `script` value is `false` in the configuration, `travis-build` will see it as `nil`, and will use the default value.

See https://travis-ci.org/BanzaiMan/travis_production_test/builds/185593888#L151 and its configuration https://github.com/BanzaiMan/travis_production_test/blob/7c155ae3b11a3e38c89e8294e134f2cc0f48a38b/.travis.yml#L5.

I'm closing this.

neoascetic:
Oh, sorry :D 

The principle of least astonishment is definitely broken for me in this case. That's why I haven't even tested the changes.

PR #885, opened 2016-12-19
document before_cache in stages

grooverdan:
as discovered in travis-ci/travis-ci/issues/6396


grooverdan:
apparently incorrect as before_cache is at the end before the cache is updated.

plaindocs:
Good catch! Thanks!

grooverdan:
Thanks @BanzaiMan for fixing this this up in 72821f1c064ff0e4952e70d0fec22d7fc0f1f1c7

PR #896, opened 2016-12-27
[FIX] Quote encapsulate encryption value

lasley:
This is a simple PR:

* Encapsule encrypt string value in quotes in order to allow special chars

Alternatively, I could add another example block slightly below this one including some special chars & the quotes. 

I figured the more concise would be better, but I'll leave that decision to you maintainers & will make the change if requested 😄


plaindocs:
Thanks! :-) No need for the second example I think.

PR #897, opened 2016-12-27
Add GitHub Pages deployment docs

timwis:
Adds documentation for travis-ci/dpl#501 :tada:

/cc @webknjaz @BanzaiMan


_data/deployments.yml
```diff
@@ -12,6 +12,7 @@ Deis: "/user/deployment/deis/"
 Engine Yard: "/user/deployment/engineyard/"
 ExoScale: "/user/deployment/exoscale/"
 GitHub Releases: "/user/deployment/releases/"
+GitHub Pages: "/user/deployment/pages"
```
webknjaz: Nit: Quotes aren't really needed in YAML

webknjaz: I would place it before Releases to keep the alphabet order

user/deployment/pages.md
```diff
@@ -0,0 +1,48 @@
+---
+title: GitHub Pages Deployment
+layout: en
+permalink: /user/deployment/github-pages/
```
webknjaz: Shouldn't this match URI, declared in YAML config above?

user/deployment/pages.md
```diff
@@ -0,0 +1,48 @@
...
+token](https://help.github.com/articles/creating-an-access-token-for-command-line-use/)
+and set the deployment provider details in `.travis.yml`.
+
+For a minimal configuration, add the following to your `.travis.yml`:
+
+```yaml
+deploy:
+  provider: pages
+  skip_cleanup: true
+  github_token: $GITHUB_TOKEN # Set in travis-ci.org dashboard
```
webknjaz: Alternatively, encrypt using travis CLI tool

user/deployment/pages.md
webknjaz: It looks fixing this will fix the build

_data/deployments.yml
timwis: Good point on the order! I'll leave the quotes just to be consistent with the rest of the existing document though

user/deployment/pages.md
timwis: Doh! thx

user/deployment/pages.md
timwis: Yeah, I mention that below ("encrypted variables"). Hopefully that's good enough? Dashboard seems like better practice anyway -- more conducive to forking.

webknjaz: I'm not sure about security of env vars in dashboard. I didn't test it, but I have an idea about how to steal someone else's env vars via PR. That it why encrypted vars look much more convenient to me  in terms of security

webknjaz:
LGTM

user/deployment/pages.md
lasley: Yup it's 100% possible to steal environment variables with a PR. 

I'm curious how the secure helps though? The environment variables are still exposed to resources being built, so similar strategies would work for both the secure vars and environment variables.

webknjaz: Yes, but `deploy` section is not being executed for PRs, so it is not possible to retrieve that data this way: the secured values won't be available at install and script steps :)

timwis: Holy cow..that's a huge hole in the system. I think I need to change a lot of travis configs now..

joepvd: Encrypted environment variables are not accessible for PR builds. This goes for both the `secure` variables in `.travis.yml`, as well as the secret ones defined through the web page.

webknjaz: Alright, I've done some testing. It seems it's impossible to steal env var by outputting them in PR. Still, if one get's merged this may be an issue :)

_data/deployments.yml
```diff
@@ -11,6 +11,7 @@ CloudFoundry: "/user/deployment/cloudfoundry/"
 Deis: "/user/deployment/deis/"
 Engine Yard: "/user/deployment/engineyard/"
 ExoScale: "/user/deployment/exoscale/"
+GitHub Pages: "/user/deployment/pages"
```
webknjaz: I think, the trailing slash `/` is missing here

user/deployment/pages.md
lasley: @webknjaz Good point, I missed the fact that this is in deploy.

@joepvd I'm not sure what you mean by the secrets not being exposed. Here is a sample repo and some sample builds that prove otherwise. You can see both the `SECRET` & `SECRET_DASHBOARD` variables being echoed in the `script`, `after_script`, and `after_success` keys:

* [Repo](https://github.com/lasley/test-travis-pr-vars)
* [Master Build](https://travis-ci.org/lasley/test-travis-pr-vars/builds/187510996)
* [Branch Build](https://travis-ci.org/lasley/test-travis-pr-vars/builds/187511229)
* [PR Build](https://travis-ci.org/lasley/test-travis-pr-vars/builds/187511938)

Obviously you can selectively expose the secrets like is done here, only showing it to deploy, but if the secret is exposed it is exposed from what I can tell.

webknjaz: @lasley you made a PR to the same repo, it doesn't work across forks, which is exactly what I checked 5 hours ago :)

lasley: Ohhhhh this makes sense, because the secret is encrypted against the repo. Sorry for my misunderstanding 😄

timwis: Yeah I just tried making a PR on that repo and the [build log](https://travis-ci.org/lasley/test-travis-pr-vars/builds/187544574) doesn't appear to expose them. Interesting.

_data/deployments.yml
timwis: Wow, how did I miss that? Quite sloppy of me.. thanks

timwis:
Thanks! Trailing slash fixed.

cfarm:
This looks awesome! I'm setting this up right now and glad to have found this. 

Would be useful to add description of the sample config parameters in addition to the others that are described here.

user/deployment/pages.md
```diff
@@ -0,0 +1,48 @@
...
+  on:
+    branch: master
```
cfarm: Why not add these config options and define them like the "Further Configuration" section? Make it explicit what this 'on' branch is for, plus all other options.

webknjaz: This doc describes only `pages` provider specific options, but there's a doc about the `deploy` section, which contains information about the common options.

timwis: Hey @cfarm! Funny running into you out here in the wild! Yeah, `on` is a standard property in `deploy`. I'm not sure whether the github pages provider is the best home for it, though it seems like there's not a standard template among provider documentation pages. My main hope was to get the provider at least documented (right now it's in production but no one knows about it).

user/deployment/pages.md
```diff
@@ -0,0 +1,48 @@
...
+token](https://help.github.com/articles/creating-an-access-token-for-command-line-use/)
+and set the deployment provider details in `.travis.yml`.
+
+For a minimal configuration, add the following to your `.travis.yml`:
+
+```yaml
+deploy:
+  provider: pages
+  skip_cleanup: true
+  github_token: $GITHUB_TOKEN # Set in travis-ci.org dashboard
```
plaindocs: I'd prefer that if you suggest suggesting adding a secure variable via the dashboard you don't add the entry in the yaml.

user/deployment/pages.md
```diff
@@ -0,0 +1,48 @@
...
+
+> Make sure you have `skip_cleanup` set to true, otherwise Travis CI will delete
+> all the files created during the build, which will probably delete what you are
+> trying to upload.
+
+## Setting the GitHub token
+
+You'll need to generate a [personal access
+token](https://help.github.com/articles/creating-an-access-token-for-command-line-use/)
+with the `public_repo` or `repo` scope. Since the token should be private,
```
plaindocs: might be worth a tiny explanation of the difference in those two scopes?

user/deployment/pages.md
timwis: Sorry, not sure I understand. Adding a secure variable via the dashboard just exposes an environment variable during the build, and you _use_ the environment variable like we're doing on this line, no?

user/deployment/pages.md
timwis: Sure. I think I went off of the page on [github releases](https://docs.travis-ci.com/user/deployment/releases/) for precedent, but I've just added a note to provide more context. Also, the github.com page where you actually create the token clarifies it a bit as well.

user/deployment/pages.md
plaindocs: Ah, my bad. The 'set' is ambiguous as to whether you should set it there, or it is set there.... 

How about removing the comment, and adding some info before the yaml

> For a minimal configuration, [set the github token...](link) and add the following...

Make sense?

timwis:
Just saw this was merged. Woohoo! Thx :D

PR #902, opened 2017-01-04
Fix broken link in environment variables documentation

ubruhin:
At the beginning of the [environment variables documentation](https://docs.travis-ci.com/user/environment-variables) there is a link to the element `#Defining-Variables-in-Repository-Settings`, which does not exist (link target: https://docs.travis-ci.com/user/environment-variables#Defining-Variables-in-Repository-Settings).

When JavaScript is enabled, the link works anyway, but with this change the link works also if JavaScript is disabled/blocked (just like the other links on that page).

### DIFF
--- a/user/environment-variables.md
+++ b/user/environment-variables.md
@@ -120,6 +120,8 @@ The encryption scheme is explained in more detail in [Encryption keys](/user/enc
 
 ## Defining Variables in Repository Settings
 
+{: #Defining-Variables-in-Repository-Settings}
+
 Variables defined in repository settings are the same for all builds, and when you restart an old build, it uses the latest values. These variables are not automatically available to forks.
 
 Define variables in the Repository Settings that:


plaindocs:
Hi @ubruhin 

Thanks for taking the time to check that out. I'm not really sure why that link does not work while the others do. 

As neither the search nor the collapsing menu work when JavaScript is blocked I dont really think it is worth adding the extra noise of the anchor for this situation. Please comment further if you disagree strongly.

ubruhin:
Hi @plaindocs

> I'm not really sure why that link does not work while the others do.

I added this line because other sections of the same file **do already have** [[1]](https://github.com/travis-ci/docs-travis-ci-com/blame/484e74223ff54617fd29018db7649b26ea2cbce5/user/environment-variables.md#L19) [[2]](https://github.com/travis-ci/docs-travis-ci-com/blame/484e74223ff54617fd29018db7649b26ea2cbce5/user/environment-variables.md#L83) such a reference (or whatever this line is called), so all other links on the same page are already working properly.

> As neither the search nor the collapsing menu work when JavaScript is blocked

Sure, but it's still no problem to read the docs (and if you were forwarded from a search engine, you neither need the collapsing menu nor the search), but it would be great if all links worked properly.

But of course the decision about merging this PR is up to you.

plaindocs:
That all makes sense. I'll file an issue to look into why those links don't work.

PR #927, opened 2017-01-20
Add note to use .com endpoint for Travis PRO - Triggering builds docs

iriberri:
The "Triggering builds from the API" docs might be confusing for Travis PRO users, as the sample requests are made to the `.org` endpoint. I added a note for this, similar as it is done before for the CLI commands for Travis PRO.

The change looks like this:

image

### DIFF
--- a/user/triggering-builds.md
+++ b/user/triggering-builds.md
@@ -15,7 +15,7 @@ travis login --org
 travis token --org
 ```
 
-If you are using Travis CI with a private repository use `--pro` instead of `--org`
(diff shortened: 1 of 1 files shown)


plaindocs:
Yay, more docs contributions! :heart: 

It might make sense to tag it along on this sentence:

> If you are using Travis CI with a private repository use `--pro` instead of `--org` and use `https://api.travis-ci.com` for all endpoints.

What do you think?

iriberri:
@plaindocs sounds good to me! I think it's more probable people will read it that way. I'll make the edit now! Thanks! <3

PR #929, opened 2017-01-23
remove unsupported Python versions from documentation

nschloe:
Cf. <https://github.com/travis-ci/travis-ci/issues/4990>.

### DIFF
--- a/user/languages/python.md
+++ b/user/languages/python.md
@@ -14,7 +14,7 @@ Python builds are not available on the OSX environment.
 
 ## Choosing Python versions to test against
 
-Travis CI supports Python versions 2.6, 2.7, 3.2, 3.3, 3.4, 3.5, 3.6 as well as recent development versions.
+Travis CI supports Python versions 2.6, 2.7, 3.2, 3.3, 3.4 as well as recent development versions.
 
 ```yaml
 language: python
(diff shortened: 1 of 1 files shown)


BanzaiMan:
@nschloe There might have been some misunderstanding about what you meant by "overhaul". It is wrong to say that `3.5` and others are not supported, because they are. You can use those versions if you specify them. https://github.com/travis-ci/travis-ci/issues/4990 is about those runtimes being not pre-installed.

We can continue discussing the issues you are having with the Python image, but this PR is cannot be accepted as is.

nschloe:
@BanzaiMan Thanks for the comment.

I had assumed that it's not supported due to the number of people which experience failures (see https://github.com/travis-ci/travis-ci/issues/4990) and [your comment](https://github.com/travis-ci/travis-ci/issues/4990#issuecomment-165282480)

>> There is no 3.5 with system site packages.
>> and that's a bug, right?
>
> No, it's not. We do not support that configuration.

I guess it's not clear to me what it means if something is "supported", but system-site packages aren't available.

Perhaps you could suggest a better wording?

BanzaiMan:
"This configuration" in the comment refers to "3.5 with system packages".

plaindocs:
Closing this due to https://github.com/travis-ci/travis-ci/issues/4990#issuecomment-310824690

PR #934, opened 2017-01-25
Styles for beta banner

lislis:
screen shot 2017-01-25 at 16 50 49
@plaindocs


plaindocs:
Looks awesome. It does not scream out BETA to me though, so could be assumed just to be an "important note" ?

lislis:
true. should we bring back the 'BETA'?

plaindocs:
That is my next tshirt!

But yes i think so. :-D

plaindocs:
Thanks @lislis ! :100: :tada:

PR #935, opened 2017-01-25
Specify database when activating PostGIS.

alexsegura:
I may be wrong, but if you don't specify the database to activate the extension it does not work, I had the following error:

```
function postgis_lib_version() does not exist
```


plaindocs:
Thanks for letting us know about that . Seems to work ok for me without specifying, but if you do specify it it has to exist. :-p

I'll close this for now, but if you can show me a build where it does not work as shown in the docs I'll take another look.

PR #957, opened 2017-02-07
Remotes deps also have to be called in Imports: and friends

maxheld83:
This is just an additional explanation to make sure newbies like myself don't stumble over this.
This is in relation to [this comment](https://github.com/hadley/devtools/issues/944#issuecomment-278013085) and @jimhester's response.


plaindocs:
Thanks @maxheld83 - looks good to me.

 CC @craigcitro, @hadley, and @jimhester.

PR #961, opened 2017-02-09
Remove faster builds note

cotsog:
I believe this information is now outdated as the specs of our sudo-enabled VMs are probably faster than what we had on the Legacy infrastructure.


Lyoness:
confirm!  :blue_heart:

cotsog:
Thanks @Lyoness  and @emdantrim! Merging.

Issue #940, opened 2017-01-27
Draw more attention to the node versions you can actually use


lislis:
I think over the last couple of weeks I had to tell a lot of people that they can use more versions of node than are listed in our docs. 
Most don’t read until the paragraph with nvm https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Choosing-Node-versions-to-test-against 
@plaindocs Any idea how we can make that more prominent? or not as easy to overread?

BanzaiMan:
How about dropping the list altogether? This would force the users to actually read, no?

BanzaiMan:
The following section `Available versions` is also lengthy and perhaps redundant.

plaindocs:
On more general level than just node, this is a pain point for docs + support - so definitely something we need to improve. Also related to https://github.com/travis-pro/team-sapphire/issues/503

One option would be to be more explicit, something like 

|   | Environment X | Environment Y | 
| ------------- | ------------- | ------ |
| Ipre-installed  | 1,2,4  | 3,4 5, |
| Available to download  | 3,4,5  | 1,2,3,4 | 

Another option is to not list minor versions in the docs at all, and say we support 4.latest, 5.latest, 6.latest and others, check your build environment to find out which.

I'm not very keen on 'forcing' people to read, but rather try to make the right information available to them where and when they need it (easier said than done).

@BanzaiMan I can clean up that section + list immediately if that helps in the short term.

PR #963, opened 2017-02-10
Tweak nodejs re version numbers

plaindocs:
This simplifies Nodejs versions a bit, maybe too much. Re #940 

Maybe some support people could give me some feedback? I'm happy to shift things back around if this is a bit drastic.

CC @MariadeAnton @lislis @BanzaiMan

### DIFF
--- a/user/languages/javascript-with-nodejs.md
+++ b/user/languages/javascript-with-nodejs.md
@@ -29,7 +29,7 @@ node_js:
 ```
 
 We also have many more [versions of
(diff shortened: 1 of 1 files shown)


MariadeAnton:
I like that we're relocating the long list and mentioning the estandarized way to use them on the top of the page, with a link to all those extra versions. 💯 

@lislis what do you think?

lislis:
I like it! 💯 from me as well

PR #964, opened 2017-02-10
add 2 and 3 column classes

lislis:


plaindocs:
Thanks for this also @lislis !

PR #967, opened 2017-02-15
Add notes running parallel make builds

carlad:
Change optimise to optimize while we're here to be consistent with the
rest of the file.


sitsofe:
I've added a few minor updates over in https://github.com/travis-ci/docs-travis-ci-com/pull/969 ...

PR #969, opened 2017-02-16
Update parallel make notes

sitsofe:
- Fix unlinked link(!)
- Change wording so link text looks nicer.
- Use -j rather than --jobs in the example to try and be bsdmake
  friendly (although travis appears to use GNU make by default in its
  OS and Linux environments).

### DIFF
--- a/user/speeding-up-the-build.md
+++ b/user/speeding-up-the-build.md
@@ -185,18 +185,18 @@ script:
 ### Makefile optimization
 
 If your makefile build consists of independent parts that can be safely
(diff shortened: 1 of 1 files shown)


carlad:
Hi @sitsofe. Thanks for the updates. Any chance you could update this branch with the latest commit on travis-ci gh-pages branch? We've updated HTML Proofer, so the failures in this build should now pass.

sitsofe:
@carlad Ah you were able to merge it anyway!

plaindocs:
Yeah, I just restarted the build. Thanks!

PR #974, opened 2017-02-20
Update AWS Lambda documentation

martincostello:
Update AWS Lambda documentation to be more likely to be correct for most users after I found the "minimal configuration to be insufficient when followed.

  1. The ```region``` value should be present (even though it is optional) as the majority of users worldwide probably _aren't_ using ```us-east-1```.
  1. Using a value of ```index.handler``` causes ```index.index.handler``` to be used instead, causing the lambda to fail. ```handler``` is sufficient.
  1. Add the ```runtime``` value and set to  ```nodejs4.3``` as ```nodejs``` is now deprecated and causes [deployment to fail](https://travis-ci.org/martincostello/alexa-london-travel/builds/196336961#L333).

### DIFF
--- a/user/deployment/lambda.md
+++ b/user/deployment/lambda.md
@@ -51,7 +51,10 @@ The AWS user that Travis deploys as must have the following IAM permissions in o
             "Sid": "DeployCode",
             "Effect": "Allow",
             "Action": [
-                "lambda:UploadFunction"
+                "lambda:ListFunctions",
+                "lambda:UploadFunction",
+                "lambda:UpdateFunctionCode",
+                "lambda:UpdateFunctionConfiguration"
             ],
             "Resource": [
                 "*"
--- a/user/deployment/lambda.md
+++ b/user/deployment/lambda.md
@@ -12,8 +12,10 @@ A minimal configuration is:
 deploy:
   provider: lambda
   function_name: "lambda-test"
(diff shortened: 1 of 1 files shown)


martincostello:
Also updated the IAM policy JSON to include permissions for other functions used in [the provider](https://github.com/travis-ci/dpl/blob/master/lib/dpl/provider/lambda.rb).

plaindocs:
Hi @martincostello thanks for taking a good look at this! 

Do you know if there is more generic runtime than `nodejs4.3` as otherwise we'll have to update when that changes.

Is this ` "arn:aws:lambda:<region>:<account-id>:function:<name-of-function>"` things that the user has to edit, or aws 'variables'?

martincostello:
@plaindocs Based on what I can see in the AWS Console for configuring lambda, there isn't a more generic runtime identifier:

image

I've used ```<thing>``` notation to mean a placeholder for the user's own values. Is there a different syntax you'd prefer?

martincostello:
(I was copying the format used in the example that was already there)

plaindocs:
Hi @martincostello 

Thanks for the reply. Fair point on the runtime!

I'm also aware you were juts copying existing formatting, I'll merge this, and probably tweak the wording a little to clarify later.

PR #979, opened 2017-02-21
update our legacy migration information

renee-travisci:
I found this page when helping a customer and found the section on the resources to be a bit misleading about the memory in our current container based infrastructure.


plaindocs:
Thanks renee!

PR #993, opened 2017-03-02
Add small comment about Hipchat room names being case-sensitive

iriberri:
It seems that for Hipchat's API v2, the room name *is* case-sensitive. We recently got one report about Hipchat notifications that stopped working with the v1 API, and after some investigation we think it's very likely that case sensitiveness might be being enforced in v1 too.

Just a small note in case someone experiences any problem again :)

Please, @plaindocs, could you please take a look? Thank you!

cc @MariadeAnton


iriberri:
✨

PR #996, opened 2017-03-02
Preparing for API V3 WIP

### LINKED ISSUE #981: Mention request payload returned from POST to /requests
On https://docs.travis-ci.com/user/triggering-builds ... would it be good to mention the payload returned from this endpoint? If the v3 API docs were public we could also link to the API docs.

Here's an example payload:

```
{
  "@type": "pending",
  "remaining_requests": 1,
  "repository": {
    "@type": "repository",
    "@href": "/repo/39521",
    "@representation": "minimal",
    "id": 39521,
    "name": "test-2",
    "slug": "svenfuchs/test-2"
  },
  "request": {
    "repository": {
      "id": 44258138,
      "owner_name": "svenfuchs",
      "name": "test-2"
    },
    "user": {
      "id": 3664
    },
    "id": 205729,
    "message": null,
    "branch": "master",
    "config": {
    }
  },
  "resource_type": "request"
}
```

plaindocs:
Thoughts on referring to API V2 in the menu as

```html
<li><a href="/api">API V2 (deprecated</a></li>
```

even thought it technically isn't deprecated?

CC: @MariadeAnton @renee-travisci 

96bf991 also fixes #981 although manually not with a link.


plaindocs:
@renee-travisci when we tell people to pass authentication tokens as parameters for [Webhooks](https://github.com/travis-ci/docs-travis-ci-com/pull/996/commits/a7212c6ebd3a8d4522ce118fd09ac9cc9e382653)

Is that the same API V3 token we generated above? And do you have to pass it as a parameter *AND* as part of the request? 

CC @MariadeAnton

_includes/sidebar.html
```diff
@@ -101,9 +101,10 @@
 
       <h3>Developer Program</h3>
       <ul>
-        <li><a href="/api">The API</a></li>
+        <li class="is-overview"><a href="/user/developer/">Overview</a></li>
+        <li><a href="http://developer.travis-ci.com/">API V3</a></li>
```
carlad: it's on .com and .org but maybe put .org here as we link to the explorer on .org below...

plaindocs:
Good catch, thanks @carlad . Fixed in dcf4cf3f
Are we serving the same content exactly to both? I think so right? Ie the examples on .com are using .org api?

carlad:
Am not sure how it's set up plaindocs - @joshk might know (Renee is on leave for a couple of weeks)

MariadeAnton:
I believe the API explorer in .com is using the .org examples.

renee-travisci:
@MariadeAnton is correct - I discussed w/ Josh and I was going to timebox seeing if I could make it .com/.org specific (so explorer on .com would connect to .com) to 2 hours. If I can do it quickly then great, if not I think we wanted to redirect to a single domain. I'll update here when I have an idea of how long making the explorer connected based on the domain will take.

PR #997, opened 2017-03-02
Add missing colon to yml example

jorinvo:
Just correcting a small typo.
Thank you all for running such a great service!


BanzaiMan:
A good catch! Thanks!

PR #998, opened 2017-03-02
Add canceled and errored states in webhook notifications

iriberri:
According to [Travis-Hub](https://github.com/travis-ci/travis-hub/blob/ac97c871a12330c05f64bc8a478e22c6dbf3562b/lib/travis/addons/serializer/webhook/build/message.rb#L22-L29), it seems we're missing the "Canceled" and "Errored" states in the webhook notification docs :)

The full list would be:
```ruby
pending:  'Pending',
passed:   'Passed',
failed:   'Failed',
broken:   'Broken',
fixed:    'Fixed',
failing:  'Still Failing',
errored:  'Errored',
canceled: 'Canceled'
```

### DIFF
--- a/user/notifications.md
+++ b/user/notifications.md
@@ -674,6 +674,8 @@ Additionally a message will be present in the `status_message`/`result_message` 
 - *Broken*: The build completed in failure after a previously successful build
 - *Failed*: The build is the first build for a new branch and has failed
 - *Still Failing*: The build completed in failure after a previously failed build
+- *Canceled*: The build was canceled
+- *Errored*: The build has errored
 
 The `type` field can be used to find the event type that caused this build to
 run. Its value is one of `push`, `pull_request`, `cron`, or `api`.  For pull requests,


WaleedAshraf:
LGTM 👍

WaleedAshraf:
@iriberri can you please also update webhook payload example given here: https://docs.travis-ci.com/user/notifications#Webhooks-Delivery-Format ?

Just came to know that travis also send `pull_request_title` : https://github.com/travis-ci/travis-hub/blob/ac97c871a12330c05f64bc8a478e22c6dbf3562b/lib/travis/addons/serializer/webhook/build.rb#L51

plaindocs:
@iriberri if at all possible, try to avoid using the same words in the explanation that you're defining, even though these are common Travis CI concepts. 

@WaleedAshraf When you say update the payload do you mean only adding `pull_request_title`?

WaleedAshraf:
@plaindocs I see there are a lot of missing attributes in payload example, compared to actual payload. I wanted to use `pull_request_title` , but it wasn't mentioned in payload so I just skipped. So it'd really helpful if you update the whole payload object according to latest implementation.

iriberri:
Thanks @plaindocs for the feedback and @WaleedAshraf for the explanation! I'll make some edits with both suggestions as soon as I can, thanks again!

plaindocs:
There is a related issue #981 but as the API isn't yet released...

plaindocs:
@iriberri should we just merge this in for now, and create another issue to update the payload?

iriberri:
I'm terribly sorry for the delay.

The problem I find here with defining the `errored` status specially, is that this status is set when "There was an error during the `before_install`, `install`, `before_script` or `before_deploy` phase of your build", and this seems like too many details to add to the description.

(See [Breaking the build](https://docs.travis-ci.com/user/customizing-the-build/#Breaking-the-Build))

I'm fine with merging this, at least for the moment being. Also happy to hear any suggestions.


I created a new issue to handle the payload update for the webhooks section: https://github.com/travis-ci/docs-travis-ci-com/issues/1199

WaleedAshraf:
My suggestion: @iriberri I think attaching a help link to [Breaking the build](https://docs.travis-ci.com/user/customizing-the-build/#Breaking-the-Build) right after the above details would be a lot helpful, to clear the confusion. 👍

PR #1000, opened 2017-03-02
Add note on caching directories and permissions

BanzaiMan:


plaindocs:
How do you feel about rewording to something like

> Caches and directory permissions
>
> When caching [custom files and directories](/user/caching/#Arbitrary-directories)
> make sure that the location you specify is readable and writable by the user. 
> If it isn't, the caching utility reports errors when it
> invokes `tar` to create the cache archive.
>
> Example error:

BanzaiMan:
@plaindocs Can we merge this?

PR #1001, opened 2017-03-03
feat(doc): Add note about safelisting branches and tags

Kiougar:
Add a note in the `Safelisting or blocklisting branches` section that notifies the user that safelisting branches causes tagged commits to be skipped. Also, propose an example regex that allows specific tags to be safelisted.

### DIFF
--- a/user/customizing-the-build.md
+++ b/user/customizing-the-build.md
@@ -257,6 +257,8 @@ branches:
   - stable
 ```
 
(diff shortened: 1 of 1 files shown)


plaindocs:
Thanks for catching that @Kiougar ! I reworded both your note and some of the surrounding text. Please comment here if you have any thoughts on any of that.

Kiougar:
Hey @plaindocs, thanks for the rewording. 👍 

The only thing I found was a typo:

> `the takes safelist takes precedence`

should be

> `the safelist takes precedence`

That is, the first instance of `takes` should be removed.

plaindocs:
Second pair of eyes always useful! Thanks. 866c7e9e

PR #1002, opened 2017-03-03
Update javascript-with-nodejs.md to include ver 7

winksaville:
Include information about newer versions, there were many references to 0.12 and no references to version 7


plaindocs:
Thanks!

We're working on a better way for these.

winksaville:
You're welcome,

Issue #982, opened 2017-02-23
Old Python versions and pytest vs py.test


iriberri:
## Background

For some old versions of pytest, which is installed by default in our Python images, the executable is named "py.test" instead of "pytest".

This affects the builds running on Python 3.4 and below, which come with pytest 2.6.4 preinstalled, whereas the images for Python 3.5 and above include pytest 3.0.5. (For the Precise environment)

For the old versions, running pytest with the `py.test` command instead should work. 

## Question / issue description

In this section of the Python environment docs: https://docs.travis-ci.com/user/languages/python/#Choosing-Python-versions-to-test-against, we show an example with several Python versions and `pytest` as the command to be run:

```yml
language: python
python:
  - "2.6"
  - "2.7"
  - "3.2"
  - "3.3"
  - "3.4"
  - "3.5"
  - "3.5-dev" # 3.5 development branch
  - "3.6"
  - "3.6-dev" # 3.6 development branch
  - "3.7-dev" # 3.7 development branch
  - "nightly" # currently points to 3.7-dev
# command to install dependencies
install: "pip install -r requirements.txt"
# command to run tests
script: pytest
```

However, in a section below: https://docs.travis-ci.com/user/languages/python/#Specifying-Test-Script, we say to use `py.test` instead of `pytest` for Python versions 3.5 and below:

```yml
script: pytest  # or py.test for Python versions 3.5 and below
```

Is there a typo in the first section that suggests using `pytest` instead of `py.test` with all Python versions? At the moment, running a build in Precise with those options in the .travis.yml file is failing for all the versions listed there with the error:

```
$ pytest
/home/travis/build.sh: line 62: pytest: command not found
The command "pytest" failed and exited with 127 during .
```

whereas using `py.test` works in those environments.

However, using Trusty, `pytest` is working for all Python versions except 3.3 and 3.2.

It seems that older versions of this sample configuration were using `nosetests` instead, so I think this was a recent addition. :)

Related: https://github.com/travis-ci/docs-travis-ci-com/pull/827

plaindocs:
Great catch. And lovely detailed issue! Seriously, can we point people to this in the 'how to report bugs'?

plaindocs:
Easiest fix would just be to symlink both in the environments of course :-p .

plaindocs:
So, other than the obvious (which is not telling people stuff that doesn't work) what is the best fix here?

[...]

PR #1020, opened 2017-03-14
py.test is not needed on old plattforms

guettli:
It is available there, too.

### DIFF
--- a/user/languages/python.md
+++ b/user/languages/python.md
@@ -85,7 +85,7 @@ For example, if your project uses pytest:
 
 ```yaml
 # command to run tests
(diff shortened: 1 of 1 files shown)


plaindocs:
Thanks for reporting.

We're taking a more detailed look at this in #982 and am not quite sure what the ideal fix is yet. Any comments welcome on that issue.

guettli:
Yes, I was wrong. The `pytest` is available in my virtualenv. But it is not available in the travis environment.

Please add this to the travis environment. This would make the integration of travis a bit more straight forward.

Here is my test which fails on travis because pytest is missing: https://travis-ci.org/guettli/compare-with-remote/jobs/210979841

PR #1024, opened 2017-03-16
Remove documentation about Exoscale

vincentbernat:
The PaaS product from Exoscale is not available anymore. Remove the
associated documentation.


plaindocs:
Hi @vincentbernat thanks for that! 

Do you have a link for the announcement?

vincentbernat:
No, the communication was only done to the existing users. If you are worried about the authenticity, I am part of the @exoscale GitHub organization and the pull request is from the organization as well.

plaindocs:
Thanks. I was more concerned about leaving the reason + dates in the commit. But all good.

Issue #1017, opened 2017-03-13
Builds with services: docker run on Trusty even if other dist is configured


iriberri:
Travis CI does internally route builds with `services: docker` configured to the `sudo: required` Trusty infrastructure.

It'd be interesting to add a note about this in the Docker section (https://docs.travis-ci.com/user/docker/), where we already specify that `sudo: required` is needed.

It seems that this is causing some confusions lately, as the distribution can be set to Precise in the .travis.yml file but activating the Docker service will make Travis CI ignore that detail.

Issue #953, opened 2017-02-03
Docker: document steps to update to last version


iriberri:
Docker 1.13 was released a couple of weeks ago and there are people trying to use its new features in Travis CI. It could be useful to document the steps required to update to the latest version (1.13 now, since we have 1.12).

@cotsog suggested adding these lines to the .travis.yml file:

```
before_install:
  - sudo apt-get update
  - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-engine
  - docker --version # for verification
```

plaindocs:
Sure - do you want to make a PR or shall I do it?

iriberri:
@plaindocs I'm happy to add the changes myself! I'll try to send you a PR later today. :)

PR #1029, opened 2017-03-17
Docker: document steps to update to last version and route to Trusty

iriberri:
Closes https://github.com/travis-ci/docs-travis-ci-com/issues/953
Closes https://github.com/travis-ci/docs-travis-ci-com/issues/1017

This PR adds a section with the required steps to update to the last Docker version. The Docker version preinstalled in Trusty sudo-required is:

```
$ docker --version
Docker version 1.12.3, build 6b644ec
```

Also, I'm including a note about builds with `services: docker` configured being routed by Trusty. I've seen a couple questions about this lately, as users seem confused because they're setting `dist: precise` but this is ignored. For the record, this is done here: https://github.com/travis-ci/travis-api/blob/d9b80d82f1b88d3b8fd1f05da4b5e74bfda68b9e/lib/travis/model/build/config/dist.rb#L10-L11

This looks like:

image

and

image

Any feedback would be appreciated, @plaindocs! 💛

### DIFF
--- a/user/docker.md
+++ b/user/docker.md
@@ -187,7 +187,6 @@ updating `docker-engine` in the `before_install` step of your `.travis.yml`:
 before_install:
   - sudo apt-get update
   - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-engine
-  - docker --version # for verification
 ```
 
 > Check what version of Docker you're running with `docker --version`
--- a/user/docker.md
+++ b/user/docker.md
@@ -180,7 +180,6 @@ before_install:
 
 ### Installing a newer Docker version
 
-The pre-installed Docker version is possibly not the latest.
 You can upgrade to the latest version and use any new Docker features by manually
 updating `docker-engine` in the `before_install` step of your `.travis.yml`:
 
@@ -190,6 +189,8 @@ before_install:
   - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-engine
   - docker --version # for verification
 ```
+
+> Check what version of Docker you're running with `docker --version`
 
 #### Examples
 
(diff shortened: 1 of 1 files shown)


user/docker.md
```diff
@@ -175,6 +178,19 @@ before_install:
   - sudo mv docker-compose /usr/local/bin
 ```
 
+### Installing a newer Docker version
+
+The Trusty sudo-required environment comes with Docker 1.12 pre-installed.
```
cotsog: I think I wouldn't include a version number as it will probably change in the next environment update. Maybe something along the line of the following would be better?

> The pre-installed Docker version is possibly not the latest. You can upgrade to ...

user/docker.md
```diff
@@ -175,6 +178,19 @@ before_install:
   - sudo mv docker-compose /usr/local/bin
 ```
 
+### Installing a newer Docker version
+
+The pre-installed Docker version is possibly not the latest.
```
plaindocs: This line doesn't add much.

user/docker.md
```diff
@@ -175,6 +178,19 @@ before_install:
...
+You can upgrade to the latest version and use any new Docker features by manually
+updating `docker-engine` in the `before_install` step of your `.travis.yml`:
+
+```yaml
+before_install:
+  - sudo apt-get update
+  - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-engine
+  - docker --version # for verification
```
plaindocs: Might be worth taking that out of the yaml, and just saing something like

> Check what version of docker you're running with `docker --version`

plaindocs:
Otherwise looks good.

Awesome idea on the note! Do we have a similar warning in the build log or linter? Ie if you set one OS but overrride with the docker service.

iriberri:
I just applied the feedback, @plaindocs! Thanks for the review :)

PR #1033, opened 2017-03-20
Corrected sentence framing

prajapati-parth:


plaindocs:
Thanks! I really should add a check for duplicate words!

Issue #607, opened 2016-06-10
Double check bash / sh / console code samples


plaindocs:
Highlighter uses `bash` not `sh` or `console`.

PR #1034, opened 2017-03-20
Prism JS uses bash not sh

plaindocs:
Fix for #607


MariadeAnton:
Yay for consistency ✨  Thank you so much, @plaindocs!

PR #1036, opened 2017-03-20
Tweak npm

### LINKED ISSUE #246: Clarification for the NPM instructions:
https://github.com/travis-ci/docs-travis-ci-com/blob/gh-pages/user/deployment/npm.md

says:

> For a minimal configuration, all you need to do is add the following to your `.travis.yml`:
> 
> ```
> deploy:
>   provider: npm
>   email: "YOUR EMAIL ADDRESS"
>   api_key: "YOUR API KEY"
> ```

Where "YOUR API KEY"  is the result of `echo -n "username:password" | base64` optionally encrypted.

I suppose that the username/password pair is the npm one, but what about the email address?

---

As a side note,  accepting non-encrypted credentials is IMO doing a disservice to your users.

plaindocs:
I reworded this slightly. 

We can probably delete the npm 1 stuff, as it is now on 4.4!

Fixes #246


MariadeAnton:
This looks great, thanks for the revamp of the npm deployment page! 

To confirm about the npm 1 stuff, you mean when we mention that the npm is ~1 or older? Do those instructions apply to all versions after ~1 ? 

Do we know when the jump happens? Checking npm releases at https://github.com/npm/npm/releases?after=v1.0.0rc9

Seems it jumps from `v0.3.17` to `v1.0.0-1-rc`

Also, what happens if the version does not fit that criteria? Needs to be updated?

plaindocs:
Right, good questions. 

As I see it we've got two formats of `~/.npmrc`

* ~1 (I'm assuming but have not checked that covers everything pre 2 - and should probably clarify)
  
   The 1 RC you link above is from 2011! 
  
* 2 and newer (also an assumption - but makes sense in that an api key is a better system than user/pass and probably has not needed to be changed)

   Dating approx September 2014

Thoughts?

plaindocs:
This never got merged in, so I'm doing that now.  We can fix if anything comes up.

PR #1042, opened 2017-03-21
Add mariadb and rethinkdb + postgres link has changed

grooverdan:


plaindocs:
Thanks, and good catch! 

I don't know why HTMLProofer thinks that link does not exist. Seems to be stripping the root slash?

grooverdan:
No - i missed one on rethinkdb

grooverdan:
thanks.

Issue #984, opened 2017-02-24
In cache section: ccache does not come preinstalled in OS X images


iriberri:
https://docs.travis-ci.com/user/caching#ccache-cache

I thought it could be interesting to add a mention here, as ccache is not available by default in the OS X infrastructure.

iriberri:
We got a suggestion about mentioning to add `/usr/local/opt/ccache/libexec` to the path too when installing `ccache` via `brew`. Reopening to include this too! :)

plaindocs:
@iriberri do you mean to `$PATH` or to the cache key in `.travis.yaml` ?

I'm guessing the first option. But that sounds kinda wrong according to https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s07.html

iriberri:
@plaindocs thanks for your response!

Yes, I mean prepending this to the `$PATH`.

I don't think the link you've posted is related. I'm talking about the `libexec` folder that ships with the `ccache` install. The folder may not have the best name, but the docs state it's meant to be placed in the PATH so it can override default compiler calls (e.g. `gcc`, `g++`) and invoke them using `ccache` as a wrapper.

By having these wrappers on the `$PATH`, all builds in the system will default to using `ccache` without the user needing to modify their Makefile. Otherwise a user would need to edit the Makefile and replace `gcc` calls with `ccache gcc`. Does this make sense? I'm sorry I didn't include this info in my previous comment!

HalfWall:
who are you

HalfWall:
who are you

peternewman:
@plaindocs your example actually appends, rather than prepends the ccache folder to path, so it will find the normal binaries in /usr/bin or wherever first and hence not use ccache.

I've fixed this here:
https://github.com/travis-ci/docs-travis-ci-com/pull/1270

I think when this is in, this issue can be closed.

iriberri:
Thanks for pointing this out, @peternewman! That was a good catch :)

iriberri:
Closing (via https://github.com/travis-ci/docs-travis-ci-com/pull/1270)

peternewman:
Thanks for merging @iriberri .

PR #1043, opened 2017-03-22
Clarify path for ccache

plaindocs:
Fix for #984 

@iriberri

### DIFF
--- a/user/caching.md
+++ b/user/caching.md
@@ -140,8 +140,8 @@ If you are using `ccache`, use:
 language: c # or other C/C++ variants
 
 cache: ccache
-``
- automatically
+```
+
 to cache `$HOME/.ccache` and automatically add `/usr/lib/ccache` to your `$PATH`.
 
 #### ccache on OSX
--- a/user/caching.md
+++ b/user/caching.md
@@ -140,9 +140,9 @@ If you are using `ccache`, use:
 language: c # or other C/C++ variants
 
 cache: ccache
-```
-
-to cache `$HOME/.ccache`, and add `/usr/lib/ccache` to the front of `$PATH`.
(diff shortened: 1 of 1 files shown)


user/caching.md
```diff
@@ -134,17 +134,27 @@ caches `$HOME/.cache/pip`.
...
+If you are using `ccache`, use:
 
 ```yaml
 language: c # or other C/C++ variants
 
 cache: ccache
 ```
 
-caches `$HOME/.ccache`, and adds `/usr/lib/ccache` to the front of `$PATH`.
+to cache `$HOME/.ccache`, and add `/usr/lib/ccache` to the front of `$PATH`.
```
iriberri: In the case of Linux, Travis automatically edits the PATH when it detects that ccache is being used: https://github.com/travis-ci/travis-build/blob/35b5046996ef1f50d9be2206e921666e0937b8d5/lib/travis/build/script/c.rb#L12-L13

plaindocs: Yeah, but don't we already mention that? 

Ah, I see what you mean. Interesting take. I'll reword.

user/caching.md
```diff
@@ -134,17 +134,27 @@ caches `$HOME/.cache/pip`.
...
 
-For caching `ccache` files, use:
+If you are using `ccache`, use:
 
 ```yaml
 language: c # or other C/C++ variants
 
 cache: ccache
-```
+``
```
iriberri: Oops! Missing quote here! :) and I think that the `automatically` of the next line is also duplicated, right?

plaindocs: :-( some keyboard gibberish. Fixed. Thanks for checking!

PR #1035, opened 2017-03-20
Fix up the docker instructions

plaindocs:
Fix for #900

* [ ] what is the difference between garnet and amethyst 
* [ ] Is the PID instruction valid for both 
* [ ] the docker command still references quay, so we need one for trusty also

@MariadeAnton @iriberri any more thoughts and suggestions?

PR #1045, opened 2017-03-22
display direct `p` children of `li` as inline, fix line break in lists

lislis:
turns
screen shot 2017-03-22 at 14 28 21
into
screen shot 2017-03-22 at 14 28 28

### DIFF
--- a/assets/stylesheets/_base.sass
+++ b/assets/stylesheets/_base.sass
@@ -76,6 +76,9 @@ li ul
   padding-left: 2.5em
   list-style-position: outside
 
+li > p
+  display: inline
+
 table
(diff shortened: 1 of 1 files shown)


MariadeAnton:
Nice, thank you 💪

plaindocs:
Funny, I couldn't place this - and wasn't sure why - but of course it is because I have an outstanding PR for that page: https://github.com/travis-ci/docs-travis-ci-com/pull/1035

@lislis thanks muchly for the fixup at short notice.

Markdown lists get really fiddly when you have multiple levels, and code blocks inside them. So some end up with `<p>` and some without.

PR #1046, opened 2017-03-22
Add Note styles to protocol details on private dependencies

iriberri:
I've seen a couple of reports today in which there was some confusion setting SSH keys to access private dependencies but the dependency URL wasn't updated appropriately to have the format to clone the submodules via SSH.

I thought that converting this paragraph to a note would highlight it. What do you think?

This change looks like:

image


plaindocs:
2 char pr for great effect! I like it!

If we *still* get queries about this we could move it from table format to 4 subsections, with an example in each.

PR #1055, opened 2017-03-23
CCMenu: Add explicit project URL for .com

iriberri:
We got a report that this endpoint was missing, so this PR adds it.

Perhaps it could be interesting to add a new section in which all the different endpoints are listed (specific repo & owner, .org & .com) to make these more accessible, as currently these are distributed through the text.


iriberri:
Also: we should specify that it's possible to use the `branch` parameter for the endpoints:

```
https://api.travis-ci.com/repos/<owner>/<repo>/cc.xml?token=<token>&branch=<branch>

https://api.travis-ci.com/repos/<owner>.xml?token=<token>&branch=<branch>​
​
​
https://api.travis-ci.org/repos/<owner>/<repo>/cc.xml?branch=<branch>

https://api.travis-ci.org/repos/<owner>.xml?branch=<branch
```

iriberri:
I just had some time to get back to this.

I've updated the image of the token used in this section. I uploaded a new image in this case to not alter (yet!) other sections that are using the same one.

Also, I included a new paragraph to explain that it's possible to fetch the feed by specifying the branch of your project that you want to track, which was pending since my previous comment.

@plaindocs does this look good to go? :)

PR #1062, opened 2017-03-28
Add xcode8.3 image

emdantrim:


plaindocs:
@emdantrim from the attached blue PR issue looks like this is good to merge?

PR #1065, opened 2017-03-28
GitHub releases deployment: limit api_key scope to subset of repositories

dhimmel:
It's not ideal that `travis setup releases` generates an `api_key` that has write access to all of a users (public) repositories. This means that users with security critical repositories cannot use `releases` deployment on non-security critical repos.

I'm assuming GitHub's SSH deploy keys don't allow creating releases. If that's not the case, then I think `travis setup releases` should use SSH deploy keys instead of OAuth tokens.

One [workaround](http://stackoverflow.com/a/29450135/4651668) is to create a GitHub machine user account that has it's scope limited to one or more repos. I recently [switched](https://github.com/dhimmel/hetio/commit/558e3791654b4e97de688a7c3d5a53bf615fbc73) my repository to use this method. You can see that [as of v0.2.5](https://github.com/dhimmel/hetio/releases), @hetio-release-bot is used to deploy rather than @dhimmel.

This PR warns users that they are granting write access to all of their repos by running `travis setup releases` and suggests the machine user workaround.


plaindocs:
Thanks.

PR #1067, opened 2017-03-29
.NET / C# Add additional required config to sample

masaeedu:
The sample given in the docs for .NET Core does not work out of the box. A supported distro and startup scripts must be configured. See travis-ci/travis-ci#6640 and travis-ci/travis-ci#6805.

### DIFF
--- a/user/languages/csharp.md
+++ b/user/languages/csharp.md
@@ -96,7 +96,6 @@ language: csharp
 mono: none
 dotnet: 1.0.1
 dist: trusty
-sudo: required
 script:
  - dotnet restore
  - ...
--- a/user/languages/csharp.md
+++ b/user/languages/csharp.md
@@ -95,12 +95,19 @@ By default, Travis CI does not test against .NET Core. To test against .NET Core
 language: csharp
 mono: none
 dotnet: 1.0.1
+dist: trusty
+sudo: required
+script:
+ - dotnet restore
+ - ...
 ...
 ```
 
 > *Note*: you need to specify the version number of the .NET Core SDK (_not_ the .NET Core Runtime). For example, the .NET Core SDK 1.0.1 contains both the .NET Core Runtime 1.0.4 and 1.1.1.
 
-The version numbers of the SDK can be found on the [.NET Core website](https://dot.net/core).
+The version numbers of the SDK can be found on the [.NET Core website](https://dot.net/core). 
+
+Since .NET Core is not supported on all Linux distros, you should specify a supported Linux distro in your Travis CI configuration. Additionally, since .NET Core does not define default scripts, you should provide a script that invokes `dotnet` to build, test and publish your solution.
 
 ### Testing Against Mono and .NET Core
 
--- a/user/languages/csharp.md
+++ b/user/languages/csharp.md
@@ -106,7 +106,7 @@ script:
 
 The version numbers of the SDK can be found on the [.NET Core website](https://dot.net/core). 
 
-Since .NET Core is not supported on all Linux distros, you should specify a supported Linux distro in your Travis CI configuration. Additionally, since .NET Core does not define default scripts, you should provide a script that invokes `dotnet` to build, test and publish your solution.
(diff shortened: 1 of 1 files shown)


user/languages/csharp.md
```diff
@@ -95,12 +95,19 @@ By default, Travis CI does not test against .NET Core. To test against .NET Core
 language: csharp
 mono: none
 dotnet: 1.0.1
+dist: trusty
+sudo: required
```
JSkimming: I don't believe `sudo` is required. I've just set-up a .NET Core build, and it builds fine configured with just the trusty distro. See below:

https://travis-ci.org/JSkimming/Castle.Core.AsyncInterceptor/jobs/221394769/config

This may be working because it's running on the [Container-based Ubuntu Trusty in Public Beta](https://blog.travis-ci.com/2016-11-08-trusty-container-public-beta "Travis CI Blog").

masaeedu: 👍  Done.

user/languages/csharp.md
```diff
@@ -95,12 +95,18 @@ By default, Travis CI does not test against .NET Core. To test against .NET Core
...
+ - ...
 ...
 ```
 
 > *Note*: you need to specify the version number of the .NET Core SDK (_not_ the .NET Core Runtime). For example, the .NET Core SDK 1.0.1 contains both the .NET Core Runtime 1.0.4 and 1.1.1.
 
-The version numbers of the SDK can be found on the [.NET Core website](https://dot.net/core).
+The version numbers of the SDK can be found on the [.NET Core website](https://dot.net/core). 
+
+Since .NET Core is not supported on all Linux distros, you should specify a supported Linux distro in your Travis CI configuration. Additionally, since .NET Core does not define default scripts, you should provide a script that invokes `dotnet` to build, test and publish your solution.
```
plaindocs: Is this about Trusty / Precise? Or about distros generally?

masaeedu: @plaindocs .NET Core doesn't support `precise`, which is the default rootfs for whatever container environment Travis is using. In the example I simply changed it to the supported Trusty release so that copy-pasting from the docs works, however in general there may be other distros that .NET Core supports. The error message you get from Travis because of the unsupported distro can be seen [here](https://travis-ci.org/vicmosin/badaboom/builds/185688296), and isn't particularly helpful.

BanzaiMan:
@Joshua-Anderson @akoeplinger @nterry Opinions, please.

user/languages/csharp.md
Joshua-Anderson: I'd possibly rephrase it to say that .NET core is  only  supported on the trusty travis CI image so it's a little less vague. Otherwise :+1: Thanks!

masaeedu: @Joshua-Anderson But it's not *only* supported on the Trusty image. I'm pretty sure it's also supported on CentOS and a dozen others. The problem this is solving is that it's *not* supported on the default distro, i.e. Precise. If you want to limit Travis CI support for .NET to the Trusty image, sure, we can put that in, but that's an artificial limitation.

Joshua-Anderson: Yep! The distinction I was intending was to specify what Travis CI images are compatible with .NET core so that users knew what option they should pick.

masaeedu: @Joshua-Anderson Ok, I've changed it to specifically refer to Ubuntu Trusty.

plaindocs:
New change looks good to me. If you could fix the merge conflict I'll merge this in.

masaeedu:
@plaindocs It looks like the conflict is because someone else already merged in the following line:

> You need to specify `dist: trusty` so your build will run on Ubuntu 14.04 because .NET Core doesn't work on Ubuntu 12.04 (which is the default Linux build environment on Travis right now).

I have to get back to some other stuff, so please feel free to merge the section concerning default scripts in with this in whatever way you deem suitable. Thanks.

PR #1068, opened 2017-03-30
Warn about using netrc methods for git submodule

joepvd:


user/private-dependencies.md
```diff
@@ -199,6 +200,22 @@ gem 'lib1', github: "myorg/lib1"
 gem 'lib2', github: "myorg/lib2"
 ```
 
+
+**Note**: In case of private git submodules, be aware that the `git submodule
```
plaindocs: Can you use the quote symbol `>` for notes (without the Note text)

plaindocs: Maybe reword to something like

> be aware that the default `git submodule update --init recursive` command runs before the netrc credentials are updated. If you are writing credentials to netrc, disable the loading of submodules first, and do this as a separate explicit build step after updating the credentials

plaindocs: And then see if you can make the code block a child of the note? 8 spaces might do the trick :-p

user/private-dependencies.md
```diff
@@ -250,6 +267,21 @@ gem 'lib1', github: "myorg/lib1"
 gem 'lib2', github: "myorg/lib2"
 ```
 
+**Note**: In case of private git submodules, be aware that the `git submodule
+update --init recursive` command will by default run before you will have the chance
+of storing the credentials. If you are going this route, disable the loading of
+submodules first, and do this as a separate step after the credentials have been
+added:
```
plaindocs: And most of the same as the comments on the previous note

plaindocs:
Sorry - more nitpicks:

* yaml not yml on the code blocks
* it is better to avoid all of the "will x", so

>In case of private git submodules, be aware that the `git submodule
> update --init recursive` command runs before the `~/.netrc` credentials
> are updated. If you are writing credentials to `~/.netrc`, disable the automatic loading of
> submodules, update the credentials and add an explicit step to update the submodules:

* and also the "this as a separate explicit step" can be confusing on first read. I missed this first time round. Suggestion above.

Sound ok?

plaindocs:
Ping. :-)

PR #1071, opened 2017-03-30
Add Troubleshooting section with solutions to common errors

wmhilton:
These are the 3 most common errors I encounter when using Travis with npm, which generally take me a half hour to solve each time, because I forget the solutions. I figure the world would benefit from having those solutions documented here in the official Travis docs.

(previously discussed in https://github.com/travis-ci/travis-ci/issues/7005)

### DIFF
--- a/user/deployment/npm.md
+++ b/user/deployment/npm.md
@@ -133,7 +133,9 @@ after_deploy:
   - ./after_deploy_2.sh
 ```
 
-### Troubleshooting common problems
+### Troubleshooting
+
+Did you expect a job to publish a package but it did not? Here are some common reasons that can happen and how to resolve them.
 
 #### Skipping a deployment with the npm provider because this is not a tagged commit
 
@@ -141,24 +143,24 @@ The most likely cause is that `git push` does not push tags. Be sure to run `git
 
 #### Skipping a deployment with the npm provider because this is not on the required runtime
 
-This is because the node version being used for the test does not match the version specified in the conditonal release section:
+This happens when the node version being used for a job does not match the version specified in the conditonal release section. This can result from copying and pasting from another Travis file or if you originally tested on multiple node versions and switch to just testing on one.
 
 ```
 language: node_js
-node_js: 6  <--- version used for test
+node_js: 6  # version used for job
 deploy:
   ...
   on:
-    node: node  <--- version condition in deployment
+    node: node  # version condition for deployment
     tags: true
     ...
 ```
 
-If you are not testing against multiple versions of node, you can remove the `node: ` condition entirely. If you are testing your project against multiple versions of node then you probably added the version condition to the deployment to prevent Travis from attempting to publish the module multiple times in one build. In that case just make sure that the version in `deploy.on.node` is one of the versions listed in `node_js`.
+This may be the intended behavior! If you are testing your project against multiple versions of node, then you probably should add the version condition to prevent concurrent Travis jobs from attempting to publish the same package version during a build. In this case make sure that the version in `deploy.on.node` is one of the versions listed in `node_js`, otherwise *none* of the jobs will do the deploy. If you are *not* testing against multiple versions of node, you can remove the `node: ` condition entirely.
 
 #### npm ERR! You need a paid account to perform this action. For more info, visit: https://www.npmjs.com/private-modules
 
-By default [scoped packages](https://docs.npmjs.com/misc/scope) are not public. If you want to publish your package publicly, you can do so manually by tacking on ` --access public` to the `npm publish` command. However, the best way is to set `access` in the `package.json` file like so:
+npm assumes that [scoped packages](https://docs.npmjs.com/misc/scope) are private by default. You can explicitly tell npm your package is a public package and avoid this error by adding the following to your `package.json` file:
 
 ```
   "publishConfig": {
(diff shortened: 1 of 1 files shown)


user/deployment/npm.md
```diff
@@ -132,3 +132,36 @@ after_deploy:
...
+
+The most likely cause is that `git push` does not push tags. Be sure to run `git push --tags`.
+
+#### Skipping a deployment with the npm provider because this is not on the required runtime
+
+This is because the node version being used for the test does not match the version specified in the conditonal release section:
+
+```
+language: node_js
+node_js: 6  <--- version used for test
```
BanzaiMan: This is valid YAML, but does not define the data I suspect you want. YAML inline commends are introduced by `#`, so:

```yaml
language: node_js
node_js: 6 # version used for test
```

user/deployment/npm.md
```diff
@@ -132,3 +132,36 @@ after_deploy:
   - ./after_deploy_1.sh
   - ./after_deploy_2.sh
 ```
+
+### Troubleshooting common problems
+
+#### Skipping a deployment with the npm provider because this is not a tagged commit
```
BanzaiMan: I am not sure if you want to characterize these things as "problems". There are very good reasons to specifically limit deployment to a single job, and calling these are "problems" is misleading.

Conditional deployment is already covered in https://docs.travis-ci.com/user/deployment/#Conditional-Releases-with-on%3A.

user/deployment/npm.md
```diff
@@ -132,3 +132,36 @@ after_deploy:
...
+
+The most likely cause is that `git push` does not push tags. Be sure to run `git push --tags`.
+
+#### Skipping a deployment with the npm provider because this is not on the required runtime
+
+This is because the node version being used for the test does not match the version specified in the conditonal release section:
```
BanzaiMan: "This is because …" is bad. Use "The reason is that …".

user/deployment/npm.md
```diff
@@ -132,3 +132,36 @@ after_deploy:
...
+
+```
+language: node_js
+node_js: 6  <--- version used for test
+deploy:
+  ...
+  on:
+    node: node  <--- version condition in deployment
```
BanzaiMan: Same.

user/deployment/npm.md
```diff
@@ -132,3 +132,36 @@ after_deploy:
...
+    tags: true
+    ...
+```
+
+If you are not testing against multiple versions of node, you can remove the `node: ` condition entirely. If you are testing your project against multiple versions of node then you probably added the version condition to the deployment to prevent Travis from attempting to publish the module multiple times in one build. In that case just make sure that the version in `deploy.on.node` is one of the versions listed in `node_js`.
+
+#### npm ERR! You need a paid account to perform this action. For more info, visit: https://www.npmjs.com/private-modules
+
+By default [scoped packages](https://docs.npmjs.com/misc/scope) are not public. If you want to publish your package publicly, you can do so manually by tacking on ` --access public` to the `npm publish` command. However, the best way is to set `access` in the `package.json` file like so:
```
BanzaiMan: Deployment with `npm` does not provide the means of "tacking on `--access public`" (right now, anyway), so mentioning that seems unnecessary.

user/deployment/npm.md
wmhilton: What if we just called the section `### Troubleshooting` then? And yes, they are not always problems - they are really just "informational messages" rather than error messages. But if you're trying to get a package published and it says "Skipping blaw blaw" then that's the phrase you end up Googling. That's why I labeled the headers that way, so they would stand out and so Google would find an exact match.

Yes, I know conditional deployment is already covered. But since this is a troubleshooting section, I thought why not be helpful and save the weary Googler from reading the entire manual?

user/deployment/npm.md
wmhilton: Hmm, you're right... how about "This happens when..."? I like the sound of that.

user/deployment/npm.md
wmhilton: Aww... you got me. Yes, it's probably wiser to use actual grammar constructs than draw free-form ASCII arrows. I can change that.

user/deployment/npm.md
wmhilton: By "manually" I meant you can just run "npm publish --access public" on your local workstation to publish the initial version, and then after the first publication I believe the "--access public" is not required for subsequent versions. I read that somewhere but haven't tested it.

I thought it worth mentioning in case changing the package.json file wasn't an available option or something... they could do a manual publish of a pre-release and then re-run the Travis build to publish the correct version. But I can't think of a situation where that would be necessary, so I may cut that part and just reword it to say "One way... " rather than "the best way" because who am I to judge what's best?

user/deployment/npm.md
```diff
@@ -132,3 +132,38 @@ after_deploy:
   - ./after_deploy_1.sh
   - ./after_deploy_2.sh
 ```
+
+### Troubleshooting
+
+Did you expect a job to publish a package but it did not? Here are some common reasons that can happen and how to resolve them.
```
BanzaiMan: As I indicated earlier, the first two items are covered elsewhere. Duplicating information is a maintenance burden which should be avoided unless the value of doing so exceeds it. I don't think it does here.

wmhilton: I disagree, as I have spent >1hr of my time on the first two issues, and far less than that documenting them. My intention is purely noble - spare others the grief of hunting down the cause and resolution of their misfortune. But I intentionally configured this PR with "Allow edits from maintainers" so you can do whatever you want with it.

PR #1072, opened 2017-03-30
Change instructions about whitelisting packages

pnorman:
The whitelist is not being extended, so suggest the recommended workaround of `sudo: required`.

### DIFF
--- a/user/migrating-from-legacy.md
+++ b/user/migrating-from-legacy.md
@@ -74,7 +74,7 @@ addons:
 ```
 
 The aliases for the allowed sources (such as `deadsnakes` above) are managed in a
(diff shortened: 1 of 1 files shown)


user/migrating-from-legacy.md
```diff
@@ -88,7 +88,7 @@ addons:
     - time
 ```
 
-The allowed packages are managed in a [whitelist](https://github.com/travis-ci/apt-package-whitelist), and any attempts to install disallowed packages will result in a log message detailing the package approval process.
+The allowed packages are managed in a [whitelist](https://github.com/travis-ci/apt-package-whitelist). If you need additional packages you must use `sudo: required`.
```
BanzaiMan: Packaging whitelisting requests are still processed.

pnorman: Done.

plaindocs:
@BanzaiMan I can't find your review to tell if it is fixed, could you merge if it is? Thanks

PR #1077, opened 2017-04-01
Replace occurrences of `whitelist` with `safelist`

li141319:


BanzaiMan:
Sorry, but changing these repositories' names takes a little more planning.

spl:
It seems like this has been done and the docs now need to be updated. See https://github.com/travis-ci/apt-package-safelist/commit/d278e527f4668a26940e20cb9e8c5a0bbdac7edd

Issue #416, opened 2015-11-03
Note on using cache on Pull Request builds


MariadeAnton:
Out of customer support.

We could highlight the line saying: **Currently Pull Requests will use the cache of the branch they are supposed to be merged into.**

at http://docs.travis-ci.com/user/caching/#Fetching-and-storing-caches

And also specify that once a PR has its own cache, in order to start with a fresh/new one, besides from clearing the PR's cache, users also need to clear out the branch the PR is going to be merged into.

from: https://secure.helpscout.net/conversation/134497874/21700/

damianavila:
:+1:

MariadeAnton:
Also relevant here from: https://github.com/travis-ci/docs-travis-ci-com/issues/433

> With a development workflow where "push builds" are disabled and caching is used, the cache for the upstream branch (where all PRs are merged) won't be populated hence the builds on that branch won't have the speedup caching is supposed to lead to.

This has been fairly explained at https://docs.travis-ci.com/user/caching#Fetching-and-storing-caches

but could use a note explaining that when working only with PRs as cache is populated from the target branch or the default branch, it will not be populated

@plaindocs does this clarify things a bit? let me know!

MariadeAnton:
This was addressed in https://github.com/travis-ci/docs-travis-ci-com/commit/1fd07be5ce29173f80c9962aadb0bcf953ad609e

plaindocs:
I think this is resolved. Please reopen if you think there is something that remains unaddressed.

PR #905, opened 2017-01-09
More cache clarification

plaindocs:
Fix for #416

I think this covers it. @MariadeAnton want to confirm?

PR #1078, opened 2017-04-01
Clarifying cache

li141319:
Regarding https://github.com/travis-ci/docs-travis-ci-com/issues/416

@MariadeAnton @damianavilla how does this look?


BanzaiMan:
This is a redo of #905.

Not sure what @li141319 is doing, but we don't need this.

PR #1079, opened 2017-04-02
Add --priviledged

derickr:
This is needed for tools like GDB and Valgrind to work properly.

### DIFF
--- a/user/common-build-problems.md
+++ b/user/common-build-problems.md
@@ -478,7 +478,7 @@ system services to run correctly, the container must be run with `/sbin/init` as
 PID 1:
 
 ``` bash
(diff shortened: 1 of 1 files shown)


plaindocs:
Thanks. Aside from the typo in the flag this looks good.

Does it have any other implications?

Ie, should we recommend it IF you need to run valgrind etc, or should we just put it in all

plaindocs:
I think this isn't relevant to most users, so closing for now. I'll rethink if it crops up again.

derickr:
This is always required if you want to use valgrind or gdb. Without it, the image is useless. There are no other implications besides of course that the container will have a few more privileges. Because the document describes debugging something in a container, I would very much welcome this be added always.

PR #1085, opened 2017-04-05
Clarify repo option for deployments

iriberri:
I've seen that we're commonly using the term "slug" around the documentation, so I thought it would be interesting to use it here too as the "name along with the owner" mention seems to be confusing.

What do you think @plaindocs?


plaindocs:
lovely!

PR #1086, opened 2017-04-05
Better  `fqdn` default description for pages deployment

mbauman:
It's not so much that there's no default, but rather that the default is to _change_ the current GitHub custom domain setting to remove custom domain support.

It might be a bit friendly if an empty `fqdn` setting didn't change the GitHub configuration.


plaindocs:
Thanks for the doc fix.

You can suggest the empty setting fix or submit a PR here: https://github.com/travis-ci/dpl

PR #1101, opened 2017-04-12
Document status update limit on GitHub

plaindocs:
* Does what it says
* I also fixed a broken merge, so an extra set of eyes on the final page would be awesome
* I'm slightly concerned about the findability of the API issue, but we can revisit that


MariadeAnton:
LGTM 🚢  - thank you both!

PR #1107, opened 2017-04-15
Updating deployment script ruby version

kmanzana:


user/deployment/script.md
```diff
@@ -53,11 +53,11 @@ deploy:
     all_branches: true
 ```
 
-## Deployment is executed by Ruby 1.9.3
+## Deployment is executed by Ruby 2.2.0
```
BanzaiMan: This is not exactly correct. The correct answer is "depends". Sorry.

This depends on the build environment. It would be quite confusing to explain *what* version is used in which, so I suggest using a different phrase, for example:

> Deployment may be executed by a Ruby runtime different from the one you used for the build.

user/deployment/script.md
```diff
@@ -53,11 +53,11 @@ deploy:
...
 
 In order to ensure that deployments are executed reliably, we use a version of Ruby that is pre-installed on all of our build images.
 
-Currently, this is Ruby 1.9.3.
+Currently, this is Ruby 2.2.0.
```
BanzaiMan: This can be rephrased as something like this:

----

You can find out what version is used for deployment by the `travis_internal_ruby` function.
If you want to perform preparation tasks for deployment, you can do so by:

```yaml
before_deploy:
  - rvm $(travis_internal_ruby) do gem install fog-core -v 1.43.0 # change "gem …" to suit your needs
```

PR #1111, opened 2017-04-18
Add Swift as its "own" language

emdantrim:
A tiny PR to make things a bit smoother for Swift users.


plaindocs:
Nice! :ship: :it: !

PR #1127, opened 2017-04-27
Slight improvements to getting started WIP

plaindocs:
In particular, where should we send people afterwards?


user/getting-started.md
```diff
@@ -4,11 +4,7 @@ layout: en
 permalink: /user/getting-started/
 ---
 
-The very short guide to using Travis CI with your GitHub hosted code repository. Totally new to all this? Try reading [Travis CI for Complete Beginners](/user/for-beginners) instead.
-
-Or if you want a more complete guide to a particular language, pick one of these:
-
-{% include languages.html %}
+The very short guide to using Travis CI with your GitHub hosted code repository. Totally new to continuous integreation and Travis CI? Start with [Travis CI for Complete Beginners](/user/for-beginners) instead.
```
mhinz: Typo: `integration`

And since this paragraph is getting touched already, maybe we can omit the `Totally` as well. :-)

plaindocs: We should always remove 'totally', and typos!

user/getting-started.md
```diff
@@ -4,38 +4,107 @@ layout: en
...
-{% include languages.html %}
+## To get started with Travis CI
 
-### To get started with Travis CI:
+You'll need a GitHub account with admin access to at least one repository, and ideally you'll already have a working build script that you run manually.
 
 1. [Sign in to Travis CI](https://travis-ci.org/auth) with your GitHub account, accepting the GitHub [access permissions confirmation](/user/github-oauth-scopes).
 
-2. Once you're signed in, and we've synchronized your repositories from GitHub, go to your [profile page](https://travis-ci.org/profile) and enable Travis CI for the repository you want to build.
+2. Once you're signed in, and we've synchronized your GitHub repositories, go to your [profile page](https://travis-ci.org/profile) and enable Travis CI for the *open source* repository you want to build. If you want to build a private repository, sign in to [Travis CI for private repositories](https://travis-ci.com/profile) instead.
```
acnagy: I think the comment about admin access being required should show up here? It's a good thing to note in my opinion - not sure we need to get rid of it?

plaindocs: I moved it to before the list, 

> You’ll need a GitHub account with admin access to at least one repository, and ideally you’ll already have a working build script that you run manually.

But can leave it there as well if that makes sense.

user/getting-started.md
```diff
@@ -4,38 +4,107 @@ layout: en
...
+## Selecting a programming language
+
+Use one of these common languages:
+
+```yaml
+language: ruby
+language: java
+language: node_js
+language: android
+language: php
```
acnagy: Can we add python? We do get a lot of python 🐍

user/getting-started.md
```diff
@@ -4,38 +4,107 @@ layout: en
...
+
+* If you need a more up-to-date version of Ubuntu on the same infrastructure, use
+the beta of Ubuntu Linux Trusty 14.04:
+
+   ```yaml
+   sudo: false
+   dist: trusty
+   ```
+
+* If you need a more customizable, fully virtualized environment, use the Sudo
```
acnagy: Should we mention this environment is a vm? It does make a difference

user/getting-started.md
acnagy: I think the comment about admin access being required should show up here? It's a good thing to note in my opinion - not sure we need to get rid of it?

user/getting-started.md
```diff
@@ -4,38 +4,107 @@ layout: en
...
+  ```
+
+* If you have tests that need to run on macOS, or your project uses Swift or
+Objective-C, use our OSX environment:
+
+  ```yaml
+  os: osx
+  ```
+
+  > You do *not* necessarily need to use macOS if you develop on a Mac, only if
```
acnagy: Thank you! 🙌

user/getting-started.md
```diff
@@ -4,38 +4,107 @@ layout: en
...
+  > you need Swift, Objective-C or other macOS software.
+
+## After running tests
+
+After a succesful build there are many things you can do with the results of your code:
```
acnagy: "successful" could mean "passing" or "valid" - I'm assuming we want "valid"? Because you can send notifications `on_failure` and whatnot, though everything else needs to pass, in general (unless you have a creative `script`)

user/getting-started.md
BanzaiMan: I am not sure about having this discussion here. Selection of infrastructure should not be a concern for the very first build.

user/getting-started.md
BanzaiMan: 👍 for adding Python.

We should drop Android from this list, given that it requires extra configuration in `android.components`.

Also, there should not be multiple `language` key-value pairs in `.travis.yml`, so I don't think we should present them like this.

Instead:

```yaml
language: ruby
```

```yaml
language: java
```

```yaml
language: node_js
```

```yaml
language: php
```

```yaml
language: python
```

user/getting-started.md
```diff
@@ -4,38 +4,107 @@ layout: en
...
 1. [Sign in to Travis CI](https://travis-ci.org/auth) with your GitHub account, accepting the GitHub [access permissions confirmation](/user/github-oauth-scopes).
 
-2. Once you're signed in, and we've synchronized your repositories from GitHub, go to your [profile page](https://travis-ci.org/profile) and enable Travis CI for the repository you want to build.
+2. Once you're signed in, and we've synchronized your GitHub repositories, go to your [profile page](https://travis-ci.org/profile) and enable Travis CI for the *open source* repository you want to build. If you want to build a private repository, sign in to [Travis CI for private repositories](https://travis-ci.com/profile) instead.
 
-   > Note: You can only enable Travis CI builds for repositories you have admin access to.
+3. Add a `.travis.yml` file to your repository to tell Travis CI what to build.
 
-3. Add a `.travis.yml` file to your repository to tell Travis CI what to build:
+   This example tells Travis CI that this is a Ruby project, so unless you change the default, Travis CI uses `rake` to build it.
```
iriberri: Perhaps we can also add a mention for the default `install` behavior, similar to the `script` stage?

user/getting-started.md
plaindocs: I think giving people an idea that there are various different environments is useful. IF it doesn't go here people won't encounter it. 

I've marked it as optional, but I'm also open to the idea of removing it entirely if people feel strongly enough.

plaindocs:
OK, I think this is good enough to merge, even if there is still work to be done. 

@acnagy I think the notice about admin access before the list of steps is sufficient. WDYT?

PR #1140, opened 2017-05-03
Add advice about recent hex.pm incompatibility

scouten:
Resolves the issue raised in https://twitter.com/scouten/status/859618158382129156.

### DIFF
--- a/user/languages/elixir.md
+++ b/user/languages/elixir.md
@@ -49,6 +49,33 @@ matrix:
       otp_release: 18.0
 ```
 
+## Trusty Beta Support Recommended
+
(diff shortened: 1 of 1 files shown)


joepvd:
Thanks for opening this!

As an alternative workaround, you might want to try adding the following to the configuration:

```
addons:
    apt:
        packages:
            - libssl1.0.0
```

This enables the download of the binary again.

We are looking into doing this by default, so it might be that the workaround really is a temporary fix.

cotsog:
Thanks for your input, @scouten!

FYI, we are publicly tracking this issue in https://github.com/travis-ci/travis-ci/issues/7688.

We'll keep your PR open until this is fixed for real.

scouten:
I've just confirmed that the config change is no longer necessary in my library. Thanks for fast response, everyone!

Closing this PR.

PR #1155, opened 2017-05-08
Rust: add example for workspace projects

edmccard:
Rust projects can define a workspace which is a set of crates that will all share the same Cargo.lock and output directory. Without modifying the build script to pass `--all` to the various build commands, only the root crate will be built and tested. Or, if the workspace has no root crate, the build and test commands will fail with the following error:

error: manifest path `/path/to/crate/root` is a virtual manifest, but this command requires running against an actual package in this workspace


plaindocs:
Thanks!

PR #1156, opened 2017-05-09
Update best-practices-security.md

Gnouc:
Since when this file is for security purpose, it's better to use best practice.

Leaving variable un-quoted in shell script is bad practice and is source of many security implications, please do not make the user think it's the right way.


plaindocs:
Thanks!

PR #1157, opened 2017-05-09
Build stages

svenfuchs:


plaindocs:
This is great stuff! Nice screen cast also!

user/build-stages.md
```diff
@@ -0,0 +1,186 @@
...
+        # ...
+```
+
+Travis CI does not set or overwrite any of your scripts, and most languages
+have a [default test script](https://docs.travis-ci.com/user/languages/ruby/#Default-Test-Script)
+defined. So in many use cases you might want to overwrite the `script` by
+specifying the keyword `skip` or `ignore`.
+
+## Examples
+
```
plaindocs: I cut the mini menu, because you're just linking to the heading, and then again to the page, so it feels like an extra jump.

plaindocs:
OK, this is ready to roll as far as I'm concerned. 

@MariadeAnton wanna take a last look? 

@svenfuchs merge at will

user/build-stages.md
```diff
@@ -0,0 +1,160 @@
+---
+title: Build Stages
+permalink: /user/build-stages/
+layout: en
+---
+
+<div id="toc"></div>
+
+> Build stages are still in BETA. There is more information about what this means, and how you can give us feedback on this new feature in the [GitHub issue](https://github.com/travis-ci/beta-features/issues/11).
```
MariadeAnton: do we have any thoughts of using `currently` as opposed to `still`  in BETA?

plaindocs: Valid point. You *could* construe still as having been there a while, so maybe currently is a better choice. We use `still` on at least one of the other beta notices.

svenfuchs: I've changed this to `currently`. Thanks @MariadeAnton!

user/build-stages.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Naming your stages
+
+Stage names are arbitrary strings, and can include spaces or emojis. The first
+letter of a stage name is automatically capitalized for aesthetical reasons, so
+you don't have to deal with uppercase strings in your `.travis.yml` file.
+
+The default stage is `test`. Jobs that do not have a stage name are assigned to
+the previous stage name if one exists, or the default stage name if there is no
+previous stage name. `test`. This means that if you set the stage name on the
```
iriberri: Not sure if the `test` mention in this line is a typo (is it?), since we already define the default stage name two lines above.

user/build-stages.md
```diff
@@ -0,0 +1,169 @@
...
+jobs:
+  include:
+    - stage: deploy
+      rvm: 2.4
+      env:
+        - FOO=foo
+      script: ./deploy
+```
+
+This will run two jobs with on Ruby 2.3 and 2.4 respectively first, and assign
```
iriberri: Shouldn't this be "This will run two jobs ~with~ on Ruby 2.3 and 2.4" instead?

user/build-stages.md
```diff
@@ -0,0 +1,169 @@
...
+    - stage: deploy
+      script: skip     # usually you do not want to rerun any tests
+      deploy: &heroku
+        provider: heroku
+        # ...
+```
+
+Travis CI does not set or overwrite any of your scripts, and most languages
+have a [default test script](https://docs.travis-ci.com/user/languages/ruby/#Default-Test-Script)
+defined. So in many use cases you might want to overwrite the `script` by
```
iriberri: I like this a lot! We got a question in Support about skipping the `install` step in one of the stages. Would it make sense to mention that it's possible to skip/redefine any of the "stages" (`build`, `script`...) of the build in a Build Stage?

iriberri:
Here's the "redefining steps in stages" example that I mentioned before after this came up in Support:

```
language: ruby
rvm: 2.3.1

env:
- TEST_SUITE=integration_tests
- TEST_SUITE=unit_tests

script: bundle exec rake test:$TEST_SUITE
 
jobs:
  include:
    - stage: custom deploy script
      env: TEST_SUITE=custom_deploy_script
      install: skip # bundle install is not required
      script: ./custom_deploy_script.sh
```

If you think this would be a nice addition, I generated an example page for it in https://gist.github.com/iriberri/ebff2f5a3544a97ac9f14d934a2907b8, using `user/build-stages/deploy-rubygems.md` and the other examples as a template ([see RAW version of this new example](https://gist.githubusercontent.com/iriberri/ebff2f5a3544a97ac9f14d934a2907b8/raw/1e7545e9cc5c165082ca578d194443e299d54e20/build-stages-demo.md))

iriberri:
Pheewww... sorry about that 😥

user/build-stages.md
svenfuchs: 👍 

I've made that:

> So in many use cases you might want to overwrite the `script` step by
specifying the keyword `skip` or `ignore`, in other cases you might want to 
overwrite other steps, such as the `install` step that runs by default on 
several languages.

user/build-stages.md
svenfuchs: 👍

user/build-stages.md
svenfuchs: 👍 

totally, thanks for catching this! i've removed it

svenfuchs:
Thanks for the review and the example, @iriberri! I've addressed the typos. Will look at the example next.

svenfuchs:
@iriberri given the discussion about the "job stages" semantics here https://github.com/travis-pro/team-sapphire/issues/613#issuecomment-300680388 ... shall we wait with adding your example until we have settled on a solution to the semantics?

svenfuchs:
@iriberri after a discussion with @MariadeAnton we decided that we can address the semantics later. i have now added your example, and added another one that uses YAML aliases.

PR #1160, opened 2017-05-11
Build stages: pull in some doc bits from the blog post

svenfuchs:


user/build-stages.md
```diff
@@ -9,45 +9,79 @@ layout: en
 > Build stages are currently in BETA. There is more information about what this means, and how you can give us feedback on this new feature in the [GitHub issue](https://github.com/travis-ci/beta-features/issues/11).
 {: .beta}
 
-With this new feature you can group jobs together in 'stages'. Jobs in a stage
-run in parallel, and the stages themselves run sequentially, one after another.
-The build fails when any stage fails, i.e. if any job in this stage fails.
+# What are Build Stages?
```
plaindocs: we start at H2 within the docs, otherwise you end up at really *weird* tocs

user/build-stages.md
```diff
@@ -9,45 +9,79 @@ layout: en
...
 
-For example, you can configure a deployment stage to run only after several test
-jobs have all completed successfully.
+In the simplest and most common use case, you can now make one job run _only_
```
plaindocs: This does not actually answer the question in the heading

plaindocs: Maybe something like

Build stages is a way to group jobs, run them in parallel, and then run other jobs depending on the overall return status of the group.

user/build-stages.md
```diff
@@ -9,45 +9,79 @@ layout: en
...
 
-![Example screencast](https://cloud.githubusercontent.com/assets/3729517/25229553/0868909c-25d1-11e7-9263-b076fdef9288.gif)
-
-## Naming your stages
+Stages are identified by their names.
 
-Stage names are arbitrary strings, and can include spaces or emojis. The first
-letter of a stage name is automatically capitalized for aesthetical reasons, so
-you don't have to deal with uppercase strings in your `.travis.yml` file.
+You are free to pick any arbitrary names you like and include emojis.  The
```
plaindocs: I'm not so keen on 

> You're free to pick

svenfuchs: Ok? I've now used:

> You can use any arbitrary names you like ...

Is that better? What else would be better?

user/build-stages.md
svenfuchs: Good point. Added one sentence in the beginning:

> Build stages is a way to group jobs, and run jobs in each stage in parallel, but run one stage after another sequentially.

user/build-stages.md
svenfuchs: Uh, of course. Fixed.

svenfuchs:
Thank you, @plaindocs ❤️

PR #1166, opened 2017-05-15
Add bundle_type parameter

nikhilo:


user/deployment/codedeploy.md
```diff
@@ -17,6 +17,7 @@ For a minimal configuration with S3, add the following to your `.travis.yml`:
         secret_access_key: "YOUR AWS SECRET KEY"
         bucket: "S3 Bucket"
         key: latest/MyApp.zip
+        bundle_type: zip
```
BanzaiMan: `bundle_type` is detected, so it should not appear in the "minimal" configuration.

BanzaiMan:
Also, please either rebase or merge the current `gh-pages` branch, so that the build passes. Thanks.

user/deployment/codedeploy.md
nikhilo: Makes sense, removing that from minimal config

PR #1167, opened 2017-05-15
eb: be more explicit about implicit zip file generation

joepvd:


alexspeller:
👍 Thanks @joepvd this would've saved me some time :)

PR #1171, opened 2017-05-16
Remove the $ before GITHUB_TOKEN

schalkneethling:
The `$` that precedes `GITHUB_TOKEN` was causing an error at build time


BanzaiMan:
How does the issue manifest?

schalkneethling:
Nope, it was a red herring, the problem was elsewhere. Closing.

PR #1173, opened 2017-05-16
Describe --add flag

schalkneethling:
Adds a note about the `--add` flag to automatically add the secure token to `travis.yml`

### DIFF
--- a/user/encryption-keys.md
+++ b/user/encryption-keys.md
@@ -35,9 +35,7 @@ This will output a string looking something like:
 secure: ".... encrypted data ...."
 ```
 
-Now you can place it in the `.travis.yml` file.
-
-> Pro Tip: You can also have travis add it automatically by running the command as follows:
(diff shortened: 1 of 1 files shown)


user/encryption-keys.md
```diff
@@ -37,6 +37,11 @@ secure: ".... encrypted data ...."
 
 Now you can place it in the `.travis.yml` file.
 
+> Pro Tip: You can also have travis add it automatically by running the command as follows:
```
plaindocs: Thanks for this.

We avoid phrases like "Pro Tip" or "Just". And always use Travis CI not Travis.

How about running the text on onto the previous sentence like

 Now you can place it in the `.travis.yml` file. Or add it to your `.travis.yml` by running `travis encrypt SOMEVAR="secretvalue" --add` instead.

schalkneethling: Thanks @plaindocs I will make the update.

plaindocs:
Thanks!

PR #1180, opened 2017-05-18
recommend panda instead of zef

robertlemmen:
also removed the versioned-panda paragraph. Is that still useful? It is also less simple with zef as the versions do not line up...


BanzaiMan:
Closing this in favor of #1190.

PR #1182, opened 2017-05-19
Add debug VM usage instructions

BanzaiMan:
This PR adds instructions on how to use debug VM.


user/using-debug-vm.md
```diff
@@ -13,6 +13,9 @@ you can restart builds in the debug mode to gain access to the actual VM or cont
 
 This feature is available for private repositories and those public repositories for which
 the feature is enabled.
+
+In addition, you need the push access to the repository is required.
```
mhinz: Something gone wrong here. :>

```
In addition, push access for the repository is required.
```
or
```
Furthermore, you need push access for the repository.
```

BanzaiMan: Thanks!

user/using-debug-vm.md
```diff
@@ -0,0 +1,168 @@
...
+
+When the build failure is only observable on Travis CI, and if you suspect that there are
+critical differences between your local development environment and Travis CI's build VM,
+you can restart builds in the debug mode to gain access to the actual VM or container
+(henceforth "debug VM").
+
+# Restarting a build to debug
+
+This feature is available for private repositories and those public repositories for which
+the feature is enabled.
```
joepvd: Indicate what to do to get the feature enabled.

user/using-debug-vm.md
```diff
@@ -0,0 +1,168 @@
...
+Debug build initiated by BanzaiMan
+Setting up debug tools.
+Preparing debug sessions.
+Use the following SSH command to access the interactive debugging environment:
+ssh DwBhYvwgoBQ2dr7iQ5ZH34wGt@ny2.tmate.io
+This build is running in quiet mode. No session output will be displayed.
+This debug build will stay alive for 30 minutes.
+```
+
+Running the `ssh` command above will drop you in on a live VM.
```
joepvd: This assumes that users have ssh with a key pair configured. I believe one cannot connect when no key pair is available.

user/using-debug-vm.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Security implications of the `ssh` command
```
joepvd: Not sure about the title. It is not the `ssh` command that is making troubles, but having it publicly available.

Maybe: _Security implications of running debug builds_  ?

user/using-debug-vm.md
```diff
@@ -0,0 +1,168 @@
...
+
+However, this also means that anyone who can see the command while the debug
+VM is live can connect to the debug VM and potentially look at your secrets.
+This is the reason that the feature is not available by default on public
+repositories.
+
+Do note that when another `ssh` connection commences, that new connection
+will be attached to the session you are already in.
+If you suspect nefarious activity, you can either close all your windows, or cancel
+the debug job on the Web UI to disconnect the potential attacker.
```
joepvd: `commence` and `nefarious` is maybe not clear for every non-native English speaker.

user/common-build-problems.md
```diff
@@ -527,3 +527,11 @@ which Docker image you are using on Travis CI.
 1. Manually install dependencies, if any.
 
 1. Manually run your Travis CI build command.
+
+### Running debug VMs
+
+In private repositories and those public repositories for which the feature is enabled,
+it is possible to run a debug VM.
+Using the debug VM, you can interact with the VM.
```
iriberri: This sentence reads funny to me :)

However, I've been trying to think for a alternative and it's not been easy. What about... "Using the debug VM [feature], you can interact with the virtual environment where your build runs."?

user/common-build-problems.md
```diff
@@ -527,3 +527,11 @@ which Docker image you are using on Travis CI.
...
+
+For more information, please consult [the debug VM document](/user/using-debug-vm/).
```
iriberri: "the debug VM documentation"?

user/using-debug-vm.md
BanzaiMan: That would imply that there is something inherently insecure about the debug build itself, which is not true.

Maybe: Security implications of public display of the `ssh` command

user/using-debug-vm.md
```diff
@@ -0,0 +1,170 @@
...
+
+For private repositories, the API endpoint is `https://api.travis-ci.com/job/${id}/debug`.
+
+The JSON payload could be empty (`-d "{}"`), in which case any command input and output
+you see while connected to the debug VM will be reflected on the build log.
+
+# Connecting to the debug VM
+
+Once the debug VM starts, it will go through the intial steps, including setting up language
+run times and installing debug utilities.
```
iriberri: Perhaps clarify if the cache has been set up too?

user/using-debug-vm.md
```diff
@@ -0,0 +1,170 @@
...
+(henceforth "debug VM").
+
+# Restarting a build to debug
+
+This feature is available for private repositories and those public repositories for which
+the feature is enabled.
+To have the feature enabled on a public repository, please email us at support@travis-ci.com
+indicating which ones.
+
+In addition, the push access to the repository is required.
```
iriberri: What about adding this requirement as a note/warning? Perhaps right before the `Starting a debug VM via API` section or at the beginning of this one.

I'd rephrase it to something similar to: It is required to have push access to the repository to start a debug build.

user/using-debug-vm.md
BanzaiMan: Or, simply: Security considerations

BanzaiMan:
I just realized that the changes to common-build-problems.md should go to the ERb template.

user/using-debug-vm.md
BanzaiMan: I feel that these words should be common enough to use here.

user/using-debug-vm.md
```diff
@@ -0,0 +1,169 @@
...
+You may choose to share the `ssh` command with another person who might be
+able to help you out in debugging the build.
+
+However, this also means that anyone who can see the command while the debug
+VM is live can connect to the debug VM and potentially look at your secrets.
+This is the reason that the feature is not available by default on public
+repositories.
+
+Do note that when another `ssh` connection commences, that new connection
+will be attached to the session you are already in.
```
BanzaiMan: Maybe we can add a little bit about what it means to have a shared session. That is to say, mention that the user can see what the other person is doing on the session. For example, if another person manages to dump the environment variables with `env`, the user will see that, too. (Thus giving an indication that the secrets are now at risk, and appropriate measures can be taken to minimize the damage.)

user/using-debug-vm.md
```diff
@@ -0,0 +1,169 @@
...
+For private repositories, the "Debug build" or "Debug job" button is available on the upper right corner of
+the build/job page.
+
+![Screenshot of debug build/job buttons](/images/debug_buttons.png)
+
+For public repositories, an API call is required, in addition to the feature being enabled.
+
+## Starting a debug VM via API
+
+In addition to using the debug button, a debug build may be started via
```
plaindocs: We've already talked about the buttons above, so we can just 

> To start debug mode via an API call, set `Authorization: token *************` to your [Travis CI API token](https://docs.travis-ci.com/user/triggering-builds/) and `${id}` to the job you want to debug. This is the *only* way to start debug mode on a public build.

> To leave command input and output from your debug session in the build log, set the JSON payload to empty (`-d "{}"`).

user/using-debug-vm.md
plaindocs: > To connect to the debug virtual machine run the ssh command shown in the build log. If you don't have an ssh key configured you'll need to [generate one](/user/using-debug-vm/#%E2%80%9CPermission-denied%E2%80%9D-error-message) before connecting.

user/using-debug-vm.md
plaindocs: I tend to agree with both of you, but I also don't think commence and nefarious convey more information than start and suspicious.

user/using-debug-vm.md
plaindocs: :+1:

user/using-debug-vm.md
```diff
@@ -0,0 +1,169 @@
...
+Are you sure you want to continue connecting (yes/no)? yes
+Warning: Permanently added 'ny2.tmate.io,104.236.9.236' (ECDSA) to the list of known hosts.
+Permission denied (publickey).
+```
+
+If you see this, run the command `ssh-keygen` (go through the prompts), and try again.
+
+# Things to do once you are inside the debug VM
+
+Once inside the debug VM, you are able to poke around.
```
plaindocs: s/poke around/explore/ ?

user/using-debug-vm.md
```diff
@@ -0,0 +1,169 @@
...
+
+You are in a [`tmate`](https://tmate.io/) session, at the point
+where your `before_install` (even if it is not defined) phase is about to
+start.
+
+## Convenience `bash` functions
+
+Various build phases are defined in convenience `bash` functions named
+`travis_run_*` (e.g., `travis_run_before_install`, `travis_run_install`).
+At the most basic level, you can simply run these build phases to observe your build:
```
plaindocs: ~~At the most basic level, you can simply run these build phases to observe your build~~

Run each of these build phases in turn to observe your build.

> QUESTION can you rereun phases?

user/using-debug-vm.md
```diff
@@ -0,0 +1,169 @@
...
+In this mode, you can move your cursor with your arrow keys to go through the
+log history.
+
+Press `q` to exit the log scroll mode.
+
+## Getting out of the debug VM
+
+Once you exit from all the live `tmate` windows, the debug VM will terminate
+after resetting the job's status to the original status.
+No more stages (`before_install`, `install`, etc.) will be executed.
```
plaindocs: By original status you mean before you ran the debug? 

Also s/stages/phases/

Rakefile
```diff
@@ -54,6 +54,23 @@ task :run_html_proofer do
...
+
+  PREREQ_REMOTE_FILES.each { |f| `curl -OsSfL #{f}` }
+
+  files = Find.find(File.join(File.dirname(__FILE__), 'user')) do |f|
+    if File.file?(f) && File.fnmatch('*.erb', f)
+      `erb -r json #{f} | tee #{f.gsub(/\.erb$/,'')}`
+    end
+  end
+end
+
```
BanzaiMan: Maybe! I am not familiar with all the standard Jekyll tools, so I reached out for the most obvious one to me, which is `rake`.

Do you want to change this now, or should we discuss it in a separate PR?

plaindocs: It is less important than the content, but definitely something we should address, as it adds considerable complexity over all. So we can raise a PR after merging this.

plaindocs: We'd still need the rake task to pull the info down, but if we pull it down, convert to yaml and drop it in the `_data` dir, the accessing of it is then part of standard jekyll.

user/using-debug-vm.md
BanzaiMan: It is certainly possible to rerun phases. They don't have to be run in the defined order, either. Running multiple times, or out of order may result in unexpected behavior, but that is up to the user.

Rakefile
BanzaiMan: OK. I'll handle this in a separate PR.

plaindocs:
Thanks for all of your work on this @BanzaiMan ! :tada:

plaindocs:
@joepvd @meatballhat  the yaml include mechanism used in this PR could easily be adapted to runtime version numbers etc from the environment packer templates. Thoughts? :smiling_imp:

PR #1184, opened 2017-05-20
Adds warning that provider is experimental.

ojongerius:
Adds warning that provider is experimental.

Since https://github.com/travis-ci/dpl/commit/c6696b38d2c5e5941d2aa51edbac94b066a9c4d2  deployments will show:
> !!! GitHub Pages support is experimental !!!

We should either remove the experimental warning, or add something to the documentation -the warning took me by surprise. Not sure what the warning should say, maybe this will do, I'll leave that up to your judgement :smile:

/cc @BanzaiMan

### DIFF
--- a/user/deployment/pages.md
+++ b/user/deployment/pages.md
@@ -4,7 +4,7 @@ layout: en
 permalink: /user/deployment/pages/
 ---
 
-> Deploying to GitHub Pages uses `git push --force` and overwrites the history on the`gh-pages` branch, which is usually only used for deployments. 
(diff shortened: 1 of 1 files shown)


plaindocs:
@MariadeAnton got an idea about this?

plaindocs:
@MariadeAnton @BanzaiMan @ojongerius any opinions on the new wording?

Although I suspect the force push is to whatever you set the target branch to, which means if you misconfigure you *could* blow away master :-p.

PR #1190, opened 2017-05-23
Update Perl 6 docs

paultcochrane:
This PR gives the Perl 6 docs a much needed update.  The biggest change here is to replace the discontinued module installer (`panda`) with the current module installer (`zef`).

### DIFF
--- a/user/languages/perl6.md
+++ b/user/languages/perl6.md
@@ -23,12 +23,12 @@ versions that your projects can be tested against. To specify them, use the
 language: perl6
 perl6:
   - latest
-  - '2015.07'
-  - '2015.04'
+  - '2017.05'
+  - '2017.04'
 ```
 
 Over time, new releases come out and we upgrade both rakudobrew and
-Perls, aliases like `2015.07` will float and point to different exact
+Perls, aliases like `2017.05` will float and point to different exact
 versions, patch levels and so on.
 
 For precise versions pre-installed on the VM, please consult "Build system
@@ -136,8 +136,9 @@ perl6:
 ```yaml
 language: perl6
 perl6:
-    - latest
-    - '2015.03'
+    - '2017.05'
+    - '2017.04'
+
 install:
     - rakudobrew build-panda ${TRAVIS_PERL6_VERSION#latest}
     - panda installdeps .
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Thank you!

zoffixznet:
@paultcochrane I see my suggestion for `zef --debug install .` got in

My fault for not bothering to read anything. I didn't realize there was a "default test" script; and currently [some people](https://github.com/Bailador/Bailador/pull/104#discussion_r118266805) pointing out that by installing the module the test script is not running in pristine environment.

However, there are problems with the test script itself:
- It uses `prove` but for several months `zef` hasn't been using it and instead using TAP::Harness module
- It doesn't test *installation* of the module succeeds. So if the user missed a file in `provides` META file section, I don't think the current default test script would catch that.

I think the default test script should install [`TAP::Harness`](https://modules.perl6.org/repo/TAP::Harness) module and then do `zef --debug install .` instead of running `prove`. 

(sidenote: why install the module if it's shipped with Rakudo? Because we'll throw it out of core for next release, leaving the ecosystem version as the only available version).

paultcochrane:
@zoffixznet my apologies that modules are breaking as a result of this change!  Since it was just a documentation change, and since the recommended process with `zef` looked well tested, I thought it should be ok.

The reason why `prove` is still used is because the script hasn't been updated since mid to late 2015, so it needs a bit of TLC.  I'll update the script, submit a PR, and send you a ping for a code review before it goes live.  The PR will be in https://github.com/travis-ci/travis-build as that's where the build script lives.

Question: if `zef --debug install .` should be in the default test script, then it doesn't need to be in the docs, right?  In other words, if the user chooses `perl6` as the language, then the script should "just run", right?  Sorry if these are silly questions; I'm a bit rusty wrt perl6 stuff atm.

zoffixznet:
> my apologies that modules are breaking

No, they aren't breaking. Just someone commented on another module 😄 

> Question: if zef --debug install . should be in the default test script, then it doesn't need to be in the docs, right?

Right, because the test script would install it. Also, I'm now realizing that `zef --debug install .` doesn't give verbose test output. Trying to find the way to make it do so.

I'm not really an expert on this stuff... hoping someone else will jump in and advise :)

zoffixznet:
@paultcochrane looks like [there's currently no good way to increase verbosity](https://github.com/perl6/tap-harness6/issues/14) for zef tests with TAP::Harness. So, never mind about changing the build script, for now at least.

PR #1194, opened 2017-05-24
Add bits on segmentation fault caused by corrupt cache

BanzaiMan:


user/common-build-problems.md
```diff
@@ -58,6 +58,14 @@ likely to show similar causes. It can be caused by memory leaks or by custom
...
 possible. Dialing these numbers down should help.
 
+## Segmentation faults from the language interpreter (Ruby, Python, PHP, Node.js, etc.)
+
+If you see your `ruby`, `python`, and other language interpreters which allows
+extension codes (gems, modules, etc.) start running into segmentation faults,
+you may have corrupt (or invalid) cache.
+Try disabling (i.e., removing `cache` from `.travis.yml`), or deleting the cache
+(https://travis-ci.org/OWNER/REPO/caches).
+
```
plaindocs: This is a tricky one. Do you think it is worth adding a notice in each language page?

How about rewording to something like

If your build is failing due to unexpected segmentation faults in the language interpreter, this may be caused by corrupt or invalid caches of your extension codes (gems, modules, etc).  This can happen with any interpreted language, such as Ruby, Python, PHP, Node.js, etc.

Fix the problem by [clearing the cache](/user/caching#Clearing-Caches) or removing the `cache` key from your `.travis.yml` (you can add it back in a subsequent commit).

plaindocs:
Ping @BanzaiMan

PR #1195, opened 2017-05-24
Add --depsonly to pre-testing installation

zoffixznet:
This way only module dependencies get installed and the `prove`
command tests the content of the repo, without having the very
same module already installed on the system.


plaindocs:
Thanks!

PR #1196, opened 2017-05-25
Add blurb on `jobs.include` + matrix expansion

BanzaiMan:
This seems a common error.

This also echoes what we have in https://docs.travis-ci.com/user/customizing-the-build/#Explicitly-Included-Jobs-need-complete-definitions.


plaindocs:
I'm gonna merge this in, but we probably still need to do a better job explaining this.

Issue #1199, opened 2017-05-25
Update Webhooks Delivery Format JSON


iriberri:
Comes from: https://github.com/travis-ci/docs-travis-ci-com/pull/998#issuecomment-283725257

The webhooks delivery format example is not up-to-date: https://docs.travis-ci.com/user/notifications#Webhooks-Delivery-Format

Currently, the following payload is shown in the docs: https://gist.github.com/roidrage/9272064

iriberri:
Closing as this is taken care of in https://github.com/travis-ci/docs-travis-ci-com/pull/1203

PR #1203, opened 2017-05-26
Keep webhook payload up to date automatically

BanzaiMan:
This PR adds a POST endpoint, which is only valid at `/update_doc_webhook_payload`.

The idea here is that the docs app will receive the live (and hence always up-to-date) webhook payload, and update the gist (to be determined later) with that payload when we receive a POST request on the new endpoint.

This resolves https://github.com/travis-ci/docs-travis-ci-com/issues/1199.

### DIFF
--- a/_plugins/post_handler.rb
+++ b/_plugins/post_handler.rb
@@ -1,0 +1,30 @@
+require 'net/http'
+require 'base64'
+require 'openssl'
+
+class Handler
+  def call(env)
+    req = Rack::Request.new(env)
+    res = Rack::Response.new(["Not found"], 404)
+
+    return res unless req.post?
+
+    payload = JSON.load(req.body.read)['payload'].to_json
+    sig = Base64.decode64 env["HTTP_SIGNATURE"]
+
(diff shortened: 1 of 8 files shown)


_plugins/post_handler.rb
```diff
@@ -0,0 +1,30 @@
+require 'net/http'
+require 'base64'
+require 'openssl'
+
+class Handler
```
BanzaiMan: This name can be better. Much better.

BanzaiMan:
Test build with deployment: https://travis-ci.org/travis-ci/docs-travis-ci-com/builds/236736815, the revision https://api.github.com/gists/4e317d6e71be6d0278be46bb751b2f78/387e18892f1bbd99c1edd6c8adc2d51ef364fa24 and the result https://gist.github.com/travisbot/4e317d6e71be6d0278be46bb751b2f78/387e18892f1bbd99c1edd6c8adc2d51ef364fa24.

BanzaiMan:
The build failed because it refers to the source code where it does not yet exist (but will, when this PR is merged).

PR #1209, opened 2017-05-29
add show-language plugin to prism build

lislis:
shows snippet language automatically on hover. works automatically.
Looks like this out of the box

screen shot 2017-05-29 at 12 52 01
screen shot 2017-05-29 at 12 52 15


plaindocs:
Awesome!

The fiddly bit is displaying the filename, which is what requires the ugly CSS in Kramdown and the extra messing around. Do we think that is worth it?

lislis:
@plaindocs I thought this would be accomplished with `{:  data-language=" a filename that is not technically a language"}` after the code snippet? Or am I missing something?

plaindocs:
I think there are some extra steps, but I get a bit lost reading through

https://github.com/PrismJS/prism/pull/891
https://github.com/PrismJS/prism/issues/910

Maybe from [issue](https://github.com/PrismJS/prism/issues/910#issuecomment-198456129)

```css
pre[data-language]:before {
  content: attr(data-file);
  position: absolute;
  top: 0;
  right: 0;
}
```

is sufficient? Either way, just setting the CSS as you suggest isn't sufficient for it to display.

lislis:
@plaindocs I made this https://github.com/travis-ci/docs-travis-ci-com/pull/1235

PR #1211, opened 2017-05-29
Add env variables for AWS credentials in S3 deployment

crashuniverse:
Putting AWS credentials directly in `.travis.yml` file is not a great idea because then anyone can use the same. This PR adds basic security step of putting AWS credentials in env variables in travis-ci dashboard and later using it in `.travis.yml` file.


plaindocs:
Thanks for that @crashuniverse 

We do mention that you should encrypt credentials. and you can do that on the CLI or by adding them in the dashboard.

PR #1217, opened 2017-05-31
Update Scala versions, add note about using 2.12.2

MaxInertia:
- Updated versions listed to most recent releases: 2.10.6, 2.11.11, and 2.12.2
- Add note that using 2.12 requires overriding Travis-CI's default JDK (oraclejdk7) to oraclejdk8


MaxInertia:
I'm not entirely sure what the standards are for pull requests, if possible the commits should be squashed, I didn't realize each time I performed a 'branch update' that it would add an extra commit.

PR #1226, opened 2017-06-02
Use gulp-cli instead of gulp

badersur:
gulp recommends using gulp-cli: https://github.com/gulpjs/gulp/blob/master/docs/getting-started.md


plaindocs:
Thanks!

badersur:
Welcome @plaindocs :)

PR #1176, opened 2017-05-17
Re-write Beginner's Guide

acnagy:
See https://github.com/travis-ci/docs-travis-ci-com/issues/1145 for the discussion!

PR #1230, opened 2017-06-04
Update for-beginners.md

JuliusKoronci:
Since HHVM is only supported on trusty could you update the guide for absolute beginners to use dist: trusty, otherwise the build fails even when the tests are fixed

thanks


plaindocs:
That is a good point. 

We're restructuring that guide in https://github.com/travis-ci/docs-travis-ci-com/pull/1176 as well.

PR #1231, opened 2017-06-04
Removed 'master' from the first line.

zwrawr:
There seems to be the word `master` at the very beggining of your readme for no obvious reason. It is also stoping the `h1` on that line being parsed correctly.

image

This PR simpily removes the word, which should fix the broken rendering too.


BanzaiMan:
Thanks!

PR #1235, opened 2017-06-06
add styles for displaying file names on code snippets

lislis:
screen shot 2017-06-06 at 18 36 54

uses `{: data-file=".travis.yml"}` after the closing ```

@plaindocs how about this?


plaindocs:
YES! Awesomeness! Thanks @lislis !

What is your opinion on the original file extension display? Is it redundant now we're showing the file name? And should it only show on mouse over? My preference would be to show it always I think.

lislis:
@plaindocs I don't think it's redundant per se. We show the file extension automatically, but for the file name you'd need to add the extra markdown to the code snippet.
So we'd have general coverage automatically and can add more specific information where needed. I quite like it even though it happened almost by accident :D

I don't have a real opinion on the show-file-extension-on-hover. If you'd rather display it all the time I can do that. Just let me know!

plaindocs:
:100: !

Now *all* we need is to start marking up the snippets! 

Thanks for the *awesome* work.

lislis:
@plaindocs do you want me to fix the hover for file extensions in this PR, too?

plaindocs:
I don't feel very strongly, but tend to lean *against* hover in general.

plaindocs:
With out docs it would almost be worth making the filename `.travis.yml` unless otherwise specified...

plaindocs:
YEAH! :tada:

PR #1247, opened 2017-06-08
Mention how to install locales using apt addon

### LINKED ISSUE #1246: Common build problems: locales - add example to install locales via apt addon
At https://docs.travis-ci.com/user/common-build-problems/#System%3A-Required-language-pack-isn%E2%80%99t-installed, we mention that it's possible to install more locales in the environment with apt-get.

It's also possible to do so with the apt addon, that doesn't require to have sudo: required and would work for container-based environments. For example:

```yaml
addons:
  apt-get:
  - language-pack-de
```

What about adding a mention to the addon for the container scenario? :)

bz2:
Fixes #1246

Also fixes redirected link on the installing dependencies page.

Thanks to @iriberri for the pointer.

### DIFF
--- a/user/common-build-problems.md
+++ b/user/common-build-problems.md
@@ -356,6 +356,17 @@ before_install:
 The above addition will reinstall the en_US language pack, as well as the de_DE
 language pack.
 
+If you are running on the container-base infrastructure and don't have access
+to the `sudo` command, install locales [using the APT addon](/user/installing-dependencies/#installing-packages-with-the-apt-addon):
+
+```yaml
+addons:
+  apt:
+    packages:
+      - language-pack-en
+      - language-pack-de
+```
+
 ## Linux: apt fails to install package with 404 error
 
 This is often caused by old package database and can be fixed by adding the following to `.travis.yml`:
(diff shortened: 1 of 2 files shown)


user/common-build-problems.md
```diff
@@ -356,6 +356,16 @@ before_install:
 The above addition will reinstall the en_US language pack, as well as the de_DE
 language pack.
 
+Alternatively, the same packages can be installed [using the APT addon](/user/installing-dependencies/#installing-packages-with-the-apt-addon). This will work when `sudo` is not available due to running on the container-based infrastructure:
```
plaindocs: Thanks for this - could you switch the sentence round so it is more clear:

> If you are running on container-based infrastructure and don't have access to the sudo comment, install locales [using the APT addon](/user/installing-dependencies/#installing-packages-with-the-apt-addon).

bz2:
Used the wording you suggested.

The reason I had the text the other way around is the apt addon works regardless of vm type, and maybe wants to be the default recommendation at some point?

plaindocs:
Thanks @bz2 that makes total sense. 

I often recommend *not* trying to future proof like that, because it makes the logic harder to explain. If the addon becomes the default, we'll just reword again :-p

PR #1248, opened 2017-06-09
RENAME: docker-engine -> docker-ce. package docker-engine is deprecated

bossjones:
Reference: https://docs.docker.com/engine/installation/linux/ubuntu/#uninstall-old-versions

Some of the conversation re: name change: https://github.com/moby/moby/issues/31514


plaindocs:
Thanks!

PR #1260, opened 2017-06-16
Update getting-started.md

DusanSacha:
Reading through the doc I found out that https://travis-ci.org/repositories leads to 404 page.


plaindocs:
Good catch, thanks!

PR #1277, opened 2017-06-22
Collections POC

plaindocs:
One of the benefits of removing permalinks is that we could entirely automate the menu, by putting all pages in one of 6 or 7 collections, (with an optional weighting so we can sort things):

> Getting Started
Installing Dependencies
Programming Languages
Deployments and Uploads
CI Environment Reference
Encrypting Files and Data
Beta Features
Integrations and Notifications
Developer Program
3rd Party Tools and Resources

Then by moving a file from one collection to the other we'd automatically update the menu, and would not have to worry about permalinks, or about adding new entries for a new deployment.

Couple of gotchas:

* Differences with menu title and page title, for example "Building an Android Project" vs "Android. Fixable by have an option `alt_title` in the front matter and using that.

* We have a few random files here and there, Community Supported Languages for example, which isn't strictly a language.


plaindocs:
Most of these commits are from the permalink PR, only 

Move the included file to _includes			                                bfc1418
Move all languages to collection. Iterate over them.			69ce09c

Are related to the collection stuff.

plaindocs:
FAO @MariadeAnton although again no hurry

plaindocs:
I think I cherry picked most of this. Anyway, looked at the conflicting files and am closing this :-D

PR #1218, opened 2017-06-01
Update javascript-with-nodejs.md

cyberhck:

PR #1278, opened 2017-06-23
Add node 8 to documentation

davidspinat:


davidspinat:
I'm not sure how fix the failing checks. May I ask for help?

Thank you

davidspinat:
@plaindocs thanks!

pimterry:
I was just about to open this PR as well. Can we get this merged?

BanzaiMan:
I'll repost my comment to the duplicate issue https://github.com/travis-ci/docs-travis-ci-com/pull/1325#issuecomment-316160958

----

I think this section needs a different approach. Enumerating latest versions and keeping the list up-to-date is cumbersome.

Instead, we should simply point out that Node.js versions can be specified by the `node_js` key, and that the list can be obtained by `nvm ls-remote`. We could also generate the list of Node.js versions in https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Even-more-Nodejs-versions as well.

BanzaiMan:
Duplicate of #1218, as it turns out.

PR #1284, opened 2017-06-29
Xcode 9beta2 information

solarce:
Xcode 9 beta 2 information
```
$ xcodebuild -version

Xcode 9.0

Build version 9M137d
```

### DIFF
--- a/_data/xcodes.yml
+++ b/_data/xcodes.yml
@@ -1,4 +1,43 @@
 osx_images:
+  - image: xcode9
+    xcode: "9.0"
+    osx_version: "10.12"
+    xcode_full_version: "9.0"
+    image_publish_date: 2017-06-29
+    sdks:
+      - macosx10.13
+      - iphoneos11.0
+      - iphonesimulator11.0
+      - appletvos11.0
+      - appletvsimulator11.0
+      - watchos4.0
+      - watchsimulator4.0
+    simulators:
+      - iOS 8.1
+      - iOS 8.2
+      - iOS 8.3
+      - iOS 8.4
+      - iOS 9.0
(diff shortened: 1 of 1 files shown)


_data/xcodes.yml
```diff
@@ -1,4 +1,43 @@
...
+      - iOS 8.4
+      - iOS 9.0
+      - iOS 9.1
+      - iOS 9.2
+      - iOS 9.3
+      - iOS 10.0
+      - iOS 10.1
+      - iOS 10.2
+      - iOS 10.3
+      - ios 10.11
```
emdantrim: this one isn't capitalized `iOS` but the rest are. also, do the versions really go from 10.3 to 10.11?

_data/xcodes.yml
```diff
@@ -1,4 +1,43 @@
 osx_images:
+  - image: xcode9
+    xcode: "9.0"
+    osx_version: "10.12"
+    xcode_full_version: "9.0"
```
emdantrim: this should say "9.0 Beta 3"

_data/xcodes.yml
```diff
@@ -1,4 +1,43 @@
 osx_images:
+  - image: xcode9
+    xcode: "9.0"
+    osx_version: "10.12"
+    xcode_full_version: "9 beta 2"
```
emdantrim: I'd prefer a capital B just to be in line with how it looks for everything else, but it's not super important. :b:

solarce: Since it's named **9 beta 2** on https://developer.apple.com/download/ I thought following that would make the most sense

PR #1289, opened 2017-07-01
MariaDB uses root as user not travis

JuliusKoronci:
It is really nowhere to find what the user should be. If you are upgrading from mysql, or read the docs athe the begining you asume the user is travis

### DIFF
--- a/user/database-setup.md
+++ b/user/database-setup.md
@@ -224,7 +224,7 @@ addons:
   mariadb: '10.0'
 ```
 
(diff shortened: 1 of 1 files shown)


plaindocs:
Thanks @JuliusKoronci 

@joepvd I'm gonna add this - any other gotches related to MariaDB while I'm here?

JuliusKoronci:
Nope just this one :) thanks

PR #1293, opened 2017-07-03
How to make a repo public

plaindocs:
@aakritigupta how does this look?

### DIFF
--- a/user/travis-pro.md
+++ b/user/travis-pro.md
@@ -71,6 +71,6 @@ the setting on GitHub, then resync your Travis CI account:
 GitHub](https://help.github.com/articles/making-a-private-repository-public/) on
 how to make a repository public.
 
-2. On [Travis CI](Travis-ci.org) go to *Acccounts* and click *Sync account*.
+2. On [Travis CI](https://travis-ci.org) go to *Acccounts* and click *Sync account*.
 
 3. When you see the repository in the list, toggle the switch to *enabled*.
--- a/user/travis-pro.md
+++ b/user/travis-pro.md
@@ -69,16 +69,16 @@ deactivate it on [Travis CI .com](https://travis-ci.com), change the repository
 settings on GitHub, and resync your Travis CI account:
 
 1. On [Travis CI .com](https://travis-ci.com) go to *Acccounts*, and toggle the
-   repository sync to *OFF*.
+   repository to *OFF*.
 
 1. Follow the [instructions on
 GitHub](https://help.github.com/articles/making-a-private-repository-public/) on
 how to make a repository public.
 
-2. On [Travis CI .org](https://travis-ci.org) go to *Acccounts* and toggle the
-   repository to *ON*.
+2. On [Travis CI .org](https://travis-ci.org) go to *Acccounts*, click *Sync
+   Account*, then toggle the repository to *ON*.
 
-5. If you're using the [Travis CI commmandline client](https://github.com/travis-ci/travis.rb#readme)
+5. If you're using the Travis CI [commmand line client](https://github.com/travis-ci/travis.rb#readme)
    reset the default endpoint to public:
 
    ```sh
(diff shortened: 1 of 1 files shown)


user/travis-pro.md
```diff
@@ -61,3 +61,16 @@ Travis CI is, and always will be, free for open source projects.
 
 For a list of plans and prices for private repositories, look at
 [travis-ci.com/plans](https://travis-ci.com/plans).
+
+## Making a private repository public
```
aakritigupta: Since the other topics are in the form of questions, I think this should be a question too. WDYT?

user/travis-pro.md
```diff
@@ -61,3 +61,16 @@ Travis CI is, and always will be, free for open source projects.
...
+
+If you have a private repository that you'd like to make public, first change
+the setting on GitHub, then resync your Travis CI account:
+
+1. Follow the [instructions on
+GitHub](https://help.github.com/articles/making-a-private-repository-public/) on
```
aakritigupta: Here, we also need to mention about changing the notify url.

But sometimes this doesn't work either, as we realized in the recent mails. So please add a line, that if these steps don't work, they should write to us

user/travis-pro.md
plaindocs: Good point?

user/travis-pro.md
plaindocs: So that doesn't happen automatically? It is just referenced as a maybe in the original

aakritigupta: Right that makes sense. But I'd much rather that we add it, for more clarity

MariadeAnton: From our experience, we normally fix this by de-activating the hook for private repos and activating it on public ones (or vice-versa). Could we adapt the instructions to match this? Thank you!

The notify URL is currently a way for us to understand what happens on the inside, besides from the hook being different. I think it's alright to leave it out, to avoid confusion and people manually updating URLs - but if we find that it's helpful support-wise we can add it later on. What do you think?

aakritigupta: :+1: and :+1:

BanzaiMan:
There are more considerations, actually.

For one, the CLI tool stores the repository's API endpoint in `~/.travis/config.yml`, and I believe one must edit this file manually if one is to keep using `travis` without specifying `--com` (or `-e https://api.travis-ci.com`) after making the repo public.

Also, it may not be as common, but some users have taken repository private as well.

MariadeAnton:
@BanzaiMan I believe both options public →  private and viceversa are fairly common. 

As for CLI considerations when open sourcing a repository - I think that's right, and that people can also use the endopint option in the CLI to do so, is that possible? 

- `travis endpoint --org --set-default`
- `travis endpoint --pro --set-default`

plaindocs:
Thanks for the feedback @BanzaiMan and @MariadeAnton -useful stuff. 

Can you clarify "Also, it may not be as common, but some users have taken repository."

BanzaiMan:
@plaindocs This PR proposes to add a section titled "Making a private repository public". I am saying that making a public repository private should also be treated, as the same kinds of considerations come into play.

user/travis-pro.md
plaindocs: @MariadeAnton when you say deactivate the hook do you mean toggling on Travis-ci or do you mean deleting the hook on GitHUb?

plaindocs:
I've clarified the Making a repo public, don't both with the reverse yet, I'll fix that up when we've got the first part correct and complete.

user/travis-pro.md
MariadeAnton: Sorry for the confusion, @plaindocs - I meant toggling off on Travis CI. 

To recap, if you were to make a private repo open source you would: 

1. Toggle off in Travis Pro travis-ci.com/profile/owner_name 
2. Log in travis-ci.org
3. If the repo does not appear yet in your open source profile, trigger a sync
4. Toggle on the repository in your open source profile page travis-ci.org/profile/owner_name 

Does that make more sense? Thank you!

plaindocs:
OK, looks about ok to me, anyone have more to add?

user/travis-pro.md
```diff
@@ -61,3 +61,49 @@ Travis CI is, and always will be, free for open source projects.
...
+settings on GitHub, and resync your Travis CI account:
+
+1. On [Travis CI .org](https://travis-ci.org) go to *Acccounts*, and toggle the
+   repository to *OFF*.
+
+1. Follow the [instructions on
+GitHub](https://help.github.com/articles/making-a-public-repository-private/) on
+how to make a repository public.
+
+2. On [Travis CI .org](https://travis-ci.org) go to *Acccounts*, click *Sync
```
BanzaiMan: Do you mean .com here?

user/travis-pro.md
```diff
@@ -61,3 +61,49 @@ Travis CI is, and always will be, free for open source projects.
...
+settings on GitHub, and resync your Travis CI account:
+
+1. On [Travis CI .com](https://travis-ci.com) go to *Acccounts*, and toggle the
+   repository to *OFF*.
+
+1. Follow the [instructions on
+GitHub](https://help.github.com/articles/making-a-private-repository-public/) on
+how to make a repository public.
+
+2. On [Travis CI .org](https://travis-ci.org) go to *Acccounts*, click *Sync
```
BanzaiMan: Using `1.` everywhere will relieve you from keeping track of the order.

PR #1298, opened 2017-07-05
Fix "ccache on OS X example"

ctrlaltca:
In order to make ccache available as the default compiler for the following commands, the modified PATH must be exported to the environment.


plaindocs:
Thanks!

PR #1300, opened 2017-07-07
Update releasing NPM with details about token info

brendo:
The terminology between Travis and NPM differs in that Travis refers to the Auth Token as an API Key.

I'm fairly new to NPM, so haven't worked with NPM 1 or 2, but I've never seen the term "api key" be used from any NPM documentation so found this a little confusing at first!


plaindocs:
Thanks for that @brendo , yeah, legacy docs from a few years back I think.

You missed one in https://github.com/travis-ci/docs-travis-ci-com/pull/1300/files#diff-dcf2c16b5fa2a9ba1c1d8da22d51ef4eR135 though :-p

brendo:
Ah, that's on purpose because [I left this alone too](https://github.com/travis-ci/docs-travis-ci-com/pull/1300/files#diff-dcf2c16b5fa2a9ba1c1d8da22d51ef4eR26). I didn't think the property name could be changed just yet, so thought this would be ok.

Happy to change if you think otherwise though :)

plaindocs:
Yeah, that makes sense. But is probably misleading - because you're encrypting the `YOUR_AUTH_TOKEN` not the property name (which indeed you don't change).

What do you think?

brendo:
Hey @plaindocs, sorry for the long delay. I had another go at this pull request because after re-reading the entire file, I realised the separation between NPM v2 and v1 content isn't necessary.

The way the document reads currently it feels like the encryption of keys is only valid for NPM v1, which is not the case. I think the proposed restructure simplifies the document and makes it less ambiguous. 

I also make a couple of minor updates to point to `npmjs.com` instead of `npmjs.org`. It's a 301 so no biggie, but thought it'd be better to be current. Let me know what you think!

user/deployment/npm.md
```diff
@@ -23,14 +21,27 @@ deploy:
...
+it can modify the `.travis.yml` for you.
 
-## NPM Auth Token
+## NPM auth token
 
 Your NPM Auth Token can be obtained by:
 
-1. Login into [npmjs.com](http://www.npmjs.com), you [generating a new token](https://www.npmjs.com/settings/tokens).
-2. Login via the CLI with [npm adduser](https://docs.npmjs.com/cli/adduser) and look at the `authToken` value in your `~/.npmrc` file.
+1. Login to your NPM account, and [generating a new token](https://www.npmjs.com/settings/tokens).
```
plaindocs: This should probably be "Log in" . Obviously the mistake predates your PR, but we may as well fix.

plaindocs: Also s/generating/generate/

user/deployment/npm.md
```diff
@@ -23,14 +21,27 @@ deploy:
...
+1. Using the NPM CLI command [`npm adduser`](https://docs.npmjs.com/cli/adduser) and using the `authToken` (npm 2+) value in your `~/.npmrc` file. For npm version ~1 use the `auth` value.
```
plaindocs: This line isn't very clear at all. Lets split it out into two bullets in a sublist, npm ~1 and npm 2+. 

And lets be explicit that first you `add user` and then look in the file for one of the values

user/deployment/npm.md
```diff
@@ -124,40 +135,6 @@ after_deploy:
   - ./after_deploy_2.sh
 ```
 
-## npm is ~1 or older
```
plaindocs: Awesome, I love deleting large blocks of docs! <3

user/deployment/npm.md
brendo: Fixed!

user/deployment/npm.md
brendo: Fixed!

plaindocs:
Thanks for your work on this @brendo !

brendo:
No worries @plaindocs, glad I could help :)

PR #1304, opened 2017-07-10
Guessable secrets can become a leak

joepvd:


plaindocs:
Maybe something like 

> To prevent leaks made by these components, we automatically filter secure environment variables and tokens that are longer than three characters at runtime, effectively removing them from the build log, displaying the string `[secure]` instead. 

> Please make sure your secret is never related to the repository or branch name, or any other guessable string. Ideally use a password generation tool such as `mkpasswd` instead of choosing a secret yourself.

plaindocs:
Awesome :-)

PR #1305, opened 2017-07-10
Remove pre-release Xcode disclaimer

abbeycode:
You do provide pre-release versions of Xcode now, so it doesn't seem like this disclaimer is necessary.


plaindocs:
Thanks for pointing this out @abbeycode. Fixed in #1756

PR #1321, opened 2017-07-17
Put this back where you found it Sam

plaindocs:
Does what it says.


joepvd:
OK to ignore broken test for now, as this PR addresses this.

PR #1325, opened 2017-07-18
Add Node 8 to list of Node.js versions

RhysBower:


BanzaiMan:
I think this section needs a different approach. Enumerating latest versions and keeping the list up-to-date is cumbersome.

Instead, we should simply point out that Node.js versions can be specified by the `node_js` key, and that the list can be obtained by `nvm ls-remote`. We could also generate the list of Node.js versions in https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Even-more-Nodejs-versions as well.

BanzaiMan:
A duplicate of #1278.

PR #1332, opened 2017-07-20
Warn about overriding `$USER`

BanzaiMan:
If the user overrides this, a Bad Thing™ can happen.


BanzaiMan:
In particular, it will cripple `travis-build`'s ability to clean up background processes.

plaindocs:
Thoughts on using another word than 'override'? Something like

> Do not change the value of $USER

PR #1336, opened 2017-07-24
fix broken links

robertfausk:
by adding missing dir 'user'.

Maybe some doc pages was moved in user dir without adopting internal links.

### DIFF
--- a/user/sauce-connect.md
+++ b/user/sauce-connect.md
@@ -11,7 +11,7 @@ The integration automatically sets up a tunnel required to get started testing
 with it. For that purpose, it uses Sauce Connect.
 
 Note that due to security restrictions, the Sauce Labs addon is not available on pull
-request builds unless you use the [JWT Addon](../jwt).
+request builds unless you use the [JWT Addon](../user/jwt).
 
 ## Setting up Sauce Connect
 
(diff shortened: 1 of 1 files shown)


BanzaiMan:
I believe the current links are correct. See https://docs.travis-ci.com/user/sauce-connect/ and https://travis-ci.org/travis-ci/docs-travis-ci-com/builds/256928903#L555-L561

Where do you see the errors?

robertfausk:
It seems that it depends on the uri the user is coming from.
Wether it has a trailing slash or not.

There is currently following behavior:
* https://docs.travis-ci.com/user/sauce-connect#stq=&stp=0 shows broken links
* also https://docs.travis-ci.com/user/sauce-connect shows broken links
* https://docs.travis-ci.com/user/sauce-connect/ shows correct links

I think we need another PR ;)

image

BanzaiMan:
Thanks for the info. This should be fixed by https://github.com/travis-ci/docs-travis-ci-com/commit/7f0030bf85b59c13066f8254767676dd392e9b27.

One interesting thing to note is that https://docs.travis-ci.com/user/sauce-connect _should_ be redirected to https://docs.travis-ci.com/user/sauce-connect/, (at least, that is how the app behaves locally) but isn't.

BanzaiMan:
Hmm. The issue seems to persist.

@robertfausk Do you mind opening a separate ticket? Thanks.

PR #1338, opened 2017-07-24
Added WSL to alternatives for Windows on file encryption

levrik:
Works fine for me with WSL. Native Windows does not work.


plaindocs:
Thanks!

PR #1345, opened 2017-08-01
Remove link to build stages feedback issue

lislis:
The issue was locked

### DIFF
--- a/user/build-stages.md
+++ b/user/build-stages.md
@@ -6,7 +6,7 @@ layout: en
 
 <div id="toc"></div>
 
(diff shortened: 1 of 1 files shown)


plaindocs:
:+1: 

We're still leaving stages in Beta though?

lislis:
@plaindocs yes, it's still in beta until all feedback from the ticket is incorporated (or at least decided upon). We have _a lot_ of feedback on the ticket already and don't need more :D

plaindocs:
Roger that!

Issue #1340, opened 2017-07-24
Broken links when calling doc URIs without trailing slash


robertfausk:
There is currently following behavior:
* https://docs.travis-ci.com/user/sauce-connect#stq=&stp=0 shows broken links
* also https://docs.travis-ci.com/user/sauce-connect shows broken links
* https://docs.travis-ci.com/user/sauce-connect/ shows correct links

![image](https://user-images.githubusercontent.com/1651297/28532743-f8f4dfd4-709a-11e7-9976-659c9403ad26.png)

According to https://github.com/travis-ci/docs-travis-ci-com/pull/1336#issuecomment-317481363 @BanzaiMan stated:
"https://docs.travis-ci.com/user/sauce-connect should be redirected to https://docs.travis-ci.com/user/sauce-connect/, (at least, that is how the app behaves locally) but isn't."

PR #1347, opened 2017-08-01
Broken links and Trailing slashes

plaindocs:
Looking into the trailing slash issue reported in #1340 and #1336


plaindocs:
There should be no links to `/user/sauce-connect` only to `/user/sauce-connect/` I think with 7f0030b and this commit, the issue should be fixed.

PR #1359, opened 2017-08-08
Rename respository variable to repository_slug

cas--:
- Docs mention that %{repository} is deprecated for %{repository_slug} so examples need changing.
 - Remove duplicated (but missing deprecation message about repository) variables list in Slack section.


plaindocs:
Thanks

PR #1372, opened 2017-08-17
Fix OS version at "Build Environment Overview"

dlibanori:
They are swapped


plaindocs:
Thanks for picking up on that!

PR #1383, opened 2017-08-21
Elaborate on how non-tag builds and GitHub Releases interact

BanzaiMan:
This explains how https://github.com/travis-ci/travis-ci/issues/8185
manifests, and advises against running GitHub Releases without a tag.

This would be rendered like so:

github_releases_uploading_-_travis_ci


Deilan:
@BanzaiMan As far as I remember, a git `tag` should be pushed to a GitHub repo before a release is deployed against it. A git `user.name` and `user.email` should be also provided beforehand.

So if Travis doesn't do all this automagically, maybe these circumstances should be reflected in the docs.

Our bash script resolving this is as following:
```bash
#!/bin/bash

GIT_USER_EMAIL="${1:-${GIT_USER_EMAIL}}"
GIT_USER_NAME="${2:-${GIT_USER_NAME}}"
GIT_TAG="${3:-${GIT_TAG}}"
GIT_COMMIT_MESSAGE="${4:-${GIT_COMMIT_MESSAGE}}"

if [[ ! $(git config user.email) ]]; then
    git config user.email "${GIT_USER_EMAIL}"
fi

if [[ ! $(git config user.name) ]]; then
    git config user.name "${GIT_USER_NAME}"
fi

if [[ ! $(git tag -l "${GIT_TAG}") ]]; then
    git tag -a "${GIT_TAG}" -m "${GIT_COMMIT_MESSAGE}"
    git push -q origin --tags
fi
```

Deilan:
@BanzaiMan It's not very clear what are the consequences of using `on.tags: false`. More generally, I think it should be described a bit more specifically, what happens when:
1) `on.tags: false` is explicitly set
2) `on.tags: true` is explicitly set
3) `on.tags` is not specified

Consider using a concise diagram to illustrate it. :)

Deilan:
@BanzaiMan What about `on.branch` and `on.all_branches` options? I hope they will continue to be working.

Deilan:
@BanzaiMan For a more complete picture I also would want to recall about a related issue travis-ci/travis-ci#8051, which results in a redundant job for every single tag in the case we are discussing.

BanzaiMan:
@Deilan Thanks for the feedback. You raise some important points.

A few thoughts (roughly in the order you raised):

1. Having to have `git` configured correctly is important. I'll add that.
1. `on.tags: false` _should_ be functionally equivalent to not writing it, since we just check for the truthiness of `on.tags`, not whether or not the key exists.
1. This PR aims to explain what happens in the cases that are not explicitly mentioned. `on.branches` as well as `on.all_branches` are, and they are not affected.
1. IMO, https://github.com/travis-ci/travis-ci/issues/8051 is related, but should be unaffected.

BanzaiMan:
On item 2 above… I now realize that there *is* a spot where we explicitly check for the value of `on.tags`; `on.tags: false` is translated to `[[ $TRAVIS_TAG = "" ]]`. So, yes, it seems to me that this requires some further exposition.

ezyang:
LGTM.

plaindocs:
@BanzaiMan sorry it has taken me so long to get to this. Your PR was great, but I found some stuff I needed to update around it. Take a quick look and merge at will?

PR #1388, opened 2017-08-24
Node JS updates for Trusty

plaindocs:
Move version info into reference

### DIFF
--- a/user/languages/javascript-with-nodejs.md
+++ b/user/languages/javascript-with-nodejs.md
@@ -4,9 +4,33 @@ layout: en
 
 ---
 
-<div id="toc"></div>
+<div id="toc">
+</div>
 
-This guide covers build environment and configuration topics specific to Node.js
+<aside markdown="block" class="ataglance">
+
+| JS and Node                                 | Default                                   |
+|:--------------------------------------------|:------------------------------------------|
+| [Default `install`](#Dependency-Management) | `npm install`                             |
+| [Default `script`](#Default-Build-Script)   | `npm test`                                |
+| [Matrix keys](#Build-Matrix)                | `env`, `node_js`                          |
+| Support                                     | [Travis CI](mailto:support@travis-ci.com) |
+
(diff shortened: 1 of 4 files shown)


user/languages/javascript-with-nodejs.md
```diff
@@ -4,9 +4,33 @@ layout: en
...
 ---
 
-<div id="toc"></div>
+<div id="toc">
+</div>
 
-This guide covers build environment and configuration topics specific to Node.js
+<aside markdown="block" class="ataglance">
+
+| JS and Node                                 | Default                                   |
```
BanzaiMan: This should be "JavaScript and Node.js", I think.

plaindocs: Yeah, we've not been too consistent before - I'll fix up the title as well.

user/languages/javascript-with-nodejs.md
```diff
@@ -245,60 +293,18 @@ Find the source code at [travis-ci-meteor-packages](https://github.com/arunoda/t
...
-.travis.yml](https://github.com/scottyapp/hook.io-amqp-listener/blob/master/.travis.yml).
-
-For precise versions pre-installed on the VM, please consult "Build system
-information" in the build log.
-
 ## Node.js v4 (or io.js v3) compiler requirements
 
-To compile native modules for io.js v3 or Node.js v4 or later, a
-[C++11 standard](https://en.wikipedia.org/wiki/C%2B%2B11)-compliant compiler is required. More specifically, either gcc 4.8 (or later), or clang 3.5 (or later) works.
+To compile native modules for io.js v3 or Node.js v4 or older, a [C++11
```
BanzaiMan: Is the switch to "older" intentional?

plaindocs: It was, but perhaps I was mistaken. It followed on from the old Node versions topic. But this is actually compiling modules not older versions.

PR #1389, opened 2017-08-24
Julia: Refer to PkgDev's generate instead of Pkg's

christopher-dG:
`Pkg.generate` has moved to `PkgDev`.


BanzaiMan:
@tkelman @ninjin @staticfloat @simonbyrne Review, please?

PR #1391, opened 2017-08-24
Populate GCE IP address range automatically

BanzaiMan:
Using the steps described in https://cloud.google.com/compute/docs/faq#where_can_i_find_short_product_name_ip_ranges
we populate the IP address range table for users' easier consumption.

We also add a new Rake task `gen_gce_ip_addr_range`, which will
be invoked by the standard `build` task.

It looks like this:

ip_addresses_-_travis_ci


Lyoness:
rad @BanzaiMan

plaindocs:
Awesome!

PR #1399, opened 2017-08-28
Bump node version up to v8

tomscholz:


BanzaiMan:
Thanks for taking time to open this PR, but I'm closing this in favor of #1388.

tomscholz:
That makes sense :)

PR #1404, opened 2017-08-29
Update Cassandra docs to specify 'sudo: required'

soulshake:
Activating Cassandra with:
```
services:
  - cassandra
```
results in the following error being displayed in the build logs:

```
0.03s$ sudo service cassandra start
/etc/init.d/cassandra: 72: ulimit: error setting limit (Operation not permitted)
```
and in at least one case, the following failure:
```
cannot open cassandra session: gocql: unable to create session: unable to discover protocol version: dial tcp 127.0.0.1:9042: getsockopt: connection refused
```

From what I can tell, the ulimit error is docker-related, and doesn't prevent Cassandra from starting[[1](https://travis-ci.org/soulshake/startup-threads-cli/jobs/269519690)]. Nevertheless, `sudo: required` seems to be a required workaround. (Explanations welcome!)


BanzaiMan:
Is this to suggest that Cassandra won't be supported on containers? I seem to remember this error message being reported somewhere, but I can't find it anywhere.

bogdanap:
@soulshake This seems like it might actually be a problem - I've opened an issue for us at https://github.com/travis-ci/packer-templates/issues/504

The build that you linked (where cassandra actually started) ran on GCE, which explains why it succeeded.

soulshake:
Oh nice catch @bogdanap. (I guess because I used sudo in the script...?)

plaindocs:
Hey @soulshake - what is the status of this?

bogdanap:
We discussed this as part of https://github.com/travis-pro/team-blue/issues/781 and decided that we should indeed only support Cassandra on GCE, so I think this can be safely merged.
Thank you for the reminder @plaindocs.

plaindocs:
Thanks!

So we should probably update the text to say so explicitly, and then delete the final note:

> If you’re using Container-based infrastructure you won’t be able to install other versions of Cassandra as the sudo command is not available.

Wanna grab that or shall I?

PR #1407, opened 2017-08-30
Clarifies `TRAVIS_BRANCH` when building for a tag.

Junkern:
The documentation is not precise enough about the `TRAVIS_BRANCH` env variable when the travis build is running for a tag. It actually contains the name of the tag, but this fact is not documented well enough. (See also this issue: https://github.com/travis-ci/travis-ci/issues/8200)


plaindocs:
Hi @Junkern, thanks for this. 

I think it makes sense to add the extra line, but remove the slightly odd 

"This is because,
  
 -      > Note that "

What do you think?

Junkern:
Huh? According to the changes, the 

> Note that for tags (...)

got changed to

> ... for tags (...)

Or am I interpreting the `"Files changed"` section wrong? :D

plaindocs:
Yeah, that is what I saw. I think we can just add the one line 

+ for builds triggered by a tag, this is the same as the name of the tag (`TRAVIS_TAG`).

and leave the Note untouched, and let people connect the two.

Junkern:
Ahh, now I understand your first comment ;) Yes, actually makes sense, although I think that some people will still be confused^^

PR #1412, opened 2017-08-30
PHP Updates for Trusty

plaindocs:


iriberri:
After checking https://packages.ubuntu.com/trusty/php/, it looks like indeed this is true:

> Specifying exact versions like 5.3.8 is discouraged as it may break your build when we update PHP versions on Travis CI. PHP version 5.5.9 is supported, however, because it’s the version of PHP that is shipped with Ubuntu 14.04 LTS.

Also, it looks like for Precise the PHP version shipped with it was 5.3.10, and in the future Xenial will come with a different one, so this a comment we might want to somehow move to the Trusty/Precise pages instead.

iriberri:
I just finished reading the rest of the work in the branch and it looks neat! :-) ✨

user/languages/php.md
```diff
@@ -62,8 +62,21 @@ php:
...
+These may not be the most recent releases. If you need to ensure the use of most recent releases, do provide the third number; e.g.,
+
+```yaml
+language: php
+php:
+  - 7.1.9
+```
+{: data-file=".travis.yml"}
+
+### PHP 5.2(.x) and 5.3(.x) support
```
plaindocs: What do you think about removing this section from this file entirely, and just leaving the info in Precise / Trusty?

BanzaiMan: I think it is important there is something being said here, very much like HHVM is treated. I expect users of PHP to look for information on this page, rather than to look at Precise/Trusty .

plaindocs: And I'd probably prefer "we don't support" over "we're unable to provide support". Maybe @MariadeAnton has thoughts on this?

BanzaiMan: Aye. That indeed sounds better.

user/languages/php.md
```diff
@@ -4,80 +4,111 @@ layout: en
...
+The default build script is PHPUnit. It comes packaged with PHP, but you can also
 install a specific version in a custom location. If you do install it
 separately, make sure you invoke the correct version by using the full path.
 
-Travis CI looks for `phpunit` in the [same order as Composer does](https://getcomposer.org/doc/articles/vendor-binaries.md#can-vendor-binaries-be-installed-somewhere-other-than-vendor-bin-)
+Travis CI looks for `phpunit` in the [same order as Composer
+does](https://getcomposer.org/doc/articles/vendor-binaries.md#can-vendor-binaries-be-installed-somewhere-other-than-vendor-bin-)
 and uses the first one found.
 
 1. `$COMPOSER_BIN_DIR/phpunit`
```
MariadeAnton: after this, when we say
> If your project uses something other than PHPUnit, you can override the default test command.

how do you feel about updating it to: 
> If your project uses something other than PHPUnit, you can override the default build script command.

So that it's consistent with the whole section **Default Build Script**?

MariadeAnton:
Added a tiny comment above - Overall this is fantastic! 💪

user/languages/php.md
BanzaiMan: @MariadeAnton Sounds good to me.

plaindocs:
Awesome

PR #1413, opened 2017-08-30
Document Git LFS

soulshake:
Note: I haven't thoroughly tested; this info has been cobbled together from what clues I could find laying around. If we're not sure this is correct info, let me know and I can test more.

### DIFF
--- a/user/customizing-the-build.md
+++ b/user/customizing-the-build.md
@@ -261,16 +261,45 @@ git:
   submodules: false
 ```
 
-## Git LFS Skip Smudge
+## Git LFS
 
-You can disable the download of LFS objects when cloning ([`git lfs smudge
---skip`](https://github.com/git-lfs/git-lfs/blob/master/docs/man/git-lfs-smudge.1.ronn))
-by setting the following in `.travis.yml`:
+
+### Linux
+
+[Git LFS](https://git-lfs.github.com/) is supported by default on our Ubuntu Trusty images.
+
+### Mac OS
+
+Installing git-lfs via brew is the recommended way to get Git LFS in Mac OS.
+
+```
+before_install:
+- brew install git-lfs
+
+before_script:
+- git lfs pull
+```
+
+### Git LFS Skip Smudge
+
+GitHub rate-limits the Git-LFS requests during the git clone process. If you run into rate-limiting issues, you can skip fetching the git-lfs files during the initial `git clone` (equivalent to [`git lfs smudge --skip`](https://github.com/git-lfs/git-lfs/blob/master/docs/man/git-lfs-smudge.1.ronn)), and download these assets during the `before_install` phase of your build. To achieve this, you can use the following configuration in `.travis.yml`:
 
 ```yaml
 git:
   lfs_skip_smudge: true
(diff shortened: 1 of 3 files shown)


user/customizing-the-build.md
```diff
@@ -261,17 +261,46 @@ git:
...
+### Git LFS Skip Smudge
+
+GitHub rate-limits the Git-LFS requests during the git clone process. If you run into rate-limiting issues, you can skip fetching the git-lfs files during the initial `git clone` (equivalent to [`git lfs smudge --skip`](https://github.com/git-lfs/git-lfs/blob/master/docs/man/git-lfs-smudge.1.ronn)), and download these assets during the `before_install` phase of your build. To achieve this, you can use the following configuration in `.travis.yml`:
 
 ```yaml
 git:
   lfs_skip_smudge: true
 ```
 
+### Authentication
```
plaindocs: Any idea when you "may need" to configure authentication? 

Also might be worth putting the Authentication before the Skip Smudge?

soulshake: @plaindocs I think it may be required when rate-limiting on unauthenticated requests kicks in, but I'm not sure.

MariadeAnton: I'm also very much in favour of recommending authenticated requests to prevent anyone from being rate-limited? @iriberri do you have any special thoughts on this?

iriberri: @MariadeAnton my +10000 for that - as @soulshake mentioned, this was something we investigated this week and I'd like to see it documented. I can help with this part if needed, just let me know!

Besides the rate-limiting, I think it would be interesting to add a mention somewhere about the bandwidth limit that an organization/user is paying for at GitHub. If this bandwidth is exceeded they're also going to see errors in their builds when downloading the git-lfs files. (See the bandwidth quota at https://help.github.com/articles/about-storage-and-bandwidth-usage/)

MariadeAnton:
This looks 💯 to me - thanks for all your work on it, @soulshake! 

And thanks for the feedback @plaindocs and @iriberri  😊 !

plaindocs:
Thanks @soulshake !

user/customizing-the-build.md
```diff
@@ -261,17 +261,52 @@ git:
...
-## Git LFS Skip Smudge
+## Git LFS
 
-You can disable the download of LFS objects when cloning ([`git lfs smudge
---skip`](https://github.com/git-lfs/git-lfs/blob/master/docs/man/git-lfs-smudge.1.ronn))
-by setting the following in `.travis.yml`:
+
+### Authentication
+
+As a general rule, we recommend using authenticated requests to avoid being rate-limited by GitHub:
```
plaindocs: Probably not worth opening another PR for now this is merged, but

> As a general rule, 

adds very little. I'd probably reword to something like

> We recommend using a read-only GitHub OAuth token to authenticate when using git LFS, this is required when connecting to private repositories, and prevents rate-limiting when connecting to open source repositories:

And I'd be tempted to add a single sentence description / link to what LFS is before the authentication section.

:-)

PR #1428, opened 2017-09-05
Elaborate on PHP versions; use of Z in x.y.Z, no 5.2 & 5.3 on Trusty

BanzaiMan:


plaindocs:
Also - would you mind checking that merges with #1412 ? (it probably doesn't)

BanzaiMan:
Ah, I should close this and add commits to your branch instead. :-)

PR #1435, opened 2017-09-07
Add mysql permissions trusty update

Lyoness:


Lyoness:
:man_facepalming: it's in the #FIXED header.

PR #1437, opened 2017-09-09
Update versions, show a matrix with multiple allow_failures

cclauss:
Be consistent with quotes.

`allow_failures` is very useful so it is good to demonstrate it.

Python 2.6  and 3.2 are EOL and 3.3 will be in just a few weeks so we should not emphasize them too much.
* https://docs.python.org/devguide/index.html#branchstatus


cclauss:
Should we indicate that `trusty` is the default?

Should we show `cache: pip`?

user/languages/python.md
```diff
@@ -49,17 +49,16 @@ levels.
...
-  - "3.2"
-  - "3.3"
-  - "3.4"
-  - "3.5"
-  - "3.5-dev" # 3.5 development branch
-  - "3.6"
-  - "3.6-dev" # 3.6 development branch
-  - "3.7-dev" # 3.7 development branch
-  - "nightly"
+  - 2.6
```
BanzaiMan: Keep quotes, here and elsewhere. `"2.6"` is a string, and `2.6` a float. They behave similarly in our applications, but they are different data structures.

user/languages/python.md
```diff
@@ -91,15 +90,15 @@ in your `.travis.yml`:
...
+  - 2.7
+  - 3.4
+  - 3.5
+  - 3.6
   # PyPy versions
-  - "pypy"  # PyPy2 2.5.0
-  - "pypy3" # Pypy3 2.4.0
-  - "pypy-5.3.1"
+  - pypy   # PyPy2 5.8.0
```
BanzaiMan: We should probably drop the PyPy version indication altogether. It is too brittle to keep up to date.

user/languages/python.md
```diff
@@ -199,6 +198,21 @@ For a real world example, see [getsentry/sentry](https://github.com/getsentry/se
 For Python projects, `env` and `python` can be given as arrays
 to construct a build matrix.
 
+```yaml
```
BanzaiMan: Not sure if this example adds much value here. There are examples using `python`, and `allow_failures` is considered elsewhere.

cclauss: The reason that I added allow_failures was because this is a year of porting in the Python world with the upcoming EOL of Python 2, the performance improvements of Python 3, and the Django 1 to 2 transition.  More people could be using CI to smoothly navigate these transitions if they understood the allow_failures approach so we should not hide it.

Let me know if you need it removed.

plaindocs:
Also related to #1540 quote issue.

user/languages/python.md
plaindocs: Hey @cclauss - if you could remove the example I'll merge this in.

> More people could be using CI to smoothly navigate these transitions if they understood the allow_failures approach so we should not hide it.

This might make more sense as a separate, language independent page on how to manage to version number transition on Travis CI using `allow_failure`, what do you think?

plaindocs:
Thanks

PR #1443, opened 2017-09-14
add a note on how to unset the _JAVA_OPTIONS

mprins:


plaindocs:
Hi @mprins

It probably isn't worth adding this to the environment updates, but perhaps you'd like to take a look at the JAVA pages and see if there is a place for the info there?

PR #1446, opened 2017-09-16
Fix Trusty Reference

SConaway:
The ... appears on a separate line.

### DIFF
--- a/user/getting-started.md
+++ b/user/getting-started.md
@@ -4,7 +4,7 @@ layout: en
 
 ---
 
-The very short guide to using Travis CI with your GitHub hosted code repository.
+This is a very short guide to using Travis CI with your GitHub hosted code repository.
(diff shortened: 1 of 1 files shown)


user/reference/trusty.md
```diff
@@ -34,7 +34,7 @@ sudo: false
 This is enabled for both public and private repositories.
 
 If you'd like to know more about the pros, cons, and current state of using
-Trusty, including details specific to container-based, read on ...
+Trusty, including details specific to container-based, read on &nbsp...
```
BanzaiMan: There is still a space between `on` and `&nbsp`. Also, the correct HTML entity is `&nbsp;`.

MariadeAnton:
Thanks for your PR,  @SConaway  - we'll look into removing those `...` instead.

PR #1464, opened 2017-09-22
Java updates for Trusty

iriberri:


user/languages/java.md
```diff
@@ -4,94 +4,120 @@ layout: en
...
 ```yaml
 language: java
 ```
 {: data-file=".travis.yml"}
 
-Older OpenJDK and Oracle JDK versions may be available on [Precise](/user/reference/precise#JDK).
-
 ## Projects Using Maven
 
-### Default script Command
```
iriberri: This and the following sections look all messed up in the diff because I moved the "Dependency Management" before "Script command", since `install` runs before `script`.

PR #1470, opened 2017-09-25
Fix flipped go get flags in Dependency Management section

zombiezen:
-t was introduced in Go 1.2.  The behavior of Travis (rightfully) is flipped from what the docs say.


plaindocs:
Thanks!

PR #1471, opened 2017-09-27
Add documentation for caching Ruby versions in non Ruby projects

javigomez:
Hi! :)

I am adding a quick trick in documentation to save more than 3 minutes in builds that are non Ruby based but that use Ruby executions in it.

Please add your thoughts. I am not English native

Thanks


plaindocs:
Hi @javigomez 

Thanks for that. In this case I'm not sure showing the `.travis.yaml` is helpful, we can probably just use the description of "For example, a NodeJS application that has a Ruby functional test suite". 

What do you think?

user/caching.md
```diff
@@ -76,6 +76,32 @@ In the cases where this is not desirable, you can use specify the [arbitrary dir
...
+  - cucumber features/
+```
+
+For these cases installing a version of ruby with `rvm install 2.3.1` may take more than 3 minutes. For these cases you can cache the ruby installation.
+
+```yaml
+ cache:
+    directories:
+     - /home/travis/.rvm/
+```
```
javigomez: @plaindocs do you mean removing lines from 99 to 103?

I am okay with it. But this are the lines that people will have to add to their YML to make it work, I think they are valuable. Actually, I feel they give more value than lines from 83 to 95. 

Still, I am not documentation expert. So feel free to modify it as you think is the best for TravisCI

plaindocs: I'm talkng about 83-95 - thanks

javigomez:
@plaindocs updated with your suggestion.

plaindocs:
Thanks!

PR #1475, opened 2017-09-28
Add a note differentiating between Travis and Maven build install steps

kdaniels:


plaindocs:
If we're using the `install` step as an example, would it make sense to this not in the dependency management section?

kdaniels:
@plaindocs the "Maven dependency management" section here? If so, yeah good catch and I can make that change!

PR #1490, opened 2017-10-06
Explain difference between Elixir x.y and x.y.z

BanzaiMan:
See https://github.com/elixir-lang/elixir/issues/6618 for discussion.


user/languages/elixir.md
```diff
@@ -44,6 +44,30 @@ language: elixir
 ```
 {: data-file=".travis.yml"}
 
+### Specify which Elixir version to test with
```
plaindocs: We should probably default to to 'build' instead of 'test' here?

plaindocs:
Looks good to me. 

Are we quoting version numbers in the yaml specifically here for a reason?

user/languages/elixir.md
BanzaiMan: Agreed.

BanzaiMan:
Quotation is for consistency; `1.5` is technically a float, not a string. I know this is not consistently enforced in our docs, but we should try to be more consistent nonetheless.

plaindocs:
So we should be quoting *all* version numbers as strings in yaml? Roger that.

PR #1492, opened 2017-10-07
Document language:generic

gliptak:


gliptak:
@plaindocs @BanzaiMan Please review. Thanks

plaindocs:
Thanks @gliptak !

I think we're going to look into this in more detail. I've opened an issue at #1507 to track it.

gliptak:
Before you can get the right attention to this, why not update the site with this PR for your users benefit ...

PR #1498, opened 2017-10-12
Update common-build-problems.md

abitrolly:
Explain build timeouts in fewer words


plaindocs:
Thanks for this @abitrolly - fewer words are often a good thing, but most of these ones add a little example or context, so I'm don't think we should remove them.

abitrolly:
It also explains `travis_wait` more accurately.

PR #1499, opened 2017-10-12
fix syntax highlighting for a code block in the rust doc

lukad:
You can see the broken code block here: https://docs.travis-ci.com/user/languages/rust/#Dependency-Management


plaindocs:
Thanks!

PR #1501, opened 2017-10-12
Remove the r_check_revdep option

jimhester:
This is rarely used and it is really better to check reverse
dependencies locally rather than using travis to do so.


plaindocs:
LGTM, will merge when the code PR goes in.

PR #1519, opened 2017-10-20
Remove TRAVIS_PULL_REQUEST=false

antiBaconMachine:
I noticed $TRAVIS_PULL_REQUEST is not present at all in non PR builds. These docs state it should have a value of false. The $TRAVIS_PULL_REQUEST_BRANCH behaviour is correct so can be used instead anyway.


BanzaiMan:
Do you have a build that shows this? Because I am not seeing it. https://travis-ci.org/BanzaiMan/travis_production_test/builds/287786955#L532

plaindocs:
Closing this for now, please reopen if you have an example build.

PR #1521, opened 2017-10-23
Consistent example in DB setup documentation

nicolasleger:
Hello,

Just above in the file there is the `env` setup `DB=postgres` at https://docs.travis-ci.com/user/database-setup/#Using-environment-variables-and-a-before_script-step

Better to be consistent in this example.

Cheers


plaindocs:
Good catch. Thanks.

PR #1522, opened 2017-10-23
Fixed indentation issue in gcc on Linux section of c11-cpp11-and-beyond-and-toolchains.md

InfiniteCoder:
there was an indentation issue in the gcc on Linux section. One line was not indented at all, which when used as it is, travis would fail to parse the yml.


plaindocs:
Thanks :-)

PR #1531, opened 2017-10-30
Add a note on overriding Default Gradle Install Command.

artem-zinnatullin:
PR addresses https://github.com/travis-ci/travis-ci/issues/8667 and points users on how they could override default `gradle assemble` `install` command.

Also syncs section title with section below:

- "Gradle Default Install Command"
- "Gradle Default Script Command"

Edit: reverted title change as it used in same form everywhere else.


plaindocs:
Thanks

PR #1533, opened 2017-11-01
Fix a typo in a config key for GitHub Releases

oncletom:
`api-key` -> `api_key`


BanzaiMan:
The parameters are normalized, so this is not technically a typo. But this is better for consistency.

BanzaiMan:
The PR build shows an error because I canceled it (it was running after I'd merged this PR, so it was going to fail anyway.)

PR #1534, opened 2017-11-03
Update how to specify rubinius

kachick:
If we can clarify how to use at here, issues as https://github.com/rvm/rvm/issues/3651 will be reduced, I think 🤔 

@brixen How about this change? 🙏


plaindocs:
Thanks!

kachick:
Thank you! ☺️

PR #1537, opened 2017-11-06
Update haskell version example list to strings

iriberri:
In YAML, `- 7.10` would evaluate to version 7.1, and not 7.10.

I'm updating the examples here to use a list of strings instead. This comes from a specific request to update this for Haskell, but I assume is something we want to check in general, as this would affect other languages too.


plaindocs:
Thanks @iriberri . Yeah, @BanzaiMan raised a similar issue recently, I've opened an issue to track the larger scale update.

iriberri:
Hi @polyzen! That's a good point and of course that approach would work too :-) What I tried with this PR was to use strings in all the versions, since I thought that anyone taking those examples and expanding them for their use case would just imitate the quoting and therefore won't stumble into this issue.

This was however kind of a quick fix from a request - I definitely think that we need to document this in a more consistent way and also explain the limitation. We'll continue working on this as @plaindocs mentioned above :-)

BanzaiMan:
@polyzen In YAML, `7.8` is a float, `"7.8"` is a string. Software version numbers should be strings, not floats. We do not current document this consistently, but this PR is correct.

PR #1538, opened 2017-11-06
Add documentation on running Firefox in headless mode

davehunt:


user/gui-and-headless-browsers.md
```diff
@@ -146,6 +146,39 @@ before_install:
...
+```python
+from selenium.webdriver import Firefox
+from selenium.webdriver.firefox.options import Options
+
+options = Options()
+options.add_argument('-headless')
+firefox = Firefox(firefox_options=options)
+```
+
+### Documentation
```
edmorley: I think this needs to be `####` for consistency with the Chrome section above

davehunt: Oops! Thanks!

plaindocs:
Awesome, thanks!

PR #1542, opened 2017-11-08
Xcode9.1-beta to xcode9.1 GA

Lyoness:
- [ ]  Tweet out when merged "xcode9.1 updates have been made :tada: :+1: read about it here

### DIFF
--- a/user/build-environment-updates/2017-11-08.md
+++ b/user/build-environment-updates/2017-11-08.md
@@ -1,0 +1,71 @@
+---
+title: Build Environment Update History
+layout: en
+permalink: /user/build-environment-updates/2017-11-08/
+category: build_env_updates
+date: 2017-11-08
+---
+
+## 2017-11-08
+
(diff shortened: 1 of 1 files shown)


user/build-environment-updates/2017-11-08.md
```diff
@@ -0,0 +1,71 @@
...
+title: Build Environment Update History
+layout: en
+permalink: /user/build-environment-updates/2017-11-08/
+category: build_env_updates
+date: 2017-11-08
+---
+
+## 2017-11-08
+
+This update applies to macos10.12-xcode9.1 image, for `os: osx` builds
```
cotsog: Maybe we should say that it applies to `osx_image: xcode9.1` builds? I think that `os: osx` will route to the default image.

cotsog: We should perhaps mention the Xcode version that was installed i.e Xcode 9.1 (build 9B55)

PR #1551, opened 2017-11-15
Document workaround for PHPUnit on HHVM

bogdanap:
References: 
- https://github.com/travis-ci/travis-cookbooks/issues/908
- https://github.com/travis-ci/travis-ci/issues/8516#issuecomment-332979598

Since we didn't think it's worth downgrading PHPUnit, documenting the workaround seems like a good middle ground.

### DIFF
--- a/user/languages/php.md
+++ b/user/languages/php.md
@@ -89,7 +89,7 @@ php
 {: data-file=".travis.yml"}
 
 
-Please note that if you want to run PHPUnit on HHVM, you have to explicitly install version 5.7 in your .travis.yml due to a compatibility issue between HHVM and PHP7:
+Please note that if you want to run PHPUnit on HHVM, you have to explicitly install version 5.7 in your `.travis.yml` due to a compatibility issue between HHVM and PHP7:
 
 ```yaml
 before_script:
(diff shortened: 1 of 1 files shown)


plaindocs:
Hi @bogdanap ,

Cheers for this. Couple of comments:

What do you thing about removing 

> +In order to do that, you can add the following bit to your `.travis.yml`: 

And instead
 
> Please note that if you want to run PHPUnit on HHVM, you have to explicitly install version 5.7 in your `.travis.yml` due to a compatibility issue between HHVM and PHP7:

Also you can add `{: data-file=".travis.yml"}` immediately after the code block. 

Finally, does it make more sense to add this to the HHVM section, the PHPUnit section, or possibly even both instead of just throwing it onto the end?

bogdanap:
@plaindocs Thank you for the suggestions - I've tried to implement them, so please feel free to take another look 😄

plaindocs:
Perfect, thanks! :-)

uglyoldbob:
I needed to use `curl -sSLf -o ~/.phpenv/versions/hhvm/bin/phpunit https://phar.phpunit.de/phpunit-5.7.phar` to get it to work

PR #1552, opened 2017-11-15
Update languages.md

Daro111:
user/languages.md


BanzaiMan:
The language should be `en`.

PR #1554, opened 2017-11-17
travisyml reference 

plaindocs:
After some consultation with @joepvd  and @joecorcoran I've cobbled together some scripts to rip keys used in `.travis.yml` from the spec, with a view to providing an alphabetic listing of possible keys with a pointer to relevant docs.

First thing is to actually populate `travisyml.yml` with useful links (**help appreciated**). And also not be lazy and adapt the template to cope with multiple links for a single entry. 

It would be rather lovely to also setup redirects from `https://docs.travis-ci.com/travisyml/XYZ` to the relevant docs for XYZ which would make linking to docs from the UI or anywhere else kinda easy.

Finally, some kinda relevant grouping of keys would be nice, maybe dumping a datafile like

```yaml
r_binary_packages:
  - TODO
  tags:
    - rlang
r_build_args:
  - TODO
  tags:
    - rlang
r_check_args:
  - TODO
  tags:
    - rlang
r_check_revdep:
  - TODO
  tags:
    - rlang
```

Oh, and probably gonna have to install `jq` as well. 

Also FAO @svenfuchs and @MariadeAnton


plaindocs:
I'm thinking of doing a first pass to add more links, and then merging as WIP, given that now I've updated the template we only display keys that *have links*.

plaindocs:
Urgh, going to have to lower case all of the anchors before merging this.

plaindocs:
Close this in favor of https://yml-staging.travis-ci.org/v1/docs/

PR #1560, opened 2017-11-22
Updating a needed scope for deployment

augsteyer:
Seems to be a requirement when creating an oAuth github token to use for deployment. Without this my build kept breaking with a 401 error when installing dependencies:
```GET https://api.github.com/user: 401 - Bad credentials // See: https://developer.github.com/v3 (Octokit::Unauthorized)```

I am not exactly sure on the description of why this is needed, but all I know that it was crucial for me to do.

See example log: https://travis-ci.org/augsteyer/cloud-integration-magento/builds/305297570


augsteyer:
Ha, oops :)

PR #1565, opened 2017-11-27
user/coveralls: mention Docker tips

ueno:
This might be obvious, but I suppose it would be a bit helpful to mention how to use coveralls from Docker builds.


plaindocs:
Thanks

PR #1575, opened 2017-11-29
Added section for multiple packages

turion:
Added a section in which it is described how to set up a build matrix containing multiple packages in subdirectories.


user/languages/haskell.md
```diff
@@ -71,3 +71,22 @@ cabal install --only-dependencies --enable-tests
 
 For Haskell projects, `env` and `ghc` can be given as arrays
 to construct a build matrix.
+
+### Multiple Packages in Subdirectories
+
+If you have multiple packages in subdirectories (each containing a `.cabal` file,
+you can enter those directories by setting an environment variable:
```
plaindocs: Thanks for this. Do you think it would be more clear to say "specify those directories by setting an environment variable" even if the build command is `cd` ?

turion: I agree, your suggestion is more accurate!

plaindocs:
Thanks

muffcrusher:
I don't think we need it that much I don't think you should change a thing

PR #1578, opened 2017-11-29
Rework and update dynamic data sources from terraform & DNS

meatballhat:
### Added
- integrating EC2 public IP address data generated in [terraform-config](https://github.com/travis-infrastructure/terraform-config/pull/305)
- `pry` and `rubocop` in Gemfile for troubleshooting help and style automation
- explicit `regen` target for cleaning and remaking generated files

### Changed
- pull GCE IP range stuff via dig.jsondns.org instead of shelling out with `nslookup`
- sorted the contents of `.gitignore`
- bump the ruby version in Gemfile to 2.4.2
- sorted the contents of the Gemfile
- update some style bits in Rakefile with the help of rubocop
- ensure the `build` target depends on the `regen` target

I notably _did not_ run `rubocop --auto-correct --auto-gen-config` across the entire codebase, as I didn't want to drift too far off from the Rakefile changes.


soulshake:
I have mixed feelings about this. It seems like a lot of complexity and new dependencies to add for relatively small gain. I feel like it would be better to figure out a way to make the NAT IPs static rather than dynamically updating the docs to reflect changes.

On the other hand, dynamic doc updates are a good thing to have a pattern for, since there are surely other things that it would be nice to keep up to date in an automated way.

P.S. It'd be nice to add a description for this PR, to make it easier to pick out of a lineup in case someone needs to track down a bug later, as well as some code comments so it's easier for others to copy this pattern in the future.

Rakefile
```diff
@@ -2,28 +2,37 @@
...
-  FileUtils.rm_r('_site') if File.exist?('_site')
+  rm_rf('_site')
 end
 
 def print_line_containing(file, str)
   File.open(file).grep(/#{str}/).each do |line| puts "#{file}: #{line}" end
 end
 
+def dns_txt(hostname)
+  require 'faraday'
```
soulshake: Is it a normal ruby pattern to have `require`s here?

BanzaiMan: I would say it is not.

Technically it is a small optimization if this method is not invoked, but it seems to me that these libraries are commonly used to the point where frequent `require`s are more nuisance.

meatballhat: My perspective is that ruby authoring in the context of a Rakefile is a sub-genre of its own, where deferring requirements like this is much more common and a generally good idea.... _but_ in _this case_, I fully admit that my tendency to defer requirements is overkill given the typical commands that are run. :grin:

meatballhat:
@soulshake My perspective is that even _if_ we could ensure the NAT IPs were completely static(ish ... I mean, it's EC2, so it's relative), I would still want this automation to ensure any changes/additions/removals/whatever were automatically updated in the docs.  The difference between the current NAT IP situation and something more static is very small, imho. ❤️

plaindocs:
I've looked at this in great detail, but I'm very much in favor of automating the details in the docs. So regarding @soulshake 's comment, that also makes sense, but I'd prefer to merge in anything that helps now, and if we can fix by using static IPs later, even better.

PR #1584, opened 2017-12-04
Add more details about our python support

bogdanap:
Since we still get a lot of tickets/questions around how we set-up the python environment, I thought I'd try and add more details.
The majority of issues users encounter seem to be around the use of `virtualenv` and `pip`, so I tried to clarify those specific bits.

Another section that is missing is about our support for tox, but that requires some changes in travis-build first, so I'll open a separate PR.


plaindocs:
Thanks @bogdanap

PR #1593, opened 2017-12-07
Be more explicit about escaping strings with spaces in Environment Variables

jstoiko:
I feel like this is a common-enough scenario to deserve a note in the docs (and in the Settings page section).


jstoiko:
I feel like some sort of dynamic `warning` - that would turn the "Value" input red for example, only if the value contains escapable characters - would be even better.

plaindocs:
Thanks

PR #1594, opened 2017-12-07
Add stages section to very first code sample

bittner:
The stages feature doesn't seem to work without an explicit `stages:` section. This PR adds that section to the very first code sample in the Build Stages section of the Travis docs.

See [build 313007058](https://travis-ci.org/edestecd/puppet-software/builds/313007058) (without) and [build 313013400](https://travis-ci.org/edestecd/puppet-software/builds/313013400) (with stages section) for comparison.

### DIFF
--- a/user/build-stages.md
+++ b/user/build-stages.md
@@ -56,6 +56,9 @@ Here’s how you’d set up the build configuration for this in your .travis.yml
 file:
 
 ```yaml
+stages:
+  - test
+  - deploy
 jobs:
   include:
     - stage: test


BanzaiMan:
The build you pointed to does not define build stages at all. https://github.com/edestecd/puppet-software/blob/639deddadf4cca2120a6da196ed45f5b0ccf449b/.travis.yml

The example here should work, as it includes `jobs.include`.

bittner:
Then I referenced a wrong build. I amended the commit to the code with the jobs.include (but without stages) section. I'm pretty sure it didn't work and the YAML looked as follows:

```yaml
---
language: ruby

install:
  - wget 'https://pm.puppetlabs.com/cgi-bin/pdk_download.cgi?dist=ubuntu&rel=16.04&arch=amd64&ver=latest' -q -O pdk.deb && sudo dpkg -i pdk.deb && rm pdk.deb

jobs:
  include:
    - stage: validate
      script: pdk validate 2> /dev/null

    - stage: test
      script: pdk test unit 2> /dev/null
```

BanzaiMan:
The [current content](https://github.com/travis-ci/docs-travis-ci-com/blob/4c04a491cff555236427507c9434120436ce1576/user/build-stages.md) is:

```yaml
jobs:
  include:
    - stage: test
      script: ./test 1
    - # stage name not required, will continue to use `test`
      script: ./test 2
    - stage: deploy
      script: ./deploy
```

I used this configuration in https://github.com/BanzaiMan/travis_production_test/commit/f58cab67d752ecd334b51eaf2d90904a4be9d277#diff-354f30a63fb0907d4ad57269548329e3R8, and produced https://travis-ci.org/BanzaiMan/travis_production_test/builds/313141642.

As you can see, this minimal example works as is. It is not very clear to me how your configuration failed, but I do not believe that this change is required.

bittner:
As you wish, no problem. Just wanted to make sure no-one needs to struggle in future as I did. But maybe I overlooked something.

bittner:
I noticed that http://lint.travis-ci.org/ doesn't yet support build stages. Is there a newer version of the linter?

PR #1595, opened 2017-12-11
Remove useless use of cat

The-Compiler:
See e.g. http://porkmail.org/era/unix/award.html

cc @schultyy


plaindocs:
LGTM thanks

schultyy:
yay - thank you for the change! 🎉

PR #1603, opened 2017-12-14
We dont use this, send people to the blog instead

plaindocs:
FYI @biancawilk


venkat47:
I did not get it fully

PR #1618, opened 2017-12-21
Add brew bundle to upgrade outdated brew packages

stephengroat:


user/reference/osx.md
```diff
@@ -84,7 +84,23 @@ Or it can result in the command not found:
 xctool: command not found
 ```
 
-This is intended behaviour from Homebrew's side, but you can get around it by running first checking if the command needs an upgrade with `brew outdated`, like this:
+This is intended behaviour from Homebrew's side, but you can get around it by using [`brew bundle`](https://github.com/Homebrew/homebrew-bundle) or by running first checking if the command needs an upgrade with `brew outdated`
```
plaindocs: would you mind rewording "running first checking"

and removing simple from "simple `Brewfile`" and easily from "You can then easily update"

Thanks

plaindocs:
Otherwise, looks great :-)

stephengroat:
@plaindocs done

plaindocs:
Thanks @stephengroat !

PR #1630, opened 2017-12-30
Fix Grammer Issue

trafort:
Comma Between ,and or and,


BanzaiMan:
This is correct as is.

PR #1631, opened 2018-01-01
use .test instead of .dev

tduehr:
.dev is owned by Google and has been added to HSTS preload. It should not be used for this purpose.


plaindocs:
Good catch! Thanks @tduehr

PR #1638, opened 2018-01-08
Add `-L` flag to `curl` command

joshtronic:
The URL to download `phpunit` is returning a 302 status code and the `curl` command was failing to download the file. Added the `-L` flag to tell `curl` to follow to the new location.


BanzaiMan:
Thank you!

joshtronic:
Always happy to help! :+1:

PR #1644, opened 2018-01-10
Explain use of `pg_config`

BanzaiMan:
Resolves https://github.com/travis-ci/travis-ci/issues/9011.

### DIFF
--- a/user/database-setup.md
+++ b/user/database-setup.md
@@ -201,9 +201,9 @@ before_install:
 
 ### Using `pg_config`
 
-If your builds rely on the use of the `pg_config` command, you need to install an additional
-apt package `postgresql-server-dev-X.Y` for the version of PostgreSQL you are using in order
-to obtain meaningful output from the command.
+If your builds rely on the `pg_config` command, you need to install an additional
+apt package `postgresql-server-dev-X.Y`, where `X.Y` matches the version of PostgreSQL
+you are using.
 
 For example:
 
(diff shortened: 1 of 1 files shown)


user/database-setup.md
```diff
@@ -199,6 +199,23 @@ before_install:
...
+apt package `postgresql-server-dev-X.Y` for the version of PostgreSQL you are using in order
+to obtain meaningful output from the command.
+
+For example:
+
+```yaml
+addons:
+  apt:
+    packages:
+      - postgresql-server-dev-9.4
```
plaindocs: Just to clarify, the X.Y / 9.4 needs to match the existing postgre they are using? So either the default, or the one they specify intentionally?

BanzaiMan: Any one that the user intends to use. This assumes:

```yaml
addons:
  postgresql: "9.4"
```

Perhaps I can add that.

BanzaiMan: That is to say:

```yaml
addons:
  postgresql: "9.4"
  apt:
    packages:
      - postgresql-server-dev-9.4
```

plaindocs: How about 

> If your builds rely on the `pg_config` command, you need to install an additional
apt package `postgresql-server-dev-X.Y` where X.Y matches the version of PostgreSQL you are using.

BanzaiMan: That sounds great! :+1:

PR #1671, opened 2018-01-19
Explain use of chrome on containers

### LINKED ISSUE #1668: Improve Chrome's documentation
Unfortunately, this is a side-effect of the work we're doing to mitigate the Spectre/Meltdown exploit, which required us to perform maintenance on our container-based infrastructure. One of the side effects of this is that it's caused issues with Chrome running on that infrastructure, probably due to the underlying kernel update.

At the moment, we recommend routing these builds to our `sudo: required` infrastructure, which is not affected by this problem. Alternatively, it's also possible to disable the sandbox by adding the --no-sandbox flag in the command you use to run Chrome.

More information in this public issue https://github.com/travis-ci/travis-ci/issues/8836#issuecomment-357441064

BanzaiMan:
This explains how to avoid https://github.com/travis-ci/travis-ci/issues/8836.

Closes travis-ci/travis-ci#8836

Closes #1668

### DIFF
--- a/user/chrome.md
+++ b/user/chrome.md
@@ -26,34 +26,32 @@ You can use Google Chrome in [headless mode](/user/gui-and-headless-browsers/#Us
 For security reasons, Google Chrome is unable to provide sandboxing when it is running in the
 [container-based environment](https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments).
 
-To use Chrome in the container-based environment:
+In that case, you may see an error message like this:
 
-1. Pass `--no-sandbox` when invoking the `chrome` command.
+```
+30 11 2017 13:35:42.245:ERROR [launcher]: Cannot start Chrome
+  [4315:4315:1130/133541.781662:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /opt/google/chrome/chrome-sandbox is owned by root and has mode 4755.
+```
 
-  ```yaml
-  sudo: false
-  addons:
-    chrome: stable
-  script:
-    - chrome --no-sandbox
-  ```
-{: data-file=".travis.yml"}
-1. Your testing framework may have a different mechanism to supply this flag to the `chrome` executable.
-  For example, with [the customlauncher plugin for Karma](https://github.com/karma-runner/karma-chrome-launcher), you would add it to the `flags` array:
-  
-  ```javascript
-  module.exports = function(config) {
-    config.set({
-      browsers: ['Chrome', 'ChromeHeadless', 'ChromeHeadlessNoSandbox'],
+To use Chrome in the container-based environment, pass the `--no-sandbox` flag to the `chrome` executable.
 
-      // you can define custom flags
-      customLaunchers: {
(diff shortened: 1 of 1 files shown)


user/chrome.md
```diff
@@ -19,3 +20,18 @@ addons:
...
+[container-based environment](https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments).
+
+To use Chrome in the container-based environment, pass `--no-sandbox` when invoking the `chrome` command.
+
+```yaml
+sudo: false
+addons:
+  chrome: stable
+script:
+  - chrome --no-sandbox
```
meatballhat: Is this `script` content a realistic use case?

BanzaiMan: I am afraid not, but I could not think of a more realistic one. 😬 Suggestions welcome!

cotsog:
@BanzaiMan Thanks for this PR!

I think you can close this issue afterward => https://github.com/travis-ci/docs-travis-ci-com/issues/1668.

user/chrome.md
meatballhat: I recommend giving an example in the context of one of the example projects referenced by a Google blog post: https://github.com/ebidel/headless-karma-travis

This would mean that it's a combination of an example `.travis.yml` change and a change to one's `karma.conf.js`, and you could say something about doing similarly for capybara or whatever else.

meatballhat: Here's the specific spot where the chrome runner flags are managed in the case of the headless-karma-travis example: https://github.com/ebidel/headless-karma-travis/blob/da2131fbe130c45d32ea3fa138d9d475baf81f7e/karma.conf.js#L16-L18

user/chrome.md
```diff
@@ -19,3 +20,40 @@ addons:
...
+To use Chrome in the container-based environment:
+
+1. Pass `--no-sandbox` when invoking the `chrome` command.
+
+  ```yaml
+  sudo: false
+  addons:
+    chrome: stable
+  script:
+    - chrome --no-sandbox
```
meatballhat: I don't see the value in keeping this YAML example, given that I don't think it even works correctly and is not a common use case AFAIK.

user/chrome.md
```diff
@@ -19,3 +20,40 @@ addons:
...
+  For example, with [the customlauncher plugin for Karma](https://github.com/karma-runner/karma-chrome-launcher), you would add it to the `flags` array:
+  
+  ```javascript
+  module.exports = function(config) {
+    config.set({
+      browsers: ['Chrome', 'ChromeHeadless', 'ChromeHeadlessNoSandbox'],
+
+      // you can define custom flags
+      customLaunchers: {
+        Chrome_no_sandbox: {
```
meatballhat: For this to be a valid configuration, the key name must match what is in the above `browsers` array, such as `MyHeadlessChrome` in this example: https://github.com/ebidel/headless-karma-travis/blob/da2131fbe130c45d32ea3fa138d9d475baf81f7e/karma.conf.js#L12-L18 (IIUC)

user/chrome.md
BanzaiMan: I agree.

BanzaiMan:
Did I write "blush up"? 😊

PR #1675, opened 2018-01-22
[Enterprise] Move ops manual "contact support" to snippets

acnagy:
This was going to be a PR for turning the ops manual into it's own subdirectory of shorter articles, but the structure didn't quite work out, and was very different from the pattern of the long documents we've done instead ('Deployments Overview' or 'Common Build Problems'), for example. However, if/when that changes, we should probably come back to this idea.  

However, instead it occurred to me that we should be re-using the contact-us text from the ops-manual. Now, this PR is just about that, with a few wording tweaks in place.


_data/snippets.yml
```diff
@@ -34,3 +34,30 @@ trusty_note_no_osx: |
   > information are in our reference pages:
   >  * [Precise](/user/reference/precise/)
   >  * [Trusty](/user/reference/trusty/)
+
+contact_enterprise_support: |
```
plaindocs: When you put the `##` heading *inside* the snippet, you need to be careful at what level you level you include it. Worth thinking about just including the text here instead?

_data/snippets.yml
```diff
@@ -34,3 +34,30 @@ trusty_note_no_osx: |
...
+  [enterprise@travis-ci.com](mailto:enterprise@travis-ci.com). If possible, 
+  please include as much of the following info as you:
+   
+  - What is the problem?
+  - Which steps did you try already?
+  - A support bundle (You can get it from `https://yourdomain:8800/support`)
+  - Worker log files (They can be found at `/var/log/upstart/travis-worker.log`) 
+  - If you're using multiple worker machines, we need the log files from all of them.
+  
+  Is anything special with your setup? While we may be able to see some
```
plaindocs: Should we reword "Is anything special with your setup?" to something like "Have you made any particular customizations to your setup?" or something similar?

Maybe use long form of Infrastructure as a Service?

_data/snippets.yml
```diff
@@ -34,3 +34,30 @@ trusty_note_no_osx: |
...
+  Therefore , we'd like to ask you to also answer the questions below in your 
+  support request (if applicable):
+  
+  - How many machines are you using?
+  - Do you use configuration management tools (Chef, Puppet)?
+  - Which other services do interface with Travis CI Enterprise?
+  - Do you use Travis CI Enterprise together with github.com or GitHub Enterprise?
+  - If you're using GitHub Enterprise, which version of it?
+  
+  We're looking forward helping!
```
plaindocs: to!

_data/snippets.yml
acnagy: works for me! 👍

_data/snippets.yml
acnagy: I like this! used `Have you made any customizations to your setup?`

_data/snippets.yml
acnagy: done ✅

_data/snippets.yml
```diff
@@ -51,3 +51,27 @@ environment_variables: |
 
   - differ per repository.
   - contain sensitive data, such as third-party credentials.
+contact_enterprise_support: |
+  To get in touch with us, please write a message to 
+  [enterprise@travis-ci.com](mailto:enterprise@travis-ci.com). If possible, 
+  please include as much of the following info as you:
```
plaindocs: missing "can" at the end

_data/snippets.yml
```diff
@@ -51,3 +51,27 @@ environment_variables: |
...
+   
+  - What is the problem?
```
plaindocs: We maybe need to rewrite these? Because "what is the problem" is not a piece of info. So it would become

- a description of the problem
...

Alternatively, reword the preceeding sentence to something like, "answer as many o the questions as you can"

_data/snippets.yml
```diff
@@ -51,3 +51,27 @@ environment_variables: |
...
+  - Which steps did you try already?
+  - A support bundle (You can get it from `https://yourdomain:8800/support`)
+  - Log files from all workers (They can be found at `/var/log/upstart/travis-worker.log` - please include as many as you can retrieve).
```
plaindocs: and these aren't pieces on info either:-p

PR #1676, opened 2018-01-22
Add more information on how to use Tox with Travis

bogdanap:
I have the feeling that since Tox is clashing a bit with the way we are running Python jobs (both create separate virtualenvs), a lot of users report issues with it.

I thought adding more details and a recommendation on what the best way to use python and tox in a Travis config would help reduce the number of questions/tickets to support.


bogdanap:
Tried to add more explanations into a separate section for Tox. 
I'm not super-good with words so am happy to hear lots of feedback on how this could be reformulated to increase clarity 😄

user/languages/python.md
```diff
@@ -150,6 +150,16 @@ If you do not provide a `script` key in a Python project, Travis CI prints a
...
 and fails the build.
 
+## Using Tox as the Build Script
+
+Due to the way Travis is designed, interaction with [tox](https://tox.readthedocs.io/en/latest/) is not straightforward. 
+As described [above](/user/languages/python/#Travis-CI-Uses-Isolated-virtualenvs), Travis already runs tests inside an isolated virtualenv whenever `language: python` is specified, so please bear that in mind whenever creating more environments with tox.
+
+If you would prefer to run tox outside the Travis-created virtualenv, it might be a better idea to use `language: generic` instead of `language: python`.
+However, this means that you will only have the default Ubuntu python versions available (2.7.6 and 3.4.3 for Trusty). Any aditional Python versions you would like to test would have to be installed manually.
+
```
plaindocs: @bogdanap this is definitely better. The added context around the problem is helpful. 

Do you think it would make sense to break it down more directly into something like:

If you need to test multiple versions of python with tox, you have two options:

* use `language: generic` (not `language: python`) and install the versions of python you need  (2.7.6 and 3.4.3 for Trusty are available on all images)

* use `language: python` and specify versions using the `python` key

I think it would need a tiny bit more detail on the intreaction of tox in each though. 

Or do you think this is not necesary?

bogdanap:
@plaindocs Sorry for being so slow with this change. Updated the text according to your suggestions 😄

plaindocs:
Thank *you* for implementing so many change requests. :-)

PR #1683, opened 2018-01-30
Update to match permissions behaviour

MariadeAnton:
Moving forward, I believe we could use a table so that this is clearer, kind of what GitHub is doing at https://help.github.com/articles/repository-permission-levels-for-an-organization/


MariadeAnton:
That would be perfect, I think - thanks a lot, Sam!

PR #1684, opened 2018-01-30
Switch to dnsjson.com for DNS lookups

meatballhat:
as it is slightly more reliable than dig.jsondns.org


plaindocs:
Awesome! :-)

PR #1686, opened 2018-02-03
Add link to encryption for env vars

nmagee:
I was using the Docker build steps and had to go looking for encryption steps. I thought it would be useful to remind users to do so, and how.


plaindocs:
Thanks!

PR #1687, opened 2018-02-04
Add JWT deprecation warning and .warning class

BanzaiMan:
jwt_addon_-_travis_ci

### DIFF
--- a/assets/stylesheets/_base.sass
+++ b/assets/stylesheets/_base.sass
@@ -100,3 +100,7 @@ table
         background-color: $cream-light
         color: $grey-dark
         border-right: 1px $cream-dark solid
+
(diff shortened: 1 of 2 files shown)


plaindocs:
Can we tone down the red slightly?

BanzaiMan:
The only other red we have defined is `$red-light`, which gives us

jwt_addon_-_travis_ci

We can choose another color, but I'd rather not.

plaindocs:
I'm ok with `red-light` if you are, and I definitely prefer it to `brick-red`

BanzaiMan:
Very well!

plaindocs:
Danke! <3

Issue #732, opened 2016-09-15
Examples on Projects with Multiple Language


acnagy:
We need more examples about how to setup a build for a project that uses multiple languages. Likely, this should be in the build matrix section, but perhaps in Customizing the Build as well. Originally requested over [Twitter](https://twitter.com/user/status/776159371269120000)

PR #1689, opened 2018-02-06
Initial thoughts on multi os

plaindocs:
Fine tuning build matrix and multi os docs, maybe heading into breakign the build territory.

Related to #732


plaindocs:
s/rows/jobs/ thanks @MariadeAnton

user/build-matrix.md
```diff
@@ -0,0 +1,294 @@
+---
+title: Build Matrix
+layout: en
+---
+
+The build matrix is a way to run builds on multiple versions of language runtimes, on multiple operating systems, and with multiple versions of environment variables.
```
BanzaiMan: > multiple versions of environment variables.

doesn't seem right to me. Perhaps "different values of environment variables". The matrix expansion over `env` is not necessarily tied to a single environment variable, either, so I'm not sure how best to say it.

```yaml
env:
  - VAR1=val1
  - VAR1=val2
  - VAR2=val2
```

user/build-matrix.md
```diff
@@ -0,0 +1,294 @@
...
+
+<div id="toc"></div>
+
+When you combine the three main configuration options of *Runtime*, *Environment* and *Exclusions/Inclusions* you have a matrix of all possible combinations.
+
+Each programming language environment on Travis CI has a specific list of keys which automatically populate the build matrix. For Ruby, those are `rvm`, `gemfile`, `env`, and `jdk`.
```
BanzaiMan: Are we enumerating the language-specific expansion keys? If so, `env` should not appear here.

At the same time, the previous paragraph mentions "*Runtime*", so having `rvm` here is also confusing.

Maybe, we can rewrite this paragraph with something like:

> The set of keys which are expanded to create the build matrix varies from one programming language to another. For Ruby, they are `env`, `gemfile`, `jdk`, and `rvm`. These are specifically mentioned in the reference documentation for each language as "matrix keys".

user/build-matrix.md
plaindocs: :100:

user/build-matrix.md
plaindocs: :100:

PR #1691, opened 2018-02-06
Update pypi.md to make command effect obvious

mueslo:
I repeated this command multiple times since there was no output. Took a bit of frustration until I realised that it directly wrote to `.travis.yml` instead of stdout.


plaindocs:
Thanks for the suggestion! 

How about something that flows a little more? Maybe

> We recommend you [encrypt](/user/encryption-keys/) your password and add it to your `.travis.yml` by running:

Issue #1649, opened 2018-01-11
ssh-known-hosts seems to encourage poor security practices


stefangotz:
The ssh-known-hosts configuration option documented in https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/ssh-known-hosts.md seems to encourage poor security practices. It may make Travis builds susceptible to man-in-the-middle attacks.

If I understand the effect of the ssh-known-hosts option correctly, it makes a Travis build instance retrieve an SSH server's key afresh every time the build instance starts. This would more or less defeat the purpose of the SSH known hosts feature.

The point of SSH's known hosts feature is to pin down the cryptographic identity of a server so that a man in the middle attacker cannot impersonate the server. On a regular computer, the standard approach of simply accepting and storing the server key on the first connection and then checking it on every subsequent connection makes sense because the client remembers and checks the server key for a long time. On short lived systems, such as Travis build instances, this approach makes less sense because almost every connection is the 'first' connection and therefore the build instance accepts any key the server (or an impersonator) presents for every single build.

Yes, of course such attacks are extremely unlikely. Still, how about updating the documentation to help users to make an informed decision about this? Would it be worth mentioning that there is a security vs. convenience trade-off here?

For short lived systems, the safer option is to obtain the server key ahead of time and to add that fixed key to the .known_hosts file of the Travis build instance. Basically, one could run ssh-keyscan on a local computer against the SSH server or git repo server of interest, get the new entries for that server from the local .known_hosts file, and configure Travis to add those entries to the .known_hosts file of each Travis build instance. That would pin down the server's cryptographic identity as SSH normally intends it to be done. And there might be other, more convenient alternatives to achieve the same effect.

BanzaiMan:
We run `ssh-keyscan` to add the SSH host key. https://github.com/travis-ci/travis-build/blob/d08e1022dca73e813fc0ed8c73482b91005c9f82/lib/travis/build/addons/ssh_known_hosts.rb#L48 Do you have a better idea?

BanzaiMan:
We also welcome your documentation PR if you have ideas for improvement. Thanks.

stefangotz:
> We run ssh-keyscan to add the SSH host key. Do you have a better idea?

The security issue is not so much ssh-keyscan itself, but the fact that the server key is retrieved for every single new build.

[...]

PR #1692, opened 2018-02-06
Reorganize ssh-known-hosts page

BanzaiMan:
Explain the new means of adding previously-known SSH host keys

An additional minor change is eschewing die.net, which tends to be
outdated.

To correspond to https://github.com/travis-ci/travis-build/pull/1304.

Also, see #1649.

### DIFF
--- a/user/ssh-known-hosts.md
+++ b/user/ssh-known-hosts.md
@@ -12,8 +12,60 @@ your git repository, which is necessary if there are git submodules
 from domains other than `github.com`, `gist.github.com`, or
 `ssh.github.com`.
 
-Both hostnames and IP addresses are supported, as the keys are
-added via `ssh-keyscan`.  A single host may be specified like so:
+Each entry in the configuration is one of the following:
+
+1. a hash with `host`, `type`, `key`
+1. a hostname
+1. a numeric IP address
+
+In the first case, we write key directly to `~/.ssh/known_hosts`.
+
+In the last two cases, we run `ssh-keyscan` to add the host key.
+See [Security Implications](#Security-Implications) below to understand what this means.
+
+## Examples
+
+Hosts may be given in various ways.
+
+### A single hash with `host`, `type`, `key`
+
+In the case where a host is defined with `host`, `type`, and `key`,
+
+```yaml
+addons:
+  ssh_known_hosts:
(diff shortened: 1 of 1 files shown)


user/ssh-known-hosts.md
```diff
@@ -12,16 +12,69 @@ your git repository, which is necessary if there are git submodules
...
 
-Both hostnames and IP addresses are supported, as the keys are
-added via `ssh-keyscan`.  A single host may be specified like so:
+Each entry in the configuration is one of the following:
+
+1. a hash with `host`, `type`, `key`
+1. a hostname
+1. a numeric IP address
+
+In the first case, we write key directly to `~/.ssh/known_hosts`.
```
plaindocs: s/write key/write the key/ 
or
s/write key/write `key`/  ?

user/ssh-known-hosts.md
```diff
@@ -12,16 +12,69 @@ your git repository, which is necessary if there are git submodules
...
+
+In the last two cases, we run `ssh-keyscan` to add the host key.
+See [Security Implications](#Security-Implications) below to understand what this means.
+
+## Examples
+
+Hosts may be given in various ways.
```
plaindocs: How about 

> Here are some examples of the different ways to add hosts

user/ssh-known-hosts.md
```diff
@@ -42,37 +97,8 @@ addons:
...
-Add the key server's public key *KEY* to the SSH `known_hosts` file, e.g., with the following addition to the installation phase:
-
-```yaml
-install:
-  - echo 'KEY' >> $HOME/.ssh/known_hosts
-```
-{: data-file=".travis.yml"}
-
-Make sure to replace *KEY* with the complete line of text containing the public key of the SSH server as obtained in the previous step.
+We recommend using the first form.
```
plaindocs: We talk about cases above, so might be worth being more explicit like 

> We always recommend adding hosts using the hash with `host`, `type`, `key`.

BanzaiMan: I avoided "always", since this is a recent improvement. I don't want existing configurations to suddenly become "always against the recommendation".

plaindocs: Makes sense.

PR #1697, opened 2018-02-07
Common Build Problems: Uses the new name of Capybara.default_wait_time

evankroske:
`Capybara.default_wait_time` is now `Capybara.default_max_wait_time`:

https://github.com/teamcapybara/capybara#using-capybara-with-rspec


plaindocs:
Thanks!

PR #1698, opened 2018-02-08
Document yarn test

xPaw:
Corresponding PR: https://github.com/travis-ci/travis-build/pull/1299


user/languages/javascript-with-nodejs.md
```diff
@@ -164,6 +164,7 @@ Note that `yarn` requires Node.js version 4 or later.
 If the job does not meet this requirement, `npm install` is used
 instead.
 
+If Yarn is supported, test command is run using `yarn test` instead of `npm test`.
```
BanzaiMan: I think the best place to put this information is the "Default Build Script" section above. But very much like here, adding a "Yarn is supported" subsection.

xPaw: Is that better

plaindocs:
Thanks for looking into this.

Once the change is made in travis-ci, the docs should probably read something like

> If `yarn.lock` exists, the default build command will be `yarn test` instead of `npm test`.

what do you think?

BanzaiMan:
@xPaw What do you think of @plaindocs' suggestion?

xPaw:
@BanzaiMan doesn't it already say pretty much the same thing? Unless you mean the exact phrasing.

BanzaiMan:
`Yarn is supported` is ambiguous here.

xPaw:
Changed it.

plaindocs:
Awesome, thanks, good to go here as soon as the feature is merged in.

PR #1700, opened 2018-02-08
Fix URL for Sauce connect proxy

takluyver:
The `https://` url doesn't work. At the suggestion of [this SO post](https://stackoverflow.com/questions/48236104/ssl-errors-using-sauce-labs-in-travis-ci-with-selenium-webriver-tests-django-pr), I changed it to `http://` and got a step further.


BanzaiMan:
Thank you!

PR #1701, opened 2018-02-09
Correct filename in database setup documentation

nholden:
The database setup documentation provides an example `config/database.yml` file for using MySQL with ActiveRecord, but it's mislabeled as `.travis.yml`. This corrects this label.

screen shot 2018-02-08 at 6 07 59 pm


plaindocs:
Good catch, thanks!

PR #1712, opened 2018-02-19
Update AWS Lambda dpl provider docs to include changes made in travis-ci/dpl#745

gebn:
Removed optional parameters from minimal example, adding an explanation of credential fall back.


plaindocs:
Thanks, looks fine to me :-)

Issue #456, opened 2015-12-19
s3 artifact documentation incorrectly describes upload path


nelhage:
The [artifact documention](https://github.com/travis-ci/docs-travis-ci-com/blob/gh-pages/user/uploading-artifacts.md) claims “Unless you programatically generate unique filenames and folders, artifacts are overwritten every build.”, which suggests every build is uploaded to an identical path.

As best I can tell, however, each build gets its own path, encoding the [repo, build number, and job number](https://github.com/travis-ci/travis-build/blob/b7e1109cbc0af202f40b1922a8a7a181815f0f0b/lib/travis/build/addons/artifacts/env.rb#L42). It'd be great for the documentation to reflect that, and also reflect the `ARTIFACTS_TARGET_PATH` variable that can be used to override that path.

glandium:
I don't think it's worth opening a separate issue so I'll put it here. The same documentation says:

> The default paths uploaded to S3 are found via git ls-files -o in order to find any files in the git working copy that aren’t tracked. If any *additional* paths need to be uploaded, they may be specified via the `addons.artifacts.paths` key

(emphasis mine)

AIUI, `addons.artifact.paths` doesn't complement the default but *replaces* it.

plaindocs:
I'm getting some clarification on this and will update the docs.

mw866:
I am so sure that the documentation can be much clearer.
Based on my testing, I believe this is the behavior of the code.
Correct me if I'm wrong.

| Source: Travis-CI      | Destination: S3                   |
| ---------------------- | --------------------------------- |
| `woring_dir\paths(1)}` | `s3:\\buket_name\target_paths(1)` |
| `working_dir\paths(2)` | `s3:\\buket_name\target_paths(2)` |
| …                      | …                                 |

PR #1715, opened 2018-02-21
Clarify artifacts' S3 upload paths

BanzaiMan:
Resolves https://github.com/travis-ci/docs-travis-ci-com/issues/456.

### DIFF
--- a/user/uploading-artifacts.md
+++ b/user/uploading-artifacts.md
@@ -9,8 +9,7 @@ layout: en
 
 Travis CI can automatically upload your build artifacts to Amazon S3, after the
 [`after success`](/user/customizing-the-build/#The-Build-Lifecycle) stage of the
-build. Unless you programatically generate unique filenames and folders,
-artifacts are overwritten every build.
+build.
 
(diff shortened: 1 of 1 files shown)


user/uploading-artifacts.md
```diff
@@ -80,6 +79,21 @@ addons:
 ```
 {: data-file=".travis.yml"}
 
+### Target Paths
+
+By default, artifacts will be uploaded to the path in the bucket
+defined by `/${TRAVIS_REPO_SLUG}/${TRAVIS_BUILD_NUMBER}/${TRAVIS_JOB_NUMBER}`.
+This value can be overridden by the `target_paths` option.
```
plaindocs: I'd prefer something like.

You can change the upload path at build time using the `target_paths` key, for example:

plaindocs: But regardless, :shipit: !

Ah, one question, `paths` plural?

BanzaiMan: Yes. `target_paths` is plural.

PR #1723, opened 2018-02-25
Update nix.md to reflect current version

badi:
The old doc indicated that only Nix 1.11.2 was installed, when in fact the current version from https://nixos.org/nix/install is used by the [travis-build](https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/nix.rb#L43) `nix.rb` script.

~`nix-shell` is removed from the list since it is now a link to `nix-build`~

cc @domenkozar @garbas @matthewbauer

---

edit: nix-shell


domenkozar:
nix-shell is a symlink, but it's a completely different interface than nix-build - so we should keep it.

badi:
Whoops. I saw the symlink and assumed that Nix 2 had deprecated it. Fixed.

PR #1725, opened 2018-02-26
[Enterprise] Docker in Docker for Trusty build images

schultyy:
This PR updates the Docker in Docker section for Trusty build images. Because these days Docker-Engine is no more and has been replaced with Docker-CE, the old approach we've done with Precise doesn't work anymore. There we had Docker actually running in Docker. 
The approach for Trusty build images looks a bit different though. Here we're using the Host's Docker daemon to execute the commands. The disadvantage is: the user has to clean up after they're done building, the advantage though: fast interaction with Docker ✨

### DIFF
--- a/user/enterprise/build-images.md
+++ b/user/enterprise/build-images.md
@@ -79,7 +79,7 @@ With Trusty build images a few additional steps are required. Since `docker-ce` 
 
 On Amazon EC2 each machine has a private IP address by default. That address will be used to connect to the Docker daemon.
 
-To get the address, please run `ifconfig` in your terminal. For EC2 Virtual Machines, usually `eth0` interface is what we're looking for.
+To get the address, please run `ifconfig` in your terminal. For EC2 Virtual Machines, usually `eth0` interface is correct.
 
 Now the Docker daemon needs to listen to that address. Please open `/etc/default/docker`. In this file, you'll find the `DOCKER_OPTS` variable. This is read by Docker during service startup. In there please configure the additional host like this:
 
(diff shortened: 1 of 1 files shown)


user/enterprise/build-images.md
```diff
@@ -32,24 +32,71 @@ For example, in order to install a particular Ruby version which is not
...
 
 ```
     export TRAVIS_WORKER_DOCKER_PRIVILEGED="true"
-```       
+```
+
+#### Additional configuration for Trusty build images
+
+With Trusty build images a few additional steps are required. Since Docker-ce can't run on its own inside another Docker container, it'll connect to the Host's Docker daemon to execute the respective commands.
+
```
acnagy: Maybe:
> Configuration for Docker Builds in Trusty Build Environments

for the header? Also: I'd do the code formatting for `docker-ce`

user/enterprise/build-images.md
```diff
@@ -32,24 +32,71 @@ For example, in order to install a particular Ruby version which is not
...
+> For non-AWS deployments, please make sure that the worker machine has a private IP address configured on one of its interfaces.
+
+On Amazon EC2 each machine has a private IP address by default. That address will be used to connect to the Docker daemon.
+
+To get the address, please run `ifconfig` in your terminal. For EC2 Virtual Machines, usually `eth0` interface is what we're looking for.
```
acnagy: > usually `eth0` interface is what we're looking for.

I'd try:

> usually `eth0` interface is needed

acnagy: that said, @plaindocs will probably tell me off for using the passive voice, so:
> usually the `eth0` interface is correct. 

is _probably_ be better

acnagy:
nice! @schultyy - I think this is good to go! 🚀

PR #1726, opened 2018-02-27
PyPI: Swap `api_key` field for `user`/`password`

di:
PyPI doesn't have any notion of "API keys", so this field shouldn't be here. This was likely a copy/paste error.


plaindocs:
Thanks!

PR #1727, opened 2018-02-28
[Enterprise] Custom Docker images for Trusty build environment

### LINKED ISSUE #1643: [Enterprise] Update Customizing Build Images
Naming convention for build images has changed for more recent versions of worker, which means the tagging instructions in our build environment customization docs need to be updated to reflect.

schultyy:
This PR updates the "Custom Build images" page with information about our Trusty build environment.

Resolves #1643

### DIFF
--- a/user/enterprise/build-images.md
+++ b/user/enterprise/build-images.md
@@ -20,6 +20,29 @@ place, you can fully customize these images according to your needs.
 **Note**: you'll need to re-apply your customizations after
 upgrading build images from [quay.io](https://quay.io/organization/travisci).
 
+### Ubuntu Trusty build environments
+
+For Ubuntu Trusty build environments we ship three Docker images in total. Depending on the user's `.travis.yml` configuration we will pick the corresponding image to run the build.
+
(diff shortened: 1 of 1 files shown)


user/enterprise/build-images.md
```diff
@@ -20,6 +20,29 @@ place, you can fully customize these images according to your needs.
...
+We're shipping the same Docker build images as travis-ci.org and travis-ci.com are using. The base image, `connie` contains all databases and frameworks preinstalled, such as postgresql, mysql, memcached, pyenv, rvm, gimme. Though there are no interpreters available. Based on `connie` there is `garnet`, which adds the following programming languages:
+
+- Ruby
+- Node.js
+- Go
+- PHP
+- Python
+- Java / JVM
+
+The third image, `amethyst`, additionaly ships with Android, Erlang, Haskell and Perl preinstalled.
```
acnagy: Hmm... I think the context here is good, but conceivably, if image names change, remembering to document this might tricky. Perhaps we could copy the work @BanzaiMan did in the 'testing docker images locally' segment here: https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/common-build-problems.md to reference which languages are where?

schultyy: Hmm good point – Though I think that this is manageable because the next time image names will change is when we're shipping Xenial build images. My feeling is that we have to update our docs for that anyways.

acnagy: fair point - and, if anything of this changes, I think we'll have enough warning to implement something more general here too

PR #1731, opened 2018-03-01
Locking configure file at latest commit

evertonfraga:
Today, every Travis job that uses the recommended Travis file (for meteor builds) taken from the official docs is trusting that arunoda's script will never change to a malicious code. If @arunoda's account gets compromised, every build that uses the former URL is at risk as well. A malicious script could steal code signing certificates, AWS keys, and so on.

It is a naive behavior and from security standpoint, it should be locked to the commit hash.


plaindocs:
Makes sense. Thanks for pointing this out.

evertonfraga:
no problem, thanks for merging.

PR #1732, opened 2018-03-02
Add strategies for `Ready state command canceled: context deadline exceeded`

schultyy:
This error can be caused by two things:
- Configuration mismatch in GitHub OAuth App
- Hostname mismatch between server and license

We've seen this a couple of times in our support requests recently and therefore adding this case to the Ops Manual.

### DIFF
--- a/user/enterprise/operations-manual.md
+++ b/user/enterprise/operations-manual.md
@@ -100,7 +100,7 @@ After a fresh installation or configuration change the Travis CI Enterprise cont
 
 #### GitHub OAuth app configuration
 
-The above mentioned error can be caused by a configuration mismatch in the GitHub OAuth Application. Please check that _both_ website and callback URL contain the Travis CI Enterprise's hostname. If you have discovered a mismatch here, please restart the Travis container from within the admin dashboard.
+The above mentioned error can be caused by a configuration mismatch in [the GitHub OAuth Application](https://docs.travis-ci.com/user/enterprise/prerequisites/#OAuth-App). Please check that _both_ website and callback URL contain the Travis CI Enterprise's hostname. If you have discovered a mismatch here, please restart the Travis container from within the admin dashboard.
 
 #### Hostname does not match license's hostname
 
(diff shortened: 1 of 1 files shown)


user/enterprise/operations-manual.md
```diff
@@ -86,12 +86,26 @@ A source for the problem could be that the worker machine is not able to communi
...
 
+## Enterprise container start fails with `Ready state command canceled: context deadline exceeded`
+
+After a fresh installation or configuration change the Travis CI Enterprise container doesn't start and fails with the error `Ready state command canceled: context deadline exceeded` in the admin dashboard (`https://travis.example.com:8800/dashboard`).
+
+### Strategies
+
+#### GitHub OAuth app configuration
+
+The above mentioned error can be caused by a configuration mismatch in the GitHub OAuth Application. Please check that _both_ website and callback URL contain the Travis CI Enterprise's hostname. If you have discovered a mismatch here, please restart the Travis container from within the admin dashboard.
```
acnagy: Link to the prereq's doc somewhere here?

user/enterprise/operations-manual.md
```diff
@@ -86,12 +86,26 @@ A source for the problem could be that the worker machine is not able to communi
...
+
+#### Hostname does not match license's hostname
+
+Your Travis CI Enterprise license has a hostname field which contains the hostname of your installation. When the license's hostname does not match the actual hostname, the container does not start. If this is the case or you suspect that this might be likely, please [get in touch with us](mailto:enterprise@travis-ci.com?subject=License%20Hostname%20Change) and we'll help you to get back on track.
```
acnagy: This is looking good! That said, I can't remember off the top of my head if this is tied to version - iirc 2.1.11+ are _less_ bound to hostname, but maybe not entirely?

schultyy: > iirc 2.1.11+ are less bound to hostname, but

That's what I thought, too. 🤷‍♂️ 
I think this requires some more investigation

PR #1741, opened 2018-03-07
An experiment in spellchecking

plaindocs:


PatMyron:
@plaindocs Do you know how this module handles capitalization? Wondering if it handles camelCasing

Also, from experience trying something similar, names and usernames were some of the most common false positives. Does `.spelling` support wildcards like `@.*`?

Also CC'ing @dabutvin because we've talked about this idea before

plaindocs:
Two excellent questions, I'll dredge up my memory from quite some time ago, and have a bit of a dig into the docs. 

Also not tied to that particular spell checker, in case you used something else that work out better.

Pezi777:
Hi @PatMyron,
as I took over Plaindocs's documentation tasks, I wanted to ask, if you want to continue working on this pull request or if I can close it? Thank you. Petra

peternewman:
See also https://github.com/codespell-project/codespell/ which isn't specifically aimed at Markdown, but is a curated list of typos, so shouldn't hit false positives.

Full disclaimer, I'm a contributor...

PR #1746, opened 2018-03-08
Show 5 latest Node.js releases

BanzaiMan:


plaindocs:
@BanzaiMan do you mind resolving?

I'm excited about automating more of these version numbers.

BanzaiMan:
Netlify doesn't seem to work with dynamic content like this, though. Am I right?

plaindocs:
Yeah, it should.

It *might* not be working due to how we set the Rake tasks up though. If that is the case, good you noticed.

BanzaiMan:
I'll go ahead and merge this. We'll look at Netifly issue separately.

PR #1742, opened 2018-03-07
Add instructions to get API token from profile page on different docs pages

iriberri:
The revert of the revert

PR #1749, opened 2018-03-09
Try to fix JS issues in API docs

iriberri:
After re-running `rake update_docs`, the js files present the changes I include here.

The diff in the unminified files looks like: https://www.diffchecker.com/fXHjdltP

I'm not sure of what I'm doing here, to be honest, but I see Sam did some tests with this yesterday at https://github.com/travis-ci/docs-travis-ci-com/pull/1742


MariadeAnton:
image
LGTM!

PR #1755, opened 2018-03-14
Update deployment.md

Overbryd:
I propose this change to have a tangible example on how to deploy from a selected list of branches.

This is easily overlooked and hard to debug because of the default `branch: master` setting, and the counter-intuitive setting of `all_branches: true`.


plaindocs:
Looks good to me. Thanks.

PR #1758, opened 2018-03-15
Explain S3's endpoint key

BanzaiMan:
To correspond to https://github.com/travis-ci/dpl/pull/712.

### DIFF
--- a/user/deployment/s3.md
+++ b/user/deployment/s3.md
@@ -265,3 +265,16 @@ deploy:
   expires: "2012-12-21 00:00:00 -0000"
 ```
 {: data-file=".travis.yml"}
+
+### Using S3-compatible Object Storage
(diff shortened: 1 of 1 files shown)


user/deployment/s3.md
```diff
@@ -265,3 +265,16 @@ deploy:
   expires: "2012-12-21 00:00:00 -0000"
 ```
 {: data-file=".travis.yml"}
+
+### Using S3-compatible Object Storage
+
+You can use an S3-compatible object storage such as Digital Ocean Spaces
+with the `endpoint` key.
```
plaindocs: nitpicky: instead of "with the endpoint key" you could say "by setting the endpoint key" or "by specificying it in the endpoint key".

PR #1759, opened 2018-03-16
Compound-clause modifier

stephendonner:
Simple hyphenation for a compound-clause modifier.


plaindocs:
Hi Stephan, thanks for pointing out that that isn't clear.

Your fix doesn't quite work, as sudo-related is also the intent. Maybe we should just reword to

> Any of the following settings related to sudo or docker route your build to a sudo-enabled linux (Ubuntu Trusty) environment on Google Cloud Engine.

stephendonner:
Thanks, @plaindocs - sorry I didn't fully grok the intent, and that reword sounds perfect!

PR #1760, opened 2018-03-16
Move Middleman config to root dir because Heroku

plaindocs:
Apparently 

> Heroku apps expect the app directory structure at the root of the repository. If your app is inside a subdirectory in your repository, it won’t run when pushed to Heroku.

And when we switched to generating the Slate on the fly we started cding into `slate` and running Ruby, which was causing the weird version issues.

:-( :-(

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-#ruby '2.4.3'
+ruby '2.4.3'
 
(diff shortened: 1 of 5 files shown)


plaindocs:
:tada:

selwinm:
Hi from me Selwin.M and I like to say good luck.

PR #1776, opened 2018-03-23
Explain what happens with skip option

virresh:
Explain what happens with skip option in case multiple commits are pushed together

Ref: https://github.com/travis-ci/travis-ci/issues/8204


plaindocs:
Nice corner case! Thanks.

PR #1778, opened 2018-03-24
Document artifact upload max-size limit

taqtiqa-mark:
Closes travis-ci/travis-ci#9378

### DIFF
--- a/user/uploading-artifacts.md
+++ b/user/uploading-artifacts.md
@@ -11,6 +11,8 @@ Travis CI can automatically upload your build artifacts to Amazon S3, after the
 [`after success`](/user/customizing-the-build/#The-Build-Lifecycle) stage of the
 build.
 
+**NOTE:** There is currently a Travis-CI imposed size limit of 1GB.
+
 For a minimal configuration, add the following to your `.travis.yml`:
 
 ```yaml


user/uploading-artifacts.md
```diff
@@ -11,6 +11,8 @@ Travis CI can automatically upload your build artifacts to Amazon S3, after the
 [`after success`](/user/customizing-the-build/#The-Build-Lifecycle) stage of the
 build.
 
+**NOTE:** There is currently a Travis-CI imposed size limit of 1GB.
```
plaindocs: We should definitely mention this, but that reads a little abrupt. What do you think about changing the first sentence to 

> Travis CI can automatically upload up to 1GB of your build artifacts to Amazon S3, after the after success stage of the build.

taqtiqa-mark: Point taken, it is abrupt.  I just couldn't see what else to add that wouldn't be misleading.  It is not a configurable option in Travis.  It is not a AWS limit.  It is not a side effect of a more generic Travis data limit, etc.

To my mind your proposal suggests the Travis-CI artifacts feature can upload more than 1GB,  just not automatically, e.g. Change a Travis-CI configuration parameter?  But that isn't accurate. Is it?

My experience is you have to install the `awscli` python package, (or something else if your not using a python or generic language setting?), **and** write your own script to upload more than 1GB.
So if you think your artifact may one day edge over 1GB you are better off biting the bullet now, drop the artifact plugin and writing your own script.

I agree that more needs to be said: What you need to do if you have artifacts >1GB.  
But it seems to me, that is a documentation section on its own?

BanzaiMan:
@meatballhat Opinions, please? What is the rationale behind the file size limit on `artifacts`?

taqtiqa-mark:
Happy for this to be closed.  I'm no longer using the plugin.

PR #1779, opened 2018-03-26
Minor fixes for repeated words and subordinate that

GuessWhoSamFoo:
In `user/ssh-known-hosts.md`, the repeated that is grammatically correct but should be rephrased to avoid confusion.


BanzaiMan:
Thank you!

PR #1780, opened 2018-03-27
Explain that services: docker only works on hosted

schultyy:
We're seeing more and more customers running into this where they're trying to use Docker in their builds running on Enterprise. Though for Enterprise the setup is a bit different here.


user/docker.md
```diff
@@ -24,6 +24,8 @@ examples.
 
 > Travis CI automatically routes builds to run on Trusty `sudo: required` when `services: docker` is configured.
 > We do not currently support use of Docker on OS X.
+> This feature is only available on travis-ci.com and travis-ci.org
```
plaindocs: Hey @schultyy 

Great idea to clear this up, but I think we need to change the wording. That to me reads like this feature isn't available, here is how to do it :-p

How about just 

> For information on how to use Docker on Travis CI Enterprise check out [ ] ...

plaindocs:
Two unrelated issues:

* The note has `sudo: required` as code, but we probably want to refer to it as sudo enebled here.
* In the linked Enterprise docker doc, we could cut down and simplify a bit:

Instead of 

 > In order to build other docker images, the Worker needs to be setup to support Docker Builds. To get started, please add the following to /etc/default/travis-worker on all your Workers:

we could have something like 

> To build docker images on Travis CI Enterprise, add the following to /etc/default/travis-worker on all your Workers:

user/docker.md
schultyy: Sounds good - yeah the words I picked weren't really the best here 😅

plaindocs: It is a pretty natural consequence of adding the Enterprise stuff on to the rest of the docs. They weren't written with that in mind, so there is often no natural connection.

plaindocs:
Awesome! Two minor regressions (which I perhaps introduced by being too brief in my comments) inline.

user/docker.md
```diff
@@ -22,9 +22,11 @@ services:
 Then you can add `- docker` commands to your build as shown in the following
 examples.
 
-> Travis CI automatically routes builds to run on Trusty `sudo: required` when `services: docker` is configured.
+> Travis CI automatically routes builds to run on our Trusty sudo-enabled infrastructure when `services: docker` is configured.
 > We do not currently support use of Docker on OS X.
 
+> For information on how to use Docker on Travis CI Enterprise check out [https://docs.travis-ci.com/user/enterprise/build-images/#Enabling-Docker-Builds](https://docs.travis-ci.com/user/enterprise/build-images/#Enabling-Docker-Builds)
```
plaindocs: Especially in the note, just using the URL is quite hard on the eye, I'd suggest

> For information on how to use Docker on Travis CI Enterprise check out [https://docs.travis-ci.com/user/enterprise/build-images/#Enabling-Docker-Builds](Enabling Docker Builds)

user/enterprise/build-images.md
```diff
@@ -65,7 +65,7 @@ you can run:
 
 ### Worker machine configuration
 
-In order to build other docker images, the Worker needs to be setup to support Docker Builds. To get started, please add the following to `/etc/default/travis-worker` on all your Workers:
+To build docker images on Travis CI Enterprise, add the following to /etc/default/travis-worker on all your Workers:
```
plaindocs: And I'd leave the path in code as you had it before `/etc/default/travis-worker`

schultyy: oops - this was definitely not intended 😅 Will fix that right away

PR #1781, opened 2018-03-27
Update PostgreSQL docs to reflect current reality

meatballhat:
by removing the outdated and doomed-to-be-outdated table of which versions are available where, and adding a section about combining `apt` and `postgresql` addons to install and enable arbitrary versions on both container-based and sudo-enabled Linux.

### DIFF
--- a/user/database-setup.md
+++ b/user/database-setup.md
@@ -170,7 +170,10 @@ before_script:
 
 ### Using a different PostgreSQL Version
 
-The Travis CI build environments use version 9.1 by default, but other versions from the official [PostgreSQL APT repository](http://apt.postgresql.org) are also available. To use a version other than the default, specify only the **major.minor** version in your `.travis.yml`:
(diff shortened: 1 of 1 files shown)


user/database-setup.md
```diff
@@ -170,34 +170,33 @@ before_script:
...
+  postgresql: "10"
+  apt:
+    packages:
+    - postgresql-10
+    - postgresql-client-10
+env:
+  global:
+  - PGPORT=5433
+```
+{: data-file=".travis.yml"}
```
plaindocs: Hey @meatballhat - thanks for the PR. :100: on losing the nasty-and-soon-to-be-outdated table.

Do you think it is worth mentioning either of the following:

* you need *both* postrgrsql and the apt packages and they must match (I know I know)
* what PGPORT is :-p (I'm also assuming if you use Postgre you'll know this, but still) (Also, when you say override, is there a default? )

PR #1785, opened 2018-03-30
Fix docker login Docker CE 18.04-rc1+

khs1994:


BanzaiMan:
Are you upgrading Docker on your builds?

khs1994:
Yes,I use latest test docker,current docker 18.04-rc1

BanzaiMan:
Thanks for the PR and the clarification.

Our documentation should reflect the versions of software that we provide on our build images. If you are updating it, it is up to you to adapt to that version.

Having said that, I note that this command is [correct](https://docs.docker.com/v17.09/engine/reference/commandline/login/) for the version that we offer (17.09). We can merge this.

PR #1786, opened 2018-04-01
Update language-specific.md

ithinker143:


ithinker143:
osx_image: xcode9.3bet

Issue #1302, opened 2017-07-07
Document "region" option for opsworks


BanzaiMan:
https://github.com/travis-ci/dpl/blob/v1.8.39/lib/dpl/provider/ops_works.rb#L29 shows that the OpsWorks provider takes the `region` option, but this is not documented in https://docs.travis-ci.com/user/deployment/opsworks (nor https://github.com/travis-ci/dpl/#opsworks)

anttitevanlinna:
I was fighting this issue for 2 hours. Only debugging with AWS cli gave me the insight that the region is really a mandatory parameter in getting deployments working.

plaindocs:
Related to https://github.com/travis-ci/dpl/issues/191

PR #1787, opened 2018-04-01
Document that the OpsWorks provider accepts a region option

billhorsman:
I'm not sure what the default region is but I think it's wise to specify it explicitly. This fixes issue #1302.


billhorsman:
Sorry, I referenced the wrong issue in the commit message. This does in fact fix issue #1302

plaindocs:
Thanks for this @billhorsman 

Looks like the default is set at https://github.com/travis-ci/dpl/blob/73eabc709312d91aba2886b99cea6c105c2f15b7/lib/dpl/provider/ops_works.rb#L15

billhorsman:
Yes, you're right @plaindocs. I looked for that but couldn't find it. I'll update my changes to mention the default (and update from master again).

billhorsman:
OK, ready again for your @plaindocs

user/deployment/opsworks.md
```diff
@@ -29,7 +32,7 @@ You can also have the `travis` tool set up everything for you:
 travis setup opsworks
 ```
 
-Keep in mind that the above command has to run in your project directory, so it can modify the `.travis.yml` for you.
+Keep in mind that the above command has to run in your project directory, so it can modify the `.travis.yml` for you. Note that the `region` isn't generated by running `travis setup opsworks`.
```
plaindocs: Nice idea to fold this into the note about `setup`.

user/deployment/opsworks.md
```diff
@@ -14,9 +14,12 @@ deploy:
   access-key-id: ACCESS-KEY-ID
   secret-access-key: SECRET-ACCESS-KEY
   app-id: APP-ID
+  region: REGION # (default: us-east-1)
 ```
 {: data-file=".travis.yml"}
 
+Strictly, `region` isn't required but it defaults to `us-east-1` so you'll need it if your stack is elsewhere. You will get an error like "OpsWorks service error: Unable to find app with ID xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" if your stack is not in `us-east-1` and you don't define `region`.
```
plaindocs: This is now a little repetitive, how about:

>  region` isn't required but it defaults to `us-east-1`. If your application is located in a different region you will get an error like "OpsWorks service error: Unable to find app with ID xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx".

Or do you think we need to be more clear?

billhorsman: Nice. That reads much better. I've incorporated your suggestion :)

plaindocs:
Thanks again!

PR #1797, opened 2018-04-11
Add patch level version to example code

non-binary:
A customer reached out that they could not use a patch level version of Haskell however they can and always can but the documentation wasn't specific/showed an example of this being done.


cotsog:
Maybe we should change the text as well i.e.

> You can specify one or more GHC versions using major.minor notation. Patch level versions (7.6.2 for example) may change any time:

What do you think?

non-binary:
Edit: I'm going to merge this since it was approved by @plaindocs. We can make improvements to the text in another PR once Dominic has gotten back and we have time to discuss it.

PR #1798, opened 2018-04-12
Add workaround for RHS fns

justien:
@svenfuchs @DrTorte could you take a look?
There is one place where your expertise is needed.
Thanks!

### DIFF
--- a/user/conditional-builds-stages-jobs.md
+++ b/user/conditional-builds-stages-jobs.md
@@ -118,7 +118,7 @@ It is currently not possible to compare function calls. This means that if you t
 
 `env(PRIOR_VERSION) != env(RELEASE_VERSION)`
 
-where PRIOR_VERSION and RELEASE_VERSION are environment variables defined elsewhere in .travis.yml, the condition will be evaluated as true, even when it is false. This sad state of affairs is being addressed in an upcoming feature release.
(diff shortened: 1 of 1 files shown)


user/conditional-builds-stages-jobs.md
```diff
@@ -112,6 +112,29 @@ Known functions are:
...
+For instance, continuing with the above example, the Deploy stage would include:
+
+```
+- stage: deploy
+   if: attribute=value
+   env:
+    - PRIOR_VERSION=$(git describe --abbrev=0 --tags)
+    - RELEASE_VERSION=$(grep SOME MAGIC HERE PLEASE HELP ME TO PUT SOMETHING THAT MAKES SENSE HERE BUT IS ALSO GENERIC)
+   script:
+    - "$PRIOR_VERSION" = "$RELEASE_VERSION" && travis_terminate || echo "Deploying latest version ..."
```
cotsog: I suggest we show an if/then statement instead as it seems more natural to me. e.g.

```
script:
  - if [ $PRIOR_VERSION == $RELEASE_VERSION ]; then travis_terminate 0; fi
  # else
  - echo "Deploying latest version ..."
```

justien: Pls ignore this branch. We are working on it elsewhere.

cotsog: Ah! I was just wondering why it was closed.

Let me know if you want to review what your are doing over there. :)

PR #1806, opened 2018-04-18
add 'travis-worker is not starting' case to ops manual

schultyy:
This PR adds a new case to our operations manual where travis-worker wouldn't start because a directory is missing. This is also the first specific case we add for Ubuntu 16.04.

### DIFF
--- a/user/enterprise/operations-manual.md
+++ b/user/enterprise/operations-manual.md
@@ -107,17 +107,15 @@ The above mentioned error can be caused by a configuration mismatch in [the GitH
 Your Travis CI Enterprise license has a hostname field which contains the hostname of your installation. When the license's hostname does not match the actual hostname, the container does not start. If this is the case or you suspect that this might be likely, please [get in touch with us](mailto:enterprise@travis-ci.com?subject=License%20Hostname%20Change) and we'll help you to get back on track.
 
 
-## travis-worker on 16.04 does not start
+## travis-worker on Ubuntu 16.04 does not start
 
-travis-worker got installed on a fresh installation of Ubuntu 16.04 (Xenial). `sudo systemctl status travis-worker` unveils that it is not running.
-
-### Strategies
-
-#### `/var/tmp/travis-run.d/travis-worker:` No such file or directory
+travis-worker got installed on a fresh installation of Ubuntu 16.04 (Xenial). `sudo systemctl status travis-worker` shows that it is not running.
 
 Either `sudo journalctl -u travis-worker` or `sudo systemctl status travis-worker` report `/usr/local/bin/travis-worker-wrapper: line 20: /var/tmp/travis-run.d/travis-worker: No such file or directory`.
 
-In order to fix this, please create `/var/tmp/travis-run.d/travis-worker` via:
+### Strategy
+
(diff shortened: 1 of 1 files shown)


christophwitzko:
Hello @schultyy, this looks perfect! Thanks for adding it. ☺️

user/enterprise/operations-manual.md
```diff
@@ -106,6 +106,24 @@ The above mentioned error can be caused by a configuration mismatch in [the GitH
 
 Your Travis CI Enterprise license has a hostname field which contains the hostname of your installation. When the license's hostname does not match the actual hostname, the container does not start. If this is the case or you suspect that this might be likely, please [get in touch with us](mailto:enterprise@travis-ci.com?subject=License%20Hostname%20Change) and we'll help you to get back on track.
 
+
+## travis-worker on 16.04 does not start
```
plaindocs: Here the issue is that travis-worker does not start. Then we need to "filter" down to Ubuntu 16.04, so 

> travis-worker does not start on Ubuntu 16.04

user/enterprise/operations-manual.md
```diff
@@ -106,6 +106,24 @@ The above mentioned error can be caused by a configuration mismatch in [the GitH
...
+
+travis-worker got installed on a fresh installation of Ubuntu 16.04 (Xenial). `sudo systemctl status travis-worker` unveils that it is not running.
```
plaindocs: Avoid 'unveils', it is quite a poetic word. You could use reveals, or even shows.

But probably they have realised that it isn't running some other way, because there are problems? And they are using `systemctl` to *confirm* this? or not?

user/enterprise/operations-manual.md
```diff
@@ -106,6 +106,24 @@ The above mentioned error can be caused by a configuration mismatch in [the GitH
...
+
+### Strategies
```
plaindocs: We're only showing one strategy.

user/enterprise/operations-manual.md
```diff
@@ -106,6 +106,24 @@ The above mentioned error can be caused by a configuration mismatch in [the GitH
...
+
+#### `/var/tmp/travis-run.d/travis-worker:` No such file or directory
```
plaindocs: This is another symptom, not a strategy.

user/enterprise/operations-manual.md
```diff
@@ -106,6 +106,24 @@ The above mentioned error can be caused by a configuration mismatch in [the GitH
...
+
+Either `sudo journalctl -u travis-worker` or `sudo systemctl status travis-worker` report `/usr/local/bin/travis-worker-wrapper: line 20: /var/tmp/travis-run.d/travis-worker: No such file or directory`.
+
+In order to fix this, please create `/var/tmp/travis-run.d/travis-worker` via:
```
plaindocs: Avoid "in order to", use "to" instead

plaindocs: Mayve we can describe the problem like

One possible reason that travis-worker isn't running is that systemctyl cannot create a temp file. To fix this please create a writable temp file by running:

user/enterprise/operations-manual.md
schultyy: They're using `systemctl status` to confirm that, correct.

PR #1811, opened 2018-04-26
Some prep for the future

plaindocs:
* [ ] How are we referring to different types of build? Free open source plans? Paid plans? etc

* [x] Build feeds. Token for all repos? Or just token for paid plans?  https://docs.travis-ci.com/user/build-feeds/
    - A token is needed for private repos, no token for public repos. 

* [x] Permissions / github oauth scopes. I know I've chatted abut this, but can't track down *exactly* what scopes we'll need in the new world order.

   You can authorize all repositories or repo by repo: 

   image

    Normally, I'd think one would authorize a few repos in a bunch, not all repositories. In any case, the permissions requested are: 
  - Read access to code
  - Read access to metadata and pull requests
  - Read and write access to administration, checks, commit statuses, and deployments

  These permissions are displayed in GitHub's UI when installing Travis CI: 

  image

* [X] Ruby, caching. Assuming available for all (before, was just .com). 
  
   Yes - caching is available for everyone, both public and private repositories and sponsored / subscribed accounts. 

* [X] Private dependencies.  (as above) - custom SSH Keys in public repositories? 

     No - custom SSH Keys will only be available for private repositories. 

* [X] Speeding up the build. Only 2 concurrent during trial. 
        Trial will add +2 concurrent jobs to either public or private repositories

* [x] Debug mode. Debug Mode will be available by default for private repositories, and activated on a case-by-case basis for public repositories (due to the security concerns). Similar to the current behaviour. 

* [X] Customizing the build [Build Timeouts] 10 / 50 / 120 ? 
        open source gets the same times as .org
        private gets the same

* [X] Developer Portal .com ? 
    - Yes, the .com API will have all the information for all projects living at travis-ci.com. 
    - Non-migrated public repositories, will still use .org for the time being.
 
* [x] Do we still need `--pro` with the cli? 
    - Unfortunately, yes - I'd recommend new people switching the default in the CLI config so that it calls the --pro / travis-ci.com endpoint by default


user/unified-platform.md
```diff
@@ -0,0 +1,41 @@
+---
+title: "Travis CI Single Platform"
+---
+
+On May 2nd, 2018 Travis CI integration open source and private projects onto a unified platform on **travis-ci.com**!
+
+This means you can manage public and private repositories from one domain, allow open source projects the ability to add extra concurrency, and allow our team to maintain and improve a single CI platform.
+
+We are implementing these changes with GitHub Apps, which not only provides a richer integration with GitHub, but also allows for many amazing features we plan to roll out in the next few months.
```
MariadeAnton: I'd remove "but also allows for many amazing features we plan to roll out in the next few months" in the docs here. I'm not sure it brings much value

user/unified-platform.md
```diff
@@ -0,0 +1,41 @@
...
+
+Travis CI will continue to support open source development as we've done in the past.  
+
+If you're an open source developer currently testing your project with Travis CI, your builds will continue to run on travis-ci.org for the moment. We will announce more soon on how to migrate your build history from travis-ci.org to travis-ci.com, but for now your currently active repositories will remain on the .org platform and you won’t notice any change at all.
+
+If you are activating a new repository on Travis CI, please use [travis-ci.com](https://www.travis-ci.com).
+
+#### Migrating to travis-ci.coming
+
+Sometime in May 2018 we'll be migrating repositories from travis-ci.org to travis-ci.coming
```
MariadeAnton: I think this will happen in June? I'd be safer on that side and use June instead cc/ @joshk 

I think autocorrect may have slipped a typo on `travis-ci.coming` instead of `travis-ci.com` used both in the title and the text here, is that possible?

user/unified-platform.md
```diff
@@ -0,0 +1,41 @@
...
+
+You can opt in to migration by manually switching your repositories to travis-ci.com now (we'll email you before the migration happens), or if you don't need to keep your build history and want to switch right now please reach out to [support@travis-ci.com](mailto:support@travis-ci.com?Subject=Hello%20Travis%20CI).
```
MariadeAnton: I'm not 100% sure I get what we mean here - 

I think I'd remove the part about opt-in and talk about how they can migrate themselves once that option is available.  What do you think?

user/deployment/pages.md
```diff
@@ -19,7 +19,7 @@ For a minimal configuration, add the following to your `.travis.yml`:
 deploy:
   provider: pages
   skip-cleanup: true
-  github-token: $GITHUB_TOKEN  # Set in travis-ci.org dashboard, marked secure
+  github-token: $GITHUB_TOKEN  # Set in travis-ci.com dashboard, marked secure
```
MariadeAnton: I think instead of the travis-ci.com dashboard, we should say: 

"Set in the settings page of your repository, marked secure" 

What do you think?

user/repository-providers.md
```diff
@@ -4,8 +4,8 @@ layout: en
 
 ---
 
-Build and test your open source projects hosted on GitHub on [travis-ci.org](https://travis-ci.org/).
+<!-- FIXME: do we need this? -->
 
-Build and test your private repositories hosted on GitHub on [travis-ci.com](https://travis-ci.com/).
+Build and test your open source and private repositories hosted on GitHub on [travis-ci.com](https://travis-ci.com/).
```
MariadeAnton: I think this normally helps with answering the question of which repository providers we support, clarifying what we support and what we don't. 

I think going forward, we can remove the full page, since it's being redirected to the "customizing the build section" https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/customizing-the-build.md 

At: customizing-the-build/#What-repository-providers-or-version-control-systems-can-I-use%3F

MariadeAnton: I've attempted to fix this in the commits below - let me know what you think!

user/deployment/pages.md
plaindocs: Perfect

PR #1815, opened 2018-04-30
confused between two file names given

RudyOnRails:
I'm not 100% but I think the data-file is incorrect. Either way, it is confusing - not sure if it should go in your `config/database.yml` or the `.travis.yml` file. 

https://docs.travis-ci.com/user/database-setup/#SQLite3
screen shot 2018-04-30 at 4 39 27 pm


plaindocs:
Well spotted, and thanks for the fix.

RudyOnRails:
Thanks, and you're welcome!

PR #1828, opened 2018-05-07
Clarify versions supported by travis

jimhester:
Specifying specific values just leads to bit-rot.


plaindocs:
> Specifying specific values just leads to bit-rot. 

Agree, and we're working on that.

Thanks for the PR!

PR #1830, opened 2018-05-09
Add TOC to for-beginners

Lin-Buo-Ren:
Noticed that there's no TOC exist in this section during scraping the index, this patch attempts to add one according to other sections' practices.

Signed-off-by: 林博仁(Buo-ren, Lin) <Buo.Ren.Lin@gmail.com>


BanzaiMan:
Thank you!

PR #1837, opened 2018-05-11
Highlight migration info 

plaindocs:
In the menu and front page.

Point of order - can you use the gem if you have less than 50 repos? Would make sense, no?

Also no mega keen on talking about open source accounts, but :man_shrugging:


plaindocs:
I suggest we merge this in asap - then consolidate into one dir or one file with appropriate redirects and a better explanation.

PR #1841, opened 2018-05-12
Update current OS X and Xcode versions

mjm:
The Objective-C/Swift pages have text that references outdated default versions of OS X and Xcode. This change corrects that text to match the OS X Build Environment page as well as the tables on both pages.


plaindocs:
Thanks!

PR #1845, opened 2018-05-17
Fix CMake version shipped with Ubuntu 14.04 Trusty

Mizux:
cf https://travis-ci.org/Mizux/cmake-cpp/jobs/338920636#L458


plaindocs:
Thanks!

PR #1859, opened 2018-05-28
Add disk space

plaindocs:
Last three lines of the table and the note that follows add the disk space.

### DIFF
--- a/user/reference/overview.md
+++ b/user/reference/overview.md
@@ -34,20 +34,23 @@ An [OS X](/user/reference/osx/) environment for Objective-C and other OS X speci
 
 The following table summarizes the differences across virtual environments and operating systems:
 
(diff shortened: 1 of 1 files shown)


iriberri:
This looks neat, thanks @plaindocs!

I wonder if the following suggestion that we usually share in Support might be helpful. Not sure how much worth it it is, but for those that might be stuck these couple extra GB might help:

> The available disk space will depend on the base image used in your build, that is chosen depending on the `language` you specify for your project. If you need a bit more space in your Ubuntu builds, we recommend using `sudo: required` and `language: minimal`, which will route you to a base image with less tools and languages preinstalled. This image has approximately ~24GB of free space.

Anyway, this is great!! 👍

plaindocs:
Hi @iriberri I'm merging this in - shout if you have any thoughts on the updated info.

iriberri:
Wonderful! Thanks again @plaindocs :)

PR #1860, opened 2018-05-28
Clarify that the `tags` option applies to npm

blgm:
I initially got confused that thought these were git tags.  Hopefully this clarification will point future users in the right direction.


plaindocs:
Thanks for the clarification.

PR #1864, opened 2018-05-29
Adding note about running apt-get update

ryndaniels:
Since we don't run apt-get automatically anymore without using the apt addon, updating the docs to clarify this.

### DIFF
--- a/user/installing-dependencies.md
+++ b/user/installing-dependencies.md
@@ -10,9 +10,24 @@ redirect_from:
 
 ## Installing Packages on Standard Infrastructure
 
-To install Ubuntu packages that are not included in the default [standard](/user/reference/precise/), use apt-get in the `before_install` step of your `.travis.yml`:
+To install Ubuntu packages that are not included in the standard [precise](/user/reference/precise/) or [trusty](/user/reference/trusty/) distribution, use apt-get in the `before_install` step of your `.travis.yml`:
 
-By default, `apt-get update` does not get run automatically unless you are using the [APT addon](#Installing-Packages-with-the-APT-Addon). If you want to update `apt-get` automatically on every build, add something like the following to your `.travis.yml`:
+```yaml
+before_install:
+  - sudo apt-get install -y libxml2-dev
+```
+{: data-file=".travis.yml"}
+
+By default, `apt-get update` does not get run automatically. If you want to update `apt-get` automatically on every build, there are two ways to do this. The first is by running `apt-get update` explicitly in the `before_install` step:
+
+```yaml
+before_install:
+  - sudo apt-get update
+  - sudo apt-get install -y libxml2-dev
+```
+{: data-file=".travis.yml"}
+
+The second way is to use the [APT addon](#installing-packages-with-the-apt-addon):
 
 ```yaml
 before_install:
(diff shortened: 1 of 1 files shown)


user/installing-dependencies.md
```diff
@@ -23,6 +23,17 @@ before_install:
 >
 > Use the `-y` parameter with apt-get to assume yes as the answer to each apt-get prompt.
 
+
+### Updating APT Packages
+
```
plaindocs: In the example before this new content we're running 

```
before_install:
  - sudo apt-get -qq update
  - sudo apt-get install -y libxml2-dev
```

is this to avoid that?

user/installing-dependencies.md
```diff
@@ -23,6 +23,17 @@ before_install:
...
+By default, `apt-get update` does not run automatically unless the [APT addon](#Installing-Packages-with-the-APT-Addon) is used. If you want to opt-in to running `apt-get update`, add the following to your `.travis.yml`:
```
plaindocs: Here I'd opt for something like

> By default, `apt-get update` does not run automatically unless you're using the [APT addon](#Installing-Packages-with-the-APT-Addon).

and maybe 

> If you want to update apt-get automatically (on every build?), add ...

user/installing-dependencies.md
```diff
@@ -12,14 +12,18 @@ redirect_from:
 
 To install Ubuntu packages that are not included in the default [standard](/user/reference/precise/), use apt-get in the `before_install` step of your `.travis.yml`:
 
+By default, `apt-get update` does not get run automatically unless you are using the [APT addon](#Installing-Packages-with-the-APT-Addon). If you want to update `apt-get` automatically on every build, add something like the following to your `.travis.yml`:
```
BanzaiMan: I believe the current logic is: 

"…you are using the [APT addon](#Installing-Packages-with-the-APT-Addon)" _or `apt-get install` is explicitly invoked in `.travis.yml`._

user/installing-dependencies.md
```diff
@@ -12,14 +12,18 @@ redirect_from:
...
+
 ```yaml
 before_install:
-  - sudo apt-get -qq update
   - sudo apt-get install -y libxml2-dev
+addons:
```
BanzaiMan: I think we should split this into two examples.

```yaml
before_install:
  - sudo apt-get install -y libxml2-dev
```

and

```yaml
addons:
  apt:
    update: true
before_install:
  - sudo ./install_meta_dependencies # a custom script which invokes `sudo apt-get install`, but not `sudo apt-get update`
```

Aside: I can't think of a compelling use case of `addons.apt.update: true`; the user can easily add `sudo apt-get update` into the custom script if needed.

user/installing-dependencies.md
```diff
@@ -12,14 +12,18 @@ redirect_from:
 
 To install Ubuntu packages that are not included in the default [standard](/user/reference/precise/), use apt-get in the `before_install` step of your `.travis.yml`:
```
plaindocs: I know this isn't related to your PR, but this would be better as:

> To install Ubuntu packages that are not included in the standard [precise](/user/reference/precise/) or [trusty](/user/reference/trusty/) distribution, use apt-get in the `before_install` step of your `.travis.yml`:

user/installing-dependencies.md
plaindocs: And ~something like~ adds nothing here.

user/installing-dependencies.md
```diff
@@ -69,9 +73,7 @@ before_install:
 
 ### Installing Packages with the APT Addon
 
-You can also use the APT addon.
-
-This addon provides declarative shortcuts to basic operations of the `apt-get` commands.
+You can also use the APT addon to declare additional APT sources and packages that you want installed, without running `apt-get` commands in your `before_install` script.
```
plaindocs: Maybe 

> You can also install packages using the APT addon, without running `apt-get` commands in your `before_install` script.

PR #1869, opened 2018-05-31
Use encrypted AWS Access Key ID

sotayamashita:
According to "[Understanding and Getting Your Security Credentials > Access Keys (Access Key ID and Secret Access Key) ](https://docs.aws.amazon.com/general/latest/gr/aws-sec-cred-types.html#access-keys-and-secret-access-keys)"

> Do not provide your access keys to a third party, even to help find your canonical user ID. By doing this, you might give someone full access to your account. 

So I think we should encrypt **AWS Access Key ID** same as AWS Secret Access Key. What do you think?


plaindocs:
Yeah, looks right. Thanks for pointing this out and fixing.

PR #1878, opened 2018-06-05
Explain how to deal with certificate errors

schultyy:
Even though this is specific to Let's Encrypt this is a valuable tutorial on how a user can manually rotate (faulty) SSL/TLS certificates on their Travis CI Enterprise instance, helping them to possibly rule
out certain sources of error.

### DIFF
--- a/user/enterprise/operations-manual.md
+++ b/user/enterprise/operations-manual.md
@@ -126,9 +126,34 @@ Either `sudo journalctl -u travis-worker` or `sudo systemctl status travis-worke
 
 One possible reason that travis-worker is not running is that `systemctl` cannot create a temporary directory for environment files. To fix this, please create the directory `/var/tmp/travis-run.d/travis-worker` and assign write permissions via:
 
-```sh
+```bash
 $ mkdir -p /var/tmp/travis-run.d/
 $ chown -R travis:travis /var/tmp/travis-run.d/
+```
+
+## Travis CI Enterprise becomes unavailable do to a certificate error
+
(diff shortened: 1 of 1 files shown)


user/enterprise/operations-manual.md
```diff
@@ -126,11 +126,36 @@ Either `sudo journalctl -u travis-worker` or `sudo systemctl status travis-worke
...
 
-```sh
+```bash
 $ mkdir -p /var/tmp/travis-run.d/
 $ chown -R travis:travis /var/tmp/travis-run.d/
 ```
 
+## Travis CI Enterprise becomes unavailable do to a certificate error
+
+Travis CI Enterprise is using a Let's Encrypt certificate which has expired. A (forced) renewal via the following commands don't have any effect:
```
MariadeAnton: are all Enterprise installations using certificates from Let's Encrypt? If not, we can make it generic and then use Let's Encrypt as an example. If yes, all good here! 

Could you elaborate on the forced renewal not having any effect? Is it that those are the usual commands to fix it but they may not work?

acnagy: +1 for making more generic, or turning this to a blog post follow up :) I think it's likely that Let's Encrypt is a minority for enterprise customers so our docs referencing very directly. 

Or, we could split the Let's Encrypt sections in the docs out to their own doc? wdyt?

plaindocs:
Hey @schultyy what is the status of this? Do you still need a review from me, or are you planning to generalize it according @acnagy 's suggestion?

plaindocs:
Closing, please reopen if still relevant.

PR #1881, opened 2018-06-08
Reconfigure the Customizing your Build menu

plaindocs:
Note: build will fail this is fine. This is just a place to talk about the reorg, I still need to move all the content around.


_includes/sidebar.html
```diff
@@ -43,6 +42,49 @@ <h3>Getting Started</h3>
...
+      <ul>
+
+<li><a href="#Installing-a-Second-Programming-language">Installing a Second Programming language</a></li>
+<li><a href="#Build-Timeouts">Build Timeouts</a></li>
+<li><a href="#Limiting-Concurrent-Jobs">Limiting Concurrent Jobs</a></li>
+<li><a href="/user/speeding-up-the-build/">Speeding up the Build</a></li>
+
+
+<!-- group these? -->
+<li><a href="#Building-only-the-latest-commit">Building only the latest commit</a></li>
```
MariadeAnton: I think Building only the latest commit groups more with speeding things up and Travis config details, above - rather than git config related. What do you think? Given that mind-mapping depends on folks, I'm super curious to hear your thoughts :)

_includes/sidebar.html
```diff
@@ -43,6 +42,49 @@ <h3>Getting Started</h3>
...
+<li><a href="#Git-Submodules">Git Submodules</a></li>
+<li><a href="#Git-LFS">Git LFS</a></li>
+<li><a href="#Git-Sparse-Checkout">Git Sparse Checkout</a></li>
+
+<!-- group these? Build *flow* or something? -->
+<li><a href="#Building-Specific-Branches">Building Specific Branches</a></li>
+<li><a href="#Skipping-a-build">Skipping a build</a></li>
+<li><a href="#Implementing-Complex-Build-Steps">Implementing Complex Build Steps</a></li>
+
+<li><a href="#Custom-Hostnames">Custom Hostnames</a></li>
```
MariadeAnton: these 3: custom hostnames, repo providers and what yaml version - sound like very specific questions, would it make sense to group them as or include as a reference?

_includes/sidebar.html
plaindocs: :100: the Git is 'fake' grouping really - as in it is a bucket where people will find stuff. You're unlikely to look for git commit depth, say, and also go OH, here is another git thing I'm interested in

_includes/sidebar.html
plaindocs: Yeah, this kinda stuff I'd like to move out of the "prime real estate" for sure.

user/build-lifecycle.md
```diff
@@ -0,0 +1,190 @@
...
+layout: en
+---
+
+Travis CI provides a default build environment and a default set of phases for each programming language. You can customize any phase in this process in your `.travis.yml`.
+
+<div id="toc"></div>
+
+## The Build Lifecycle
+
+A build or job on Travis CI is made up of two main parts:
```
iriberri: Should we use "job" here, instead of "build or job"? I think it might help clarify the concept that, at the end, what is running these commands is a job.

I might be missing background on some previous decisions/discussions/conventions, so please excuse me if this is on purpose :) I think this PR might be also about "moving things around" so apologies if this is a too detailed comment!

user/common-build-problems.md
```diff
@@ -489,7 +489,7 @@ Most of our internal build commands are wrapped with `travis_retry` to reduce th
 impact of network timeouts.
 
 Note that `travis_retry` does not work in the `deploy` step of the build, although it
-does work in the [other steps](/user/customizing-the-build/#The-Build-Lifecycle).
+does work in the [other steps](/user/build-lifecycle/).
```
iriberri: Should we (or are we) redirect the old `/user/customizing-the-build/` to a similar page, such as the new `/user/build-lifecycle/`? I was thinking about external links/tutorials that might be pointing to the old section

user/build-lifecycle.md
plaindocs: This is great, and a valid point!

user/common-build-problems.md
plaindocs: Definitely.

PR #1884, opened 2018-06-11
Document conditions v1, split out a separate pages for the versioned conditions specs

svenfuchs:
This is in preparation of officially shipping https://github.com/travis-ci/travis-conditions/pull/1. This is currently in dark release, and we'll blog about it soon.

This PR splits out two version specific pages for conditions `v0` and `v1`. I've kept the structure and wording on `v0`, but changed things a little for `v1` (mostly since the "left hand" vs "right hand" side distinction does not apply any more, the parser has been much improved).


user/conditional-builds-stages-jobs.md
```diff
@@ -33,19 +27,9 @@ stages:
...
-    if: tag =~ ^v1
-
-stages:
-  - name: deploy
-    # require the event type to be either `push` or `pull_request`
-    if: type IN (push, pull_request)
 ```
 
-At the moment, stages that are found to be excluded will be skipped silently (an improvement to this is on the roadmap, giving more explicit feedback on filtering).
+At the moment, stages that are found to be excluded will be skipped silently.
```
plaindocs: Above, in  line 23:

> Stages can be included or excluded by specifying a condition in the `stages` section:

You can be more direct here:

> Build only specific stages by including them in your .travis.yml

user/conditions-v1.md
```diff
@@ -0,0 +1,320 @@
+---
+title: Conditions
+layout: en
+---
+
+> This page documents the behavior of conditions `v1`, which is currently
+> available using `conditions: v1` in your `.travis.yml` file. This new
```
plaindocs: This whole note isn't very easy to parse. What do you think of something like 

> This page documents conditions v1, which you can use by setting `conditions: v1` in your `.travis.yml`. v1  replaces [`v0`](/user/conditions-v0) which is being deprecated.

user/conditions-v1.md
```diff
@@ -0,0 +1,320 @@
...
+> be found [here](/user/conditions-v0).
+
+Conditions can be used to filter out, and reject builds, stages, and jobs by
+specifying conditions in your build configuration (your `.travis.yml` file).
+See [Conditional Builds, Stages, and Jobs](/user/conditional-builds-stages-jobs)
+for details.
+
+## Examples
+
+This condition requires the branch name to be master (note for PRs this is the
```
plaindocs: This one is also really hard to parse. 

I think it would be useful to start off earlier explaining that by including a condition we're automtically exlcuding others? Ie, with branches. (not related to v0 and v1)

plaindocs: Maybe it would be more clear to describe the condition as 

> Build only when on the master branch.

user/conditions-v1.md
```diff
@@ -0,0 +1,320 @@
...
+base branch name):
+
+```
+branch = master
+```
+
+This requires the tag name to match a regular expression (enclose in slashes
+for more complicated expressions):
```
plaindocs: And likewise

Build only tags that match the specificed regular expression. 

Here my preference would be to only document the `//` format :-p and not have to explain the difference

user/conditions-v1.md
```diff
@@ -0,0 +1,320 @@
...
+
+```
+tag =~ ^v1
+tag =~ /^(v1|v2)/
+```
+
+This requires the event type to be either `push` or `pull_request`:
```
plaindocs: By event type we really mean build type, right?

user/conditions-v1.md
```diff
@@ -0,0 +1,320 @@
...
+type IN (push, pull_request)
+```
+
+This requires the branch name to not be one of several names:
+
+```
+branch NOT IN (master, dev)
+```
+
+This require the sender login name to match a given name (use quotes for
```
plaindocs: Sender = commit author?

user/conditions-v1.md
```diff
@@ -0,0 +1,320 @@
...
+> Note that this means conditions do not have access to the build environment,
+> and they are **not** evaluted in Bash. Bash variables or subprocesses can **not**
+> be evaluated. Variable names and unquoted strings starting with a dollar char
+> `$` raise a parse error, causing the build request to be rejected. Quoted
+> strings still can start with a dollar char, so if you definitely need a string
+> to start with a dollar char you can enclose it in quotes.
+
+## Specification
+
+The following expressions are parsed and evaluated as expected.
```
plaindocs: > as expected 

probably isn't helpful here.

user/conditional-builds-stages-jobs.md
```diff
@@ -57,121 +41,12 @@ jobs:
...
-  include:
-    - # require the event type to be either `push` or `pull_request`
-      if: type IN (push, pull_request)
-      env: FOO=foo
 ```
 
 Jobs need to be listed explicitely, i.e., using `jobs.include` (or its alias `matrix.include`), in order to specify conditions for them. Jobs created via [matrix expansion](/user/customizing-the-build/#Build-Matrix) currently cannot have conditions.
 
-At the moment, jobs that are found to be excluded will be skipped silently (an improvement to this is on the roadmap).
+At the moment, jobs that are found to be excluded will be skipped silently.
```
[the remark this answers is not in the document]
justien: explicitely should be explicitly

user/conditions-v0.md
```diff
@@ -0,0 +1,124 @@
+---
+title: Conditions (v0, deprecated)
```
acnagy: > Conditions v0 (Deprecated)

That seems clearer to me?

user/conditions-v0.md
```diff
@@ -0,0 +1,124 @@
...
+layout: en
+---
+
+> This page documents the behavior of conditions `v0` (beta), which is
```
acnagy: I think we can drop `(beta)` here, as it's already deprecated?

user/conditions-v0.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Right hand side
+
+It is currently not possible to compare function calls. This means that if you
+try to evaluate something similar to:
+
+`env(PRIOR_VERSION) != env(RELEASE_VERSION)`
+
+where `PRIOR_VERSION` and `RELEASE_VERSION` are environment variables defined
+elsewhere in .travis.yml, the condition will be evaluated as true, even when it
```
acnagy: > the condition will always be evaluated as true, even..

I think that helps a bit for emphasis?

user/conditions-v0.md
```diff
@@ -0,0 +1,124 @@
...
+is false.
+
+A work-around is to create a script which expresses the conditional test in
+reverse, and applies `travis_terminate` when the condition is false. This
```
acnagy: hm, can we clarify/define the purpose of `travis_terminate` before this example/workaround? Otherwise, this is  🆒

user/conditions-v1.md
acnagy: this is true iiuc, though I think it should be defined here

user/conditional-builds-stages-jobs.md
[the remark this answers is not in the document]
svenfuchs: @plaindocs i've remove `at the moment` and kept the rest. this has nothing to do with the travis-conditions version, but with the fact that Gatekeeper doesn't have a great way of communicating such messages to the user ... so it skips the stage or job, but won't be able to tell.

user/conditional-builds-stages-jobs.md
```diff
@@ -14,12 +14,6 @@ Builds can be included or excluded by specifying a condition as follows:
...
 if: branch = master
-
-# require the tag name to match a regular expression
-if: tag =~ ^v1
-
-# require the event type to be either `push` or `pull_request`
-if: type IN (push, pull_request)
 ```
 
 Build requests that are found to be excluded will not generate a build, but will be listed on the "Requests" tab.
```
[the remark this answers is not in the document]
svenfuchs: Used `do not match the condition` for now, and used the same for stages, and jobs.

user/conditional-builds-stages-jobs.md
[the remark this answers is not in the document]
svenfuchs: @plaindocs removed `at the moment` (as above)
@justien done, thanks!

user/conditions-v1.md
```diff
@@ -0,0 +1,276 @@
+---
+title: Conditions
+layout: en
+---
+
+> This page documents the behavior of conditions `v1`, which is currently
```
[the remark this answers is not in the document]
svenfuchs: I've removed `currently` in both of these (on `v0` and `v1`) ... we are still discussing how this rollout will happen, but IF we are going with a conservative rollout plan then we'll have a "step one" and "step two" phase ... and the default version will change then. so we will have to revisit this banner, and probably remove it on `v1` in the second phase. 

does that address the concern about aging well?

user/conditional-builds-stages-jobs.md
svenfuchs: @plaindocs I feel that doesn't express the same thing though? I feel we'd want to say something like `add a condition` ...

how about

> Stages can be filtered out by specifying a condition in the `stages` section:

?

i'm gonna use that for now. lemme know if it needs improvement

user/conditions-v1.md
svenfuchs: sounds good! i've kept the link to the `v0` docs though?

> This page documents conditions `v1`, which you can use by setting conditions:
> `v1` in your `.travis.yml`. `v1` replaces `v0` which is being deprecated.
> The documentation for `v0` can be found [here](/user/conditions-v0).

user/conditions-v1.md
svenfuchs: 👍 

I've now made it:

> Build only when the tag name matches the given regular expression:

... and removed the format without slashes

user/conditions-v1.md
svenfuchs: jup. we talk about event types internally, but i've changed it to build type now.

user/conditions-v1.md
svenfuchs: Nope, the sender is the person that sends the build request. This can be: 

* the person that does `git push`, 
* the person that opens a pull request, 
* the person that sends an API build request
* the person that creates a cron job (i believe, not 100% sure)

I'm not sure where that should be best documented. It's not currently on API v3 but i believe it might be added soon.

user/conditions-v1.md
svenfuchs: really? hmm, ok ... i've removed it.

user/conditions-v0.md
svenfuchs: done!

user/conditions-v0.md
svenfuchs: done!

user/conditions-v0.md
svenfuchs: @acnagy all of the text on `conditions-v0` is what is currently in production on https://docs.travis-ci.com/user/conditional-builds-stages-jobs/ ... i've just split this out to a separate page so we can have two version specific pages.

i'd rather not change this (and i wouldn't feel comfortable changing the workaround text as that was added by someone else later?)

user/conditions-v0.md
svenfuchs: see above :)

user/conditions-v1.md
svenfuchs: Maybe i can add a footnote? do we have such things in the docs?

svenfuchs:
@plaindocs i've changed it to `conditions syntax`

@plaindocs @acnagy wanna give it another review?

plaindocs:
Will do

user/conditions-v1.md
svenfuchs: for crons `sender` can actually be a random person with permissions on the repo, or the repo owner (which in turn can be an `org` ... so this actually looks a little buggy)

user/conditional-builds-stages-jobs.md
```diff
@@ -9,169 +9,44 @@ You can filter out, and reject builds, stages, and jobs by specifying conditions
 
 ### Conditional Builds
 
-Builds can be included or excluded by specifying a condition as follows:
+Builds can be rejected by specifying a condition as follows:
```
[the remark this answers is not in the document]
svenfuchs: sounds good 👍 

(although that condition *does* reject builds, namely every build that is not on the master branch?)

user/conditional-builds-stages-jobs.md
```diff
@@ -9,169 +9,44 @@ You can filter out, and reject builds, stages, and jobs by specifying conditions
...
-if: type IN (push, pull_request)
 ```
 
-Build requests that are found to be excluded will not generate a build, but will be listed on the "Requests" tab.
+Build requests that do not match the condition will not generate a build, but will be listed on the Requests tab.
 
 ### Conditional Stages
 
-Stages can be included or excluded by specifying a condition in the `stages` section:
+Stages can be filtered out by specifying a condition in the `stages` section:
```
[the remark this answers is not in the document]
svenfuchs: > You can also use stages as part of the condition to trigger a build:

I don't really understand? A stage is not part of a condition ... and it does not trigger a build?

How about

> Configure Travis CI to only include stages when certain conditions are met, such as only including the stage on the master branch. Stages that do not match the given condition are silently skipped.

user/conditional-builds-stages-jobs.md
```diff
@@ -9,169 +9,44 @@ You can filter out, and reject builds, stages, and jobs by specifying conditions
...
-    if: type IN (push, pull_request)
 ```
 
-At the moment, stages that are found to be excluded will be skipped silently (an improvement to this is on the roadmap, giving more explicit feedback on filtering).
+Stages that do not match the condition will be skipped silently.
 
 ### Conditional Jobs
 
-Jobs can be included or excluded by specifying a condition on `jobs.include`:
+Jobs can be filtered out by specifying a condition on `jobs.include`:
```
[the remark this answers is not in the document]
svenfuchs: How about

> Configure Travis CI to only include jobs when certain conditions are met, such as only including a job on the master branch. Jobs that do not match the given condition are silently skipped.

user/conditional-builds-stages-jobs.md
```diff
@@ -9,169 +9,44 @@ You can filter out, and reject builds, stages, and jobs by specifying conditions
...
-
-jobs:
-  include:
-    - # require the event type to be either `push` or `pull_request`
-      if: type IN (push, pull_request)
-      env: FOO=foo
 ```
 
-Jobs need to be listed explicitely, i.e., using `jobs.include` (or its alias `matrix.include`), in order to specify conditions for them. Jobs created via [matrix expansion](/user/customizing-the-build/#Build-Matrix) currently cannot have conditions.
+Jobs need to be listed explicitly, i.e., using `jobs.include` (or its alias `matrix.include`), in order to specify conditions for them. Jobs created via [matrix expansion](/user/customizing-the-build/#Build-Matrix) currently cannot have conditions.
```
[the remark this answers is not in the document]
svenfuchs: Again, the job is not part of the condition? and if given on the job it does not apply to the build, but the job only? 

How about

> In order to apply a condition to a job include it explicitly using `jobs.include` or `matrix.include`. Jobs created via matrix expansion ~currently~ cannot be used to define conditions.

user/conditions-v0.md
```diff
@@ -0,0 +1,123 @@
+---
+title: Conditions v0 (deprecated)
+layout: en
+---
+
+> This page documents conditions v0, which is deprecated. Please opt in to the
```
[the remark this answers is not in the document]
svenfuchs: 👍

plaindocs:
Sven will you be making final changes to the PR based on the outstanding comments / slack discussion?

user/conditions-v0.md
```diff
@@ -4,7 +4,7 @@ layout: en
 ---
 
 > This page documents conditions v0, which is deprecated. Please opt in to the
-> new version v1 by adding `conditions: v1` to your `.travis.yml` file. The
+> new version v1 by adding conditions: v1 to your .travis.yml file. The
```
tooomm: Hmm, why did you remove the highlighting here?

plaindocs: Accidentally. Fixed.

user/conditions-v0.md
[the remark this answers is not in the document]
acnagy: I agree with @plaindocs' change 👍

user/conditions-v0.md
acnagy: Hm, I think I wasn't clear with my comment - re: the workaround text - I was more just picking at defining a "work-around" in the official docs. If it's what we recommend in this case, it's what we recommend in this case. Instead of this:
> A work-around is to create a script which expresses the conditional test in reverse, ..

I'd recommend this:
> The best way to handle the behavior that comparison of function calls always evaluates to `true` is to create a script which expresses the conditional test in reverse, ...

acnagy: this all is to say, I don't particularly want to change the body of the text, just change the use of the word "work-around", to nod instead to the fact that this is how the official docs recommend handling this potential gotcha

acnagy: regarding where the doc came from (split from v0, and edited) -- good to know! that makes sense 👍

user/conditions-v1.md
acnagy: I agree with @plaindocs I think this wording works well:
> ## Examples
> 
> This condition defines a job which only builds for the master branch.

user/conditions-v0.md
acnagy: right! so, re: `travis_terminate` - I wanted to be ultra clear on what it does. From what I can see, `travis_terminate` isn't documented anywhere in the `Deploys` docs and it only turns up for on the Conditionals page in `docs`, and in the code in `travis-build`(I searched via codesearch).

Given that we're referencing it in this example, though, I think we should clarify this: 
> and applies `travis_terminate` when the condition is false.
to:
> and applies `travis_terminate` when the condition is false. (`travis_terminate` is a key word which fails the build)

user/conditions-v1.md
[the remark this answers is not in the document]
acnagy: We have updates to the roll-out - so, we're gonna stick to the opt-in, then a (v v short bit) later move to v1 GA (i.e. everyone is gonna be on the new version). So, sounds good to remove `currently`, and agreed - we'll need to come back to the banner

user/conditions-v1.md
acnagy: LGTM

user/conditions-v1.md
acnagy: Ah, right - yeah, I was thinking of non-cron, non-automated builds

what if we kept the text the same and instead of `sender` used `build creator`?

user/conditions-v0.md
acnagy: As we discussed, this can be it's own topic: https://github.com/travis-ci/docs-travis-ci-com/issues/1899

user/conditions-v0.md
acnagy: subbed in my replacement for the word for "work-around", upgrading it to "the best way". Separately, the issue for `travis_terminate` is here: https://github.com/travis-ci/docs-travis-ci-com/issues/1899

PR #1885, opened 2018-06-11
Code styles readability

svenfuchs:
The current `code` styles make text rather hard to read, especially when used a lot in a single sentence or lists. E.g.:

image

This PR is meant to dial this down a little. I generally feel a more "flat" style feels a little more clean and modern, too:

image

It also removes the odd curved underline from the format tag on the right side:

image

Now:

image

It also removes the border and white background color from code tags in `blockquote` tags (which have a blue background). This looks especially hardcore when there are lots of code tags inside a blockquote. E.g.:

image

Actual example from current production:

image

Now:

image

### DIFF
--- a/assets/stylesheets/_base.sass
+++ b/assets/stylesheets/_base.sass
@@ -11,8 +11,6 @@ p
   @include linkStyle
 
 %code-styles
-  border: 1px solid $cement-grey
-  border-radius: 4px
   font-family: monaco, monospace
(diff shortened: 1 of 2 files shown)


plaindocs:
Also seems fine to me. :+1: :+1:

svenfuchs:
Thank you @ctrlaltjustine @plaindocs! 

😻

PR #1893, opened 2018-06-17
Corrected order of install steps in example

venthur:
You should first `pip install -r requirements.txt` and then `pip install .` and not the other way round. The reason is, that the `requrements.txt` carries the dependencies with pinned versions and the `setup.py` only the dependencies (and conflicting version dependencies if necessary).


plaindocs:
Thanks for the fix.

PR #1896, opened 2018-06-19
remove docs for "Build only if .travis.yml is present" 

MariadeAnton:
This setting is the default behaviour and is non-editable


plaindocs:
LGTM

PR #1902, opened 2018-06-21
document concat in conditions

svenfuchs:


user/conditions-v1.md
```diff
@@ -190,22 +190,42 @@ branch =~ /(master|foo)/
...
+The function `env` returns the value of given environment variable:
 
 ```
 env(FOO)
 ```
 
-The function `env` supports environment variables that are given in your build
-configuration (e.g. on `env` or `env.global`), and environment variables
-specified in your repository settings.
+`env` supports environment variables that are given in your build configuration
```
plaindocs: Maybe say

> `env` supports environment variables that are set in the `env` key of your `.travis.yml` or specified in your repository settings. 

(as env.global is part of env I didn't include it, butfeeel free to include it if that seems important)

svenfuchs: ok 👍 

(yes, i think it's relevant, added it in parentheses)

PR #1904, opened 2018-06-22
fix: install MySQL 5.7 link

akagane99:


plaindocs:
Hi @akagane99 

Thanks for the PR.

I'm changing the toc / heading id system, and switched the branch to make that clear.

PR #1916, opened 2018-06-29
Add missing "toc" blocks for the docs in "Getting Started"

MathExpert:
A bit unsure about the TOC placement, because some of the documents have a bit of text preceding the TOC (like in "Getting Started"), and some of them don't (like in "Core Concepts for Beginners").


plaindocs:
Nice one! 

Yeah, we should probably be more consistent about that.

PR #1918, opened 2018-06-30
Update OpenSSL warning to reflect Python 3.7

cclauss:
Now that [Python 3.7 has shipped](https://docs.python.org/3/whatsnew/3.7.html), this warning not only covers development branches but also production branches of CPython.


plaindocs:
Thanks

PR #1821, opened 2018-05-03
Doc update: Endpoint change when debugging builds on public repositories

mmoayyed:
Sending API requests to the .com endpoint causes `access-denied` errors for public repositories.

PR #1922, opened 2018-07-04
Add note about using --pro flag for public repositories on travis-ci.com

non-binary:
I directed a customer to this document who was having a hard time encrypting their Slack API key for a private repository but they were running on travis-ci.com so the docs were outdated/misleaidng for them. This pull request adds a note about using the `--pro` flag for public repositories when on travis-ci.com/using the GitHub Apps integration.


cotsog:
I'm wondering if this is the right way to inform customers about this subtlety. For example, nowhere in this page we mention the other option i.e. `--org`. Or to use `--pro` for private projects.

Thoughts?

non-binary:
Isn't `--org` deprecated? Since it's the default option when running these commands. Or are you saying that we should specify that they should use `--pro` if on travis-ci.com irregardless of if their project is public or private/

cotsog:
My understanding is that `--org` isn't deprecated yet for as long as we have two sites i.e. travis-ci.org and travis-ci.com. It should be the case until GitHub deprecates service hooks which is [scheduled for January 31st, 2019](https://developer.github.com/v3/guides/replacing-github-services/#deprecation-timeline).

`--org` might be the default when it's the first time you use the CLI for a repo but otherwise, the endpoint will be cached locally which can cause troubles when changing the repo's type (public vs private)

What I'm saying is that I believe it's not giving the full picture of the possible endpoints. Maybe the note should be to make sure to use the right endpoint and link to somewhere else in the docs that would explain the 3 current options i.e.

1. public repo on .org => `--org`
1. public repo on .com => `--pro`
1. private repo on .com => `--pro`

non-binary:
Okay, yeah - this makes more sense. Do we have a page for the Travis CLI where I could add a section for this and then link to it on each deployment section to check which flag you should use.

cotsog:
The closest I can think of are:

* https://docs.travis-ci.com/user/encryption-keys
* https://docs.travis-ci.com/user/environment-variables#Defining-encrypted-variables-in-.travis.yml

non-binary:
I'm wondering if we should create a new page that goes into some specifics of installing the gem and how to use it like a general guide for this situation that solution however that's out of scope for this PR. So maybe we do that first and come back to it.

Also I think "org" and "pro" are not great names for that flag and generally. "pro" would ideally be "com" instead. Is that a feasible change to make?

plaindocs:
Also related to #1821

plaindocs:
Thanks for raising this @non-binary , and thanks all for the discussion everyone. 

Pending making `--com` an alias of `--pro` I think #1961 can replace this PR.

PR #1924, opened 2018-07-09
Move this out of the sidebar

plaindocs:
Just decluttering a little.  FAO @MariadeAnton anything else to add?

Do we need to sync the footer with everywhere else? @paulthegeek


paulthegeek:
@plaindocs I'm not aware of anywhere else. Those links are both in the footer of `.com` and `.org` already.

plaindocs:
I added one that was in the side bar, and not in the footer though. :man_shrugging:

paulthegeek:
@plaindocs Ah I see. I can add this one to the footer in `Travis Web`.

PR #1927, opened 2018-07-12
Updated Company name and CHEF_VERSION(s)

jjasghar:
Opscode was dropped from the company name years ago, and the releases
are significantly higher then "0.y.z" now.

Signed-off-by: JJ Asghar <jj@chef.io>


plaindocs:
Thanks

PR #1909, opened 2018-06-27
Rip out the old JS TOC generator and use liquid instead

plaindocs:
Inspired by #1904  and #1883 but this is something I've wanted to do forever!

Uses https://github.com/allejo/jekyll-toc

This way the link checker will use the same TOC links as people viewing the site, and so we'll all have a better experience.

- [x] Going to need a little styling work, and some double checking of pages that don't need a TOC.

Current menu style:

https://docs.travis-ci.com/user/customizing-the-build/

As it looks with this PR:

https://deploy-preview-1909--musing-lamport-7fb66c.netlify.com/user/customizing-the-build/

Or with an extra level of nested UL

```
            <div id="toc" class="toc">
              <ul class="list-languages">
  <li><a href="#what-this-guide-covers">What This Guide Covers</a></li>
  <li><a href="#using-sauce-labs">Using Sauce Labs</a></li>
  <li><a href="#using-xvfb-to-run-tests-that-require-a-gui">Using xvfb to Run Tests That Require a GUI</a>
    <ul>
      <li><a href="#using-the-xvfb-run-wrapper">Using the xvfb-run wrapper</a></li>
      <li><a href="#using-xvfb-directly">Using xvfb directly</a></li>
      <li><a href="#starting-a-web-server">Starting a Web Server</a></li>
      <li><a href="#using-the-chrome-addon-in-the-headless-mode">Using the Chrome addon in the headless mode</a></li>
      <li><a href="#using-the-firefox-addon-in-headless-mode">Using the Firefox addon in headless mode</a></li>
    </ul>
  </li>
  <li><a href="#using-phantomjs">Using PhantomJS</a></li>
  <li><a href="#examples">Examples</a>
    <ul>
      <li><a href="#real-world-projects">Real World Projects</a></li>
      <li><a href="#ruby">Ruby</a></li>
    </ul>
  </li>
  <li><a href="#troubleshooting">Troubleshooting</a>
    <ul>
      <li><a href="#selenium-and-firefox-popups">Selenium and Firefox popups</a></li>
      <li><a href="#karma-and-firefox-inactivity-timeouts">Karma and Firefox inactivity timeouts</a></li>
    </ul>
  </li>
</ul>

            </div>
```

PR #1932, opened 2018-07-14
Fix broken anchor link

noplanman:


plaindocs:
Thanks for the PR @noplanman - I'm working on fixing up the anchors so they don't use javascript in #1909 .

In either case, special characters are stripped from the anchors.

PR #1934, opened 2018-07-16
Update Stages WIP

plaindocs:


user/build-stages.md
```diff
@@ -97,6 +97,25 @@ jobs:
...
+
+You can also name specific jobs within build stages. We recommend unique job names, but 
+do not enforce it (though this may change in the future). Job names are defined like other 
+job configuration elements within a build stage, but the value of the `name` must be in quotes.
+Here's an example of a build using build stages and job names:
+
+```yaml
+  include:
+    - stage: test
+      name: "test 1"
```
plaindocs: Shall we call this Test Job 1.... for extra clarity?

plaindocs:
Looks great to me @acnagy !

user/build-stages.md
```diff
@@ -97,6 +97,25 @@ jobs:
...
 {: data-file=".travis.yml"}
 
+### Naming your Jobs within Build Stages
+
+You can also name specific jobs within build stages. We recommend unique job names, but 
+do not enforce it (though this may change in the future). Job names are defined like other 
+job configuration elements within a build stage, but the value of the `name` must be in quotes.
+Here's an example of a build using build stages and job names:
+
+```yaml
```
svenfuchs: this should probably include the top-level key `jobs` (or `matrix`) shouldn't it?

user/build-stages.md
```diff
@@ -97,6 +97,25 @@ jobs:
 ```
 {: data-file=".travis.yml"}
 
+### Naming your Jobs within Build Stages
+
+You can also name specific jobs within build stages. We recommend unique job names, but 
+do not enforce it (though this may change in the future). Job names are defined like other 
+job configuration elements within a build stage, but the value of the `name` must be in quotes.
```
svenfuchs: > Job names are defined ... within a build stage.

They're not really defined within a stage, but on the job, aren't they? 

> the value of the name must be in quotes

That's only true if the name has spaces, and ... as this is YAML ... there technically would be other ways to signify it as a string.

How about:

> Jobs defined in the `jobs.include` section can be given a name attribute as follows:

I'm also wondering if, after the example, we should mention that:

> Jobs coming out of the matrix expansion cannot be named.

user/build-stages.md
```diff
@@ -97,6 +97,25 @@ jobs:
 ```
 {: data-file=".travis.yml"}
 
+### Naming your Jobs within Build Stages
+
+You can also name specific jobs within build stages. We recommend unique job names, but 
+do not enforce it (though this may change in the future). Job names are defined like other
```
svenfuchs: > though this may change in the future

I was once told that time references in docs do not age well :) Not sure.

user/build-stages.md
svenfuchs: How about `Unit tests` and `Integration tests`, as more meaningful example names?

user/customizing-the-build.md
```diff
@@ -455,6 +455,29 @@ matrix:
 
 > All build matrixes are currently limited to a maximum of **200 jobs** for both private and public repositories. If you are on an open-source plan, please remember that Travis CI provides this service free of charge to the community. So please only specify the matrix you *actually need*.
 
+### Naming Jobs within Matrices
+
+You can define names for specific jobs within a matrix. We recommend unique job names, but 
+do not enforce it (though this may change in the future). Job names are defined like other 
+job configuration elements within a build matrix, but the value of the `name` must be in quotes.
```
svenfuchs: Same as above about YAML syntax

user/build-stages.md
plaindocs: +1

user/build-stages.md
plaindocs: +1

user/build-stages.md
plaindocs: ;-) sounds sensible -- must have been said by a very wise person. Although here we're really saying "don't rely on this", not "we'll fix this soon".

user/build-stages.md
acnagy: thanks! yeah - I gave them some better names and changed the section from `Test` to `Tests` to show, well, multiple tests happen here

user/build-stages.md
acnagy: whoops! yes - thanks

user/build-stages.md
acnagy: ah, thanks for catching the unclarity here... I knew what _I_ was thinking...

user/build-stages.md
acnagy: yeah, I think we need to keep this for now, but I'll set a reminder for myself to come back to this and remove it if needed

user/build-stages.md
```diff
@@ -97,6 +97,25 @@ jobs:
...
+
+You can also name specific jobs within build stages. We recommend unique job names, but 
+do not enforce it (though this may change in the future). Jobs defined in the `jobs.include`
+section can be given a name attribute as follows: 
+
+```yaml
+jobs:  
+  include:
+    - stage: "Tests"
+      name: "Unit Tests"
```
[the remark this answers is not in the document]
acnagy: ahhh.... got it - I'll add a comment in the example

PR #1938, opened 2018-07-20
python.md: Replace broken pypy2.7 with pypy

jayvdb:
pypy2.7 is broken.

Related to https://github.com/travis-ci/travis-ci/issues/9452

It was introduced to the docs in https://github.com/travis-ci/docs-travis-ci-com/pull/1704

### DIFF
--- a/user/languages/python.md
+++ b/user/languages/python.md
@@ -86,19 +86,17 @@ virtualenv:
 
 Travis CI supports PyPy and PyPy3.
 
-To test your project against PyPy, add "pypy" or "pypy3" to the list of Pythons
+To test your project against PyPy, add "pypy3.5" to the list of Pythons
 in your `.travis.yml`:
 
 ```yaml
 language: python
 python:
-  - "2.6"
   - "2.7"
   - "3.4"
   - "3.5"
   - "3.6"
   # PyPy versions
-  - "pypy2.7"
   - "pypy3.5"
 # command to install dependencies
 install:


user/languages/python.md
```diff
@@ -102,7 +102,7 @@ python:
   - "3.5"
   - "3.6"
   # PyPy versions
-  - "pypy2.7"
+  - "pypy"
```
jayvdb: We could go even further and remove it altogether.
And then remove it from above:

> To test your project against PyPy, add "pypy" or "pypy3" to the list of Pythons
 in your `.travis.yml`:

But it will look a bit strange that we have "2.6" listed (which is almost impossible to use on Travis now because Python devs have removed 2.6 support from critical packages) and not "pypy".

user/languages/python.md
```diff
@@ -102,7 +102,7 @@ python:
...
   - "pypy3.5"
```
jayvdb: could also add
```yaml
  - pypy3.5-5.8.0
  - pypy3.5-5.9.0
  - pypy3.5-5.10.1
```

user/languages/python.md
plaindocs: So we could just remove 2.6 and pypy2.7 ?

user/languages/python.md
plaindocs: We could but there probably isn't a lot of value in adding many minor versions.

plaindocs:
Thanks!

PR #1940, opened 2018-07-20
clarify GitHub Releases uploading directory

StrikerRUS:
Refer to https://github.com/travis-ci/travis-ci/issues/6172#issuecomment-226825661 .

The current documentation states that all paths in `file` field are relative to `$TRAVIS_BUILD_DIR`, but it's not true. For instance, I've spent two hours to understand why Travis creates new Release and it's empty. The thing was I changed the current directory in at `before_deploy` stage.

Also, Appveyor's artifacts path is **always** relative to the building directory, so many users may expect the same behavior from Travis.

### DIFF
--- a/user/deployment/releases.md
+++ b/user/deployment/releases.md
@@ -4,7 +4,7 @@ layout: en
 
 ---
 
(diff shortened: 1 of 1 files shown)


plaindocs:
Thanks!

PR #1956, opened 2018-07-30
Fixed a typo

JonasJurczok:
I fixed a typo that might be confusing to newstarters.


plaindocs:
Oops! Thanks for spotting and fixing!

PR #1968, opened 2018-08-04
Remove misleading pytest comment in Python docs

adamchainz:
`pytest` is the new executable name introduced by Pytest 3.0, it's nothing to do with what version of Python is in use. The old `py.test` continues to work but is deprecated.


plaindocs:
Thanks for pointing that out!

PR #1970, opened 2018-08-09
Update julia docs

simonbyrne:
cc: @travis-ci/julia-maintainers

### DIFF
--- a/user/languages/julia.md
+++ b/user/languages/julia.md
@@ -16,7 +16,7 @@ This guide covers build environment and configuration topics specific to
 Travis CI support for Julia is contributed by the community and may be removed
 or altered at any time. If you run into any problems, please report them in the
 [Travis CI issue tracker](https://github.com/travis-ci/travis-ci/issues/new?labels=julia)
-and cc @tkelman @ninjin @staticfloat @simonbyrne.
+and cc [`@travis-ci/julia-maintainers`](https://github.com/orgs/travis-ci/teams/julia-maintainers).
 
 ## Choosing Julia versions to test against
 
@@ -30,8 +30,9 @@ versions, use the `julia:` key in your `.travis.yml` file, for example:
 language: julia
 julia:
   - nightly
-  - 0.5
-  - 0.5.2
+  - release
+  - 0.7
+  - 0.6.4
 ```
 {: data-file=".travis.yml"}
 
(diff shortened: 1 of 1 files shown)


user/languages/julia.md
```diff
@@ -30,8 +30,9 @@ versions, use the `julia:` key in your `.travis.yml` file, for example:
 language: julia
 julia:
   - nightly
-  - 0.5
-  - 0.5.2
+  - release
```
tkelman: this shouldn't be mentioned in docs or encouraged - support for it should be deleted from the code, really

simonbyrne: why so?

tkelman: Same reason it's been discouraged for years - "latest release" changes what's being tested against over time, but the information about supported versions of Julia for a package is contained in REQUIRE or Project.toml. Those versions are different and will get out of sync over time. It'd be great if the versions to test against could be auto-detected from the same source of information that Pkg uses and populate the matrix from that, but that's non-trivial. `release` is a legacy, long-obsolete holdover that made porting ancient `.travis.yml` files that were referencing the old juliareleases ppa a direct translation when `language: julia` was brand new. The ppa never had the direct flexibility to keep testing against all julia versions a package claims to support, the modern recommended way that declares explicit versions does.

simonbyrne: I hadn't seen any discouragement before, but that seems reasonable. In that case, we should probably throw a warning, and change the defaults.

tkelman: The defaults in pkgdev were changed in https://github.com/JuliaLang/PkgDev.jl/pull/108 and I went and submitted PR's to any package that had still been using `release` around that time, and removed it from the docs here in https://github.com/travis-ci/docs-travis-ci-com/pull/1188. Without it being in the PkgDev default script any more, I suspect very few people have intentionally added it back since. Hopefully it's not used in PkgTemplates or whatever people are using nowadays.

simonbyrne: On a related note: do you think we should now allow a single digit? i.e. `1` will give you the latest `1.y.z` release?

tkelman: I'd have to check the code, but that might already work automatically if there's a julia-1-latest binary download put in the right place on s3

simonbyrne:
I think this is ready to go. Can this be merged soon as the current information is outdated (#1979).

user/languages/julia.md
```diff
@@ -16,51 +16,49 @@ This guide covers build environment and configuration topics specific to
...
-`PkgDev.generate("$name", "$license")`, then the following default script will be run:
+If your repository contains `JuliaProject.toml` or `Project.toml` file, and you are
+building on Julia v0.7 or later, the default build script will be:
+```julia
+using Pkg
+Pkg.build()
+Pkg.test()
+```
 
+Otherwise it will use the older form:
```
thofma: Does one need a `using Pkg`?

simonbyrne: Good point, I've made this consistent.

plaindocs:
Looks great to me, merging.

PR #1972, opened 2018-08-11
Update s3.md

transdigiware:
Added dot_match example


plaindocs:
Happy to merge this is if you can just resolve the merge conflict first. Thanks.

user/deployment/s3.md
```diff
@@ -268,7 +268,20 @@ deploy:
...
 
-### Using S3-compatible Object Storage
+## Set dot_match flag to upload files starting with a period
+
+S3 uploads can be set to upload all files starting with a `.`
+
+```yaml
+deploy:
+  provider: s3
+  ..
```
BanzaiMan: Is this meant to be ellipses? I don't think this is necessary.

user/deployment/s3.md
```diff
@@ -268,7 +268,20 @@ deploy:
 ```
 {: data-file=".travis.yml"}
 
-### Using S3-compatible Object Storage
+## Set dot_match flag to upload files starting with a period
+
+S3 uploads can be set to upload all files starting with a `.`
```
BanzaiMan: I think "all files starting with a `.`" is somewhat misleading. Namely, it reads as though those are the only files that will be uploaded.

May I suggest changing this to: "all files including those starting with a `.`"?

user/deployment/s3.md
```diff
@@ -279,4 +292,4 @@ deploy:
   ..
   endpoint: https://nyc3.digitaloceanspaces.com
 ```
-{: data-file=".travis.yml"}
+{: data-file=".travis.yml"}
```
BanzaiMan: Not sure how this line is showing up in a diff (I can't detect the white space change), but I do prefer this not being a part of the PR.

user/deployment/s3.md
DrTorte: Agreed, let's do that.

user/deployment/s3.md
DrTorte: Let's take them out. :)

user/deployment/s3.md
DrTorte: Seems to be an oddity from fixing the conflict, but agreed.

user/deployment/s3.md
DrTorte: Looking over the doc I see them used a few times, which I believe was the reason for their original inclusion. I think taking them all out would be beneficial and more consistent with other docs, unless I'm mistaken and there are other places we use them?

PR #1974, opened 2018-08-12
improve readbility of rust.md

jdburnet:
Hi Travis CI! Thanks for all your work in open source 👏 

It seems to me that if we remove an extra 'always' from Default Build Scripts section of rust.md , it will improve the flow of that section. Of course I am open to making additional changes or just leaving as is, let me know 😄


plaindocs:
You are 100% correct. Merging.

PR #1980, opened 2018-08-15
Update pages.md

brianhou:
This is to reflect the new behavior supported by https://github.com/travis-ci/dpl/pull/756.


plaindocs:
Nice.

With the two options the example doesn't make much sense, we can probably delete, what do you think?

brianhou:
Done!

PR #1984, opened 2018-08-19
Fix php-fpm example

marcelklehr:
see https://github.com/travis-ci/travis-ci/issues/8968


plaindocs:
Thanks!

marcelklehr:
I just found out that apparently, the line I removed was necessary for php 7. It doesn't work when using php5, though. :/

Issue #910, opened 2017-01-12
Document `language: generic`


BanzaiMan:
The only current mention of this is in multi-os doc.

This is worth mentioning either as a section the languages page, or a page of its own.

acnagy:
There are a [few different ways](https://github.com/travis-ci/packer-templates/blob/master/cookbooks/travis_ci_connie/attributes/default.rb#L60-L67) to end up in the same place as `language: generic`. Is it worth listing them all?

BanzaiMan:
@acnagy IMO, yes, all except for `__connie__`, which is an obscure implementation detail that is likely to change over time.

hessam61:
Is there a difference between minimal and generic for language?  Any update on adding documentation for either of these languages? It seems that minimal is no longer supported, isa that correct?

Wilfred:
Apparently there's `language: bash` too, according to [this comment](https://github.com/travis-ci/travis-ci/issues/5175#issuecomment-215016943). It's not clear if there's any advantage of using that.

Wilfred:
OK, it looks like `bash` and `generic` are equivalent. See [this PR](https://github.com/travis-ci/travis-build/pull/306/files). [This PR](https://github.com/travis-ci/travis-yaml/pull/33/files) suggests that `sh` and `shell` are equivalent too.

[This comment](https://github.com/travis-ci/travis-ci/issues/2856#issuecomment-58609666) suggests that Ruby is the default if you don't use generic, so I imagine the build will be a little slower.

Finally, the only documentation mention of `generic` is [in this section of the multi-os docs](https://docs.travis-ci.com/user/multi-os/#Python-example-(unsupported-languages)), as mentioned above.

MariadeAnton:
We've introduced a language tag switch and now `language: minimal` is available instead.
Using `language: minimal` in either container-based and fully virtualized infras, it comes with fewer pre-installed languages and services and more disk space.

As Anna mentioned, language: generic points to the same image as `language: bash`, `language: sh` and `language: shell`. 

As a first step we're updating our multiOS docs to reflect this here: https://github.com/travis-ci/docs-travis-ci-com/pull/1647

And we're working on documenting this option in our language and infrastructure reference docs.

Thanks much for your patience y'all!

MariadeAnton:
As an extra note - this time claryfing `languague: generic`.

Using `language: generic` with our stable, Trusty image, will point to a complete image so that someone can use many/none programming language: 

[...]

PR #1989, opened 2018-08-22
Minimal and generic info

### LINKED ISSUE #1267: Document how to select minimal vs full Trusty build image
The documentation for the beta Trusty environment ([here](https://docs.travis-ci.com/user/trusty-ci-environment/#Image-differences-from-Precise)) says:

> For our Trusty based environments, we’re making a smaller set of images that includes:
> * minimal image which contains a small subset of interpreters, as well as docker and packer.
> * considerably larger image which contains (almost) all runtimes and services present in Precise environments.

However I can't find a mention of how Travis chooses between each of those images, and what the larger image contains that the smaller doesn't?

Initially I thought perhaps selecting the smaller image was performed by not specifying a `language` key in `.travis.yml` at all, however doing so resulted in the job installing ruby instead, which was a bit puzzling:
https://travis-ci.org/mozilla/treeherder/builds/242954322#L226

My use-case is trying to select the smaller image to speed up job times, since I'm experimenting with using Docker Compose (so that we can test against Ubuntu 16.04) - which means we don't need anything but Docker/Compose installed.

Many thanks!

plaindocs:
Fixes #1267 #910


user/languages/minimal-and-generic.md
```diff
@@ -0,0 +1,52 @@
...
+> Note that `language: minimal` is not the same as omitting the `language` key, if you do that the default language is set to Ruby.
+
+<aside markdown="block" class="ataglance">
+
+| Ruby              | Default                                   |
+|:------------------|:------------------------------------------|
+| Default `install` | N/A                                       |
+| Default `script`  | N/A                                       |
+| Matrix keys       | N/A                                       |
+| Support           | [Travis CI](mailto:support@travis-ci.com) |
```
MariadeAnton: can we add a subject line in the link here? That way it can help us track where people come from 🤔

plaindocs:
Hi @edmorley when you have a moment could you take a quick glance at this and see if it leaves any of your questions unresolved?

user/languages/minimal-and-generic.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Minimal and Generic images
+layout: en
+
+---
+
+## What This Guide Covers
+
+Travis CI supports many popular programming languages, but can never hope to support them all. `language: minimal` and `language: generic` are images running Ubuntu Trusty  that are not tailored to any particular programming language. Both are available on both sudo-enabled and container-based infrastructure. As their names suggest, one is optimized be faster and use less disk space, the other to have more languages and services available.
```
edmorley: There's a stray extra space character in the middle of `Trusty  that`

PR #1995, opened 2018-08-24
How to perform simple uptime checks with Travis Enterprise

schultyy:
Explain how to integrate Travis CI into a monitoring system by querying it for its uptime.


user/enterprise/platform-tips.md
```diff
@@ -309,6 +309,16 @@ Additionally, please the following command to clean up all Docker build images:
 $ sudo docker images | grep travis | awk '{print $3}' | xargs sudo docker rmi -f
 ```
 
+## Integrate Travis CI Enterprise into your monitoring
+
+To implement a simple uptime check for Travis CI Enterprise, you can query the `/api/uptime` endpoint from your instance.
```
plaindocs: It is best to avoid using words like "simple" here.

plaindocs: Also, this doesn't check "uptime" it just checks "upness" :-p

How about 

> To check if your Travis CI Enterprise (what? installation? cluster? Instance?) is up and running, query the `/api/uptime` endpoint of your instance.

We do mean of, not from, right?

user/enterprise/platform-tips.md
```diff
@@ -309,6 +309,16 @@ Additionally, please the following command to clean up all Docker build images:
...
+
+```
+$ curl -H "Authorization: token XXXXX" https://travis.example.com/api/uptime
+```
+
+If everything is up and running, it answers with a `HTTP 200 OK`.
```
plaindocs: Do we need to mention anything apart from a timeout here? Are there error codes?

user/enterprise/platform-tips.md
plaindocs: I've just noticed the endpoint is called uptime. :/

schultyy: Yeah, it's called `uptime` but returns a simple `OK` if everything checks out.

user/enterprise/platform-tips.md
schultyy: Good call! I checked [here](https://github.com/travis-ci/travis-api/blob/72fac430b2615e6f5ff7e3371f74bfa621ef59df/lib/travis/api/app/endpoint/uptime.rb) and we can either expect `200 OK` or a `500` in case of failure.

PR #1996, opened 2018-08-24
Explain how to reset stuck is_syncing user accounts

schultyy:
This addition to the operations manual explains how to deal with user accounts that are stuck in the `is_syncing = true` state on Travis CI Enterprise.


user/enterprise/operations-manual.md
```diff
@@ -145,6 +145,31 @@ This can have various causes, including an automatic nvm update or a caching err
 
 This error is most likely caused by a self-signed certificate. During the build, the worker container attempts to fetch different files from the platform machine. If the server got provisioned with a self-signed certificate, curl doesn't trust this certificate and therefore fails. While we're working on resolving this in a permanent and sufficient way, currently the only solution is to install a certificate issued by a trusted Certificate Authority (CA). This can be a free Let's Encrypt certificate or any other trusted CA of your choice. We have a section in our [Platform Administration Tips](/user/enterprise/platform-tips/#Use-a-Lets-Encrypt-SSL-Certificate) page that walks you through the installation process using Let's Encrypt as an example.
 
+## Users are stuck in syncing state
```
plaindocs: User *accounts* as below is probably better here

user/enterprise/operations-manual.md
```diff
@@ -145,6 +145,31 @@ This can have various causes, including an automatic nvm update or a caching err
...
+
+One or more user accounts are stuck in the `is_syncing = true` state. When you query the database, the number of users which are currently syncing is not decreasing over the time. Example:
```
plaindocs: s/is not decreasing/does not decrease/

user/enterprise/operations-manual.md
```diff
@@ -145,6 +145,31 @@ This can have various causes, including an automatic nvm update or a caching err
...
+
+```bash
```
plaindocs: s/bash/sql/ right? Assuming there is a sql. We might need to add it.

user/enterprise/operations-manual.md
```diff
@@ -145,6 +145,31 @@ This can have various causes, including an automatic nvm update or a caching err
...
+travis_production=> select count(*) from users where is_syncing=true;
+ count
+-------
+  1027
+(1 row)
+```
+
+### Strategy
+
+Log into the platform machine via ssh. Then execute `travis console` to get into Travis' Ruby console. Via that console, reset the `is_syncing` flag for the users by executing the following command:
```
plaindocs: Try to avoid excessive use of "via". 

Here you could say

1. Log into ...
2. Run `travis console` to open the Travis CI console
3. Reset the `is_syncing` for user accounts that are stuck by runing
    ```
    SQL
    ````

user/enterprise/operations-manual.md
```diff
@@ -145,6 +145,31 @@ This can have various causes, including an automatic nvm update or a caching err
...
+
+```bash
+$ travis console
+>> User.where(is_syncing: true).count
+>> ActiveRecord::Base.connection.execute('set statement_timeout to 60000')
+>> User.update_all(is_syncing: false)
+```
+
+It can happen that organizations are also stuck in the syncing state. Though an organization itself cannot be synced, but once all users belonging to it are synced the organization itself is considered synced as well.
```
plaindocs: I don't really know what this means.

user/enterprise/operations-manual.md
schultyy: Good catch. Hmm, given that it's SQL together with a CLI, I'll just omit the identifier here.

PR #2001, opened 2018-08-29
`brew outdated` always exits successfully

teor2345:
Recent versions of brew always exit 0 from `brew outdated`.

So `brew outdated <package-name> || brew upgrade <package-name>` never upgrades packages.

Use `brew install <package-name>` instead:
* if the package is not installed, install it
* if the package is outdated, upgrade it
* if the package is up to date, do nothing
* always exit successfully

### DIFF
--- a/user/reference/osx.md
+++ b/user/reference/osx.md
@@ -81,7 +81,7 @@ Or it can result in the command not found:
 xctool: command not found
 ```
 
-This is intended behaviour from Homebrew's side, but you can get around it by using [`brew bundle`](https://github.com/Homebrew/homebrew-bundle) or by first checking if the command needs an upgrade with `brew outdated`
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Hello, there. Thanks for the PR. Do you have references related to this change? How recent is this change in Homebrew?

teor2345:
I'm sorry, I mis-diagnosed the issue we were seeing.

Our issue is actually that `brew outdated <package-name> || brew upgrade <package-name>` only upgrades packages that are already installed. It doesn't install missing packages.

Recommending `brew outdated <package-name> || brew upgrade <package-name>` means that users have to guess whether Travis will install a particular package in the build image. `brew install <package-name>` does an update if the package is installed, and installs it if it isn't installed.

See #2017 for details.

PR #2002, opened 2018-08-29
OS X gcc is actually an alias for clang

teor2345:
OS X developers who use `compiler: gcc` are actually getting clang.

In the toolchain include and C reference, warn developers that they need to set a specific gcc version, and link to the details in the toolchain section.

### DIFF
--- a/_includes/c11-cpp11-and-beyond-and-toolchains.md
+++ b/_includes/c11-cpp11-and-beyond-and-toolchains.md
@@ -64,7 +64,8 @@ before_install:
 
 ### GCC on OS X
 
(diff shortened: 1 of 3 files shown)


plaindocs:
Cheers!

`_includes/c11-cpp11-and-beyond-and-toolchains.md` gets included in C++ as well, so presumable we need the same warning there?

teor2345:
> _includes/c11-cpp11-and-beyond-and-toolchains.md gets included in C++ as well, so presumably we need the same note there?

I updated the C++ docs, and also updated the included file to explicitly mention CXX.

I also merged master. Let me know if you'd like me to rebase and squash instead.

plaindocs:
Thanks!

PR #2012, opened 2018-09-10
Correct hyphen vs underscore typo

poch345:


DrTorte:
Thanks for opening this! We'll be merging this in to increase consistency.  

As a note, hyphens are treated as underscores either way (see here: https://github.com/travis-ci/dpl/blob/master/lib/dpl/cli.rb#L18) so you won't have to make any changes, if you are using hyphens already. 😸

PR #2015, opened 2018-09-12
Update docs for installing dependencies for Homebrew addon

mjm:
Expand the macOS section of Installing Dependencies with instructions on using the new Homebrew addon instead of running manual `brew` commands.

Also update the section on supporting macOS and Linux in the same .yml to show how to use both addons in tandem.

### DIFF
--- a/user/installing-dependencies.md
+++ b/user/installing-dependencies.md
@@ -275,9 +275,9 @@ addons:
 ```
 {: data-file=".travis.yml"}
 
-### Taps and Casks
+### Installing Casks
 
-The Homebrew addon supports more than just packages. It also supports installing [casks][homebrew-cask]: GUI apps and other things that need to run installers rather than being built from source code. Many casks are not very useful for automated building and testing, but there may be some that you need for your builds. You can add them to the `casks` key in the Homebrew addon configuration to install them:
+The Homebrew addon also supports installing [casks][homebrew-cask]. You can add them to the `casks` key in the Homebrew addon configuration to install them:
 
 [homebrew-cask]: https://github.com/Homebrew/homebrew-cask
 
@@ -289,12 +289,17 @@ addons:
 ```
 {: data-file=".travis.yml"}
 
-Homebrew also supports installing casks and packages from third-party repositories called [taps][homebrew-tap]. You can add these with the Homebrew addon. For instance, Homebrew maintains a tap of older versions of certain casks at [`homebrew/cask-versions`][cask-versions]. If you wanted to install Java 8 on an image with Java 10 installed, you can add that tap and then install the `java8` cask:
+### Installing From Taps
+
+Homebrew supports installing casks and packages from third-party repositories called [taps][homebrew-tap], and you can use these with the Homebrew addon.
+
+For instance, Homebrew maintains a tap of older versions of certain casks at [`homebrew/cask-versions`][cask-versions]. If you wanted to install Java 8 on an image with Java 10 installed, you can add that tap and then install the `java8` cask:
 
 [homebrew-tap]: https://docs.brew.sh/Taps
 [cask-versions]: https://github.com/Homebrew/homebrew-cask-versions
 
 ```yaml
+osx_image: xcode10
 addons:
   homebrew:
     taps: homebrew/cask-versions
@@ -313,8 +318,16 @@ addons:
   homebrew:
     brewfile: true
 ```
+{: data-file=".travis.yml"}
 
-You can also provide a path if your Brewfile is in a different location. For example: `brewfile: Brewfile.travis`.
+You can also provide a path if your Brewfile is in a different location.
+
+```yaml
+addons:
+  homebrew:
+    brewfile: Brewfile.travis
+```
+{: data-file=".travis.yml"}
 
 ## Installing Dependencies on Multiple Operating Systems
 
--- a/user/installing-dependencies.md
+++ b/user/installing-dependencies.md
@@ -254,20 +254,82 @@ addons:
 
 ## Installing Packages on OS X
 
-To install packages that are not included in the [default OS X environment](/user/reference/osx/#Compilers-and-Build-toolchain) use [Homebrew](http://brew.sh) in your `.travis.yml`. For example, to install beanstalk:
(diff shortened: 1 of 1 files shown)


user/installing-dependencies.md
```diff
@@ -254,20 +254,82 @@ addons:
...
+  homebrew:
+    packages:
+    - beanstalk
+    update: true
+```
+{: data-file=".travis.yml"}
+
+### Taps and Casks
+
+The Homebrew addon supports more than just packages. It also supports installing [casks][homebrew-cask]: GUI apps and other things that need to run installers rather than being built from source code. Many casks are not very useful for automated building and testing, but there may be some that you need for your builds. You can add them to the `casks` key in the Homebrew addon configuration to install them:
```
BanzaiMan: A few things on this paragraph:

1. "supports more than just packages" seems a bit wordy here. Just say "The addon also supports…".
1. "Many casks are not very useful for automated building and testing…" is probably unnecessary. Don't pass judgement on what is useful or not for the users' builds. If they are adding casks, they need it.
1. Also, if the user is adding a cask, it seems acceptable to assume that they know what a cask is.

user/installing-dependencies.md
```diff
@@ -254,20 +254,82 @@ addons:
...
+Homebrew also supports installing casks and packages from third-party repositories called [taps][homebrew-tap]. You can add these with the Homebrew addon. For instance, Homebrew maintains a tap of older versions of certain casks at [`homebrew/cask-versions`][cask-versions]. If you wanted to install Java 8 on an image with Java 10 installed, you can add that tap and then install the `java8` cask:
+
+[homebrew-tap]: https://docs.brew.sh/Taps
+[cask-versions]: https://github.com/Homebrew/homebrew-cask-versions
+
+```yaml
+addons:
+  homebrew:
+    taps: homebrew/cask-versions
+    casks: java8
```
BanzaiMan: This example requires an `osx_image` key (unless the default image now has Java 10).

user/installing-dependencies.md
```diff
@@ -254,20 +254,82 @@ addons:
...
+
+```yaml
+addons:
+  homebrew:
+    casks:
+    - dotnet-sdk
+```
+{: data-file=".travis.yml"}
+
+Homebrew also supports installing casks and packages from third-party repositories called [taps][homebrew-tap]. You can add these with the Homebrew addon. For instance, Homebrew maintains a tap of older versions of certain casks at [`homebrew/cask-versions`][cask-versions]. If you wanted to install Java 8 on an image with Java 10 installed, you can add that tap and then install the `java8` cask:
```
BanzaiMan: "For instance" should start a new paragraph.

user/installing-dependencies.md
```diff
@@ -254,20 +254,82 @@ addons:
...
+```yaml
+addons:
+  homebrew:
+    packages:
+    - beanstalk
+    update: true
+```
+{: data-file=".travis.yml"}
+
+### Taps and Casks
```
BanzaiMan: Perhaps split this section into "Casks" and "Taps". The header is "Taps and Casks", but Casks is treated first. Seems strange.

user/installing-dependencies.md
```diff
@@ -254,20 +254,82 @@ addons:
...
+  homebrew:
+    taps: homebrew/cask-versions
+    casks: java8
 ```
 {: data-file=".travis.yml"}
 
-> To speed up your build, try installing your packages *without* running `brew update` first, to see if the Homebrew database on the build image already has what you need.
+### Using a Brewfile
+
+Under the hood, the Homebrew addon works by creating a `~/.Brewfile` and running `brew bundle --global`. You can also use the addon to install dependencies from your own [Brewfile][] that is checked in to your project. By passing `brewfile: true`, the addon will look for a `Brewfile` in the root directory of your project:
```
BanzaiMan: Looks like you mean to add the ref:

    [Brewfile][brewfile]

user/installing-dependencies.md
```diff
@@ -254,20 +254,82 @@ addons:
...
+
+[brewfile]: https://github.com/Homebrew/homebrew-bundle
+
+```yaml
+addons:
+  homebrew:
+    brewfile: true
+```
+
+You can also provide a path if your Brewfile is in a different location. For example: `brewfile: Brewfile.travis`.
```
BanzaiMan: I think it is better to give a more complete YAML snippet here.

user/installing-dependencies.md
```diff
@@ -254,20 +254,82 @@ addons:
...
+Under the hood, the Homebrew addon works by creating a `~/.Brewfile` and running `brew bundle --global`. You can also use the addon to install dependencies from your own [Brewfile][] that is checked in to your project. By passing `brewfile: true`, the addon will look for a `Brewfile` in the root directory of your project:
+
+[brewfile]: https://github.com/Homebrew/homebrew-bundle
+
+```yaml
+addons:
+  homebrew:
+    brewfile: true
+```
+
```
BanzaiMan: Did you miss `{:data-file}`?

user/installing-dependencies.md
mjm: That's true. Only the `xcode10` image has Java 10, so adding `osx_image: xcode10` will make this a more real example.

user/installing-dependencies.md
mjm: Ok, I can adjust this.

user/installing-dependencies.md
mjm: 👍

user/installing-dependencies.md
mjm: In Markdown, an empty ref is the same as a ref with the contents of the link text (case-insensitive), so those are equivalent (and indeed the link is working in the preview). If you feel like using that property is confusing, though, I can change this to be more explicit.

user/installing-dependencies.md
mjm: Sure thing.

user/installing-dependencies.md
mjm: Yup!

mjm:
Thanks for the feedback, @BanzaiMan!

PR #2017, opened 2018-09-14
Recommend `brew install` rather than `brew outdated`

teor2345:
This is a fixed version of pull request #2001.

`brew outdated <package-name>` exits with 0 if <package-name> is
not installed.

So `brew outdated <package-name> || brew upgrade <package-name>` does
not install missing packages.

Use `brew install <package-name>` instead:
* if the package is not installed, install it, and exit 0
* if the package is outdated, upgrade it, and exit 0
* if the package is up to date, do nothing, and exit 0

### DIFF
--- a/user/reference/osx.md
+++ b/user/reference/osx.md
@@ -81,7 +81,9 @@ Or it can result in the command not found:
 xctool: command not found
 ```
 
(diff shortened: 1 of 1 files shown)


teor2345:
Obsoleted by #2023.

PR #2019, opened 2018-09-14
[julia] document coverage options

simonbyrne:
To be merged after https://github.com/travis-ci/travis-build/pull/1450


simonbyrne:
@BanzaiMan This can also be merged now.

plaindocs:
What is the reason for reverting the team? Seems easier to maintain and use.

simonbyrne:
I agree, unfortunately it doesn't work for users who aren't members of the Travis org, which makes it useless in this context: see https://github.com/travis-ci/travis-ci/issues/10003

PR #2020, opened 2018-09-16
Logo links to travis-ci.com instead of .org

tanelso2:
It's been bothering me whenever I click on the header when looking at the docs and I get taken to a different website than what I expect.

Let me know if linking to .org is still intentional!


DrTorte:
So, we will eventually be moving everything over to .com, and new users can sign up for public repositories there rather than on .org - the change will happen sooner or later, anyway, so might as well do it now. Gonna go ahead and merge this in. Thanks!

PR #2023, opened 2018-09-18
OS X: recommend the Travis Homebrew addon in the Homebrew section

teor2345:
@mjm suggested adding the Homebrew addon to the OS X documentation in https://github.com/travis-ci/docs-travis-ci-com/pull/2017#issuecomment-421444250

This pull request obsoletes #2017 and #2001.

### DIFF
--- a/user/reference/osx.md
+++ b/user/reference/osx.md
@@ -59,63 +59,11 @@ Travis CI uses OS X 10.13 and Xcode 9.4.1 by default . You can use another versi
 
 Homebrew is installed and updated every time the virtual machines are updated.
 
-> To speed up your build, try installing your packages *without* running `brew update` first, to see if the Homebrew database on the build image already has what you need.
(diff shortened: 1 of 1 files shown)


mjm:
Thanks for the PR, @teor2345!

I'm thinking that maybe we should take a stronger approach here. Instead of adding mentions of the addon around the existing instructions about manually installing packages, this could be an opportunity to clean up and simplify this page by removing a lot of that information, since the addon eliminates the need to worry about most of it.

If the "Homebrew" section of this page was just a short paragraph introducing and linking to the addon, I think we could eliminate most of these other headings completely, and that would make for clearer documentation. Can we give that a try?

teor2345:
> If the "Homebrew" section of this page was just a short paragraph introducing and linking to the addon, I think we could eliminate most of these other headings completely, and that would make for clearer documentation. Can we give that a try?

Sure! Please see the updates I force-pushed to the pull request branch.

I'm not sure if you want to preserve the old heading anchors, or how to do that in markdown.

mjm:
This looks great to me! Thanks @teor2345!

PR #2025, opened 2018-09-19
update how yarn is detected for node_js language

chippers:
The documentation states the "root directory of the repository", however it is really the current directory that travis-ci is in that it is detected.  In example, if you have a build matrix that has multiple projects in a repository to test then you might be misled that you need it in the root directory of your repository instead of just normally having it in your sub-directory - [like in my example](https://github.com/chippers/celur/commit/db23613d3c8d8c86871029e99e30f98b3ee68ce2).


plaindocs:
Hi @chippers - thanks for the PR.

PR #2029, opened 2018-09-25
Fix anchor link in API 2.1 docs

iriberri:
@plaindocs hi! 👋 

It was brought to our attention that there was a typo here, so this corrects it. Let me know if this is not the way to update these old API docs, I honestly can't barely remember how these were built.

Thanks!


plaindocs:
Looks great to me - thanks!

We used to have the html output of the slate in the repo as well, which led to some folks updating the html directly, and then having to backport changes. Now we only have the source. There is even a mention in the READE I think.

PR #2041, opened 2018-10-02
Set up CI with Azure Pipelines

oscarg933:


BanzaiMan:
Sorry, but this is not valid for our docs site.

I'm closing this now, but if you feel strongly, please tell us why this is a good idea.

PR #2045, opened 2018-10-02
Remove mention of a container-based Linux default

meatballhat:
since it is going away and the current default is in transition

### DIFF
--- a/user/reference/overview.md
+++ b/user/reference/overview.md
@@ -94,6 +94,11 @@ Many different parts of your `.travis.yml` affect what infrastructure your build
 
 * Using `os: osx`, setting a version of Xcode using `osx_image:`, or using a macOS specific language such as `language: objective-c` routes your build to macOS infrastructure.
 
+> Between middle of October 2018 and end December 2018 the default infrastructure
+> your builds runs on will depend on a [few different
+> factors](https://blog.travis-ci.com/2018-10-04-combining-linux-infrastructures)
+> while we consolidate everything onto sudo-enabled virtual machines.
+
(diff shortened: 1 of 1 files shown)


edmorley:
@meatballhat what will the new default be, out of curiosity? (And hi again!)

edmorley:
Ah thank you!

plaindocs:
I'd prefer to still explicitly mention a default, even if it is short term link to the blog post that we can switch out in a few weeks. What do you think @meatballhat ?

acnagy:
@plaindocs would it be wrong to link to the blog post for the timeline for the transition? right now it's `.org` that's all vm-based, starting 🔜 `.com` default will move over with percent-based sampling. Thennnn.. there's a few weeks of "everyone hopefully updates their builds" then we start overriding `sudo: false` specifications

plaindocs:
How about something like

> Between middle of October 2018 and end November 2018 the default infrastructure your builds runs on will depend on a [few different factors](https://blog.travis-ci.com/2018-10-04-combining-linux-infrastructures) while we consolidate everything onto sudo-enabled virtual machines. 

adjusting dates as needed. 

@acnagy as an aside, what do you think about using virtual machines in full in docs and blogs?

acnagy:
@plaindocs sounds perfect to me. Re:dates - we're planning Oct 2018 - Dec 2018 (last overrides will run into December). 

As to "vms" vs. "virtual machines" vs. "sudo enabled" - yeah, def agreed we should use "virtual machine"... and now I realize I need to fix my blog post: "vm" --> "virtual machine" 😅

PR #2046, opened 2018-10-03
Environment variables: disambiguation

igneus:
[I was not the only one](https://stackoverflow.com/questions/40294597/how-to-properly-define-multiple-env-variables-on-travis-yml) confused by the paragraph on "multiple environment variables per build" and reading it so that if each entry in the `env` array contains exactly one variable, they will be all applied to a single build.

This PR attempts to describe the actual behavior of the build system unambiguously.


igneus:
Found both remarks reasonable, tried to implement them.

plaindocs:
Thanks.

PR #2060, opened 2018-10-15
Remove irrelevant tag from the output

pothitos:
Hello!

I think that the commit is self-descriptive :relaxed:


BanzaiMan:
Indeed! Thank you.

PR #2073, opened 2018-10-21
Fix broken links in Enterprise docs

jamrok:


jamrok:
Hey @plaindocs,

Thanks for merging!

Anchor links are harder to check if you're trying to validate if the link returns a 200 response or not. At first glance it seems that that's what the html-proofer is doing, which would explain why the build passes (the main pages/links are reporting a 200, but you can't check if it jumped to the correct tag like that).

I assume Github made changes recently that broke the anchor links. The fix I did was to update the anchor links to those that "should" be generated based on the markdown headers and I did other verifications to confirm accuracy.

I wrote a (bash) script to fix and check for broken anchor links, but I need to review the code for html-proofer in more detail and tweak it to test for anchor links as well. If you figure it out and fix it first, then that's great, but I'll still do it eventually (when I have time) to learn and test out Travis :smile:

PR #2093, opened 2018-10-30
Tweak treatment of `brew` command

BanzaiMan:
Explicitly state the use of `brew` as a standalone command,
and recommend switching to Ruby 2.3+ when an older version is
in use in such a case.

To correspond with https://github.com/travis-ci/travis-build/pull/1554

### DIFF
--- a/user/installing-dependencies.md
+++ b/user/installing-dependencies.md
@@ -254,7 +254,10 @@ addons:
 
 ## Installing Packages on OS X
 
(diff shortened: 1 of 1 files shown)


user/installing-dependencies.md
```diff
@@ -329,6 +332,22 @@ addons:
 ```
 {: data-file=".travis.yml"}
 
+### Using Homebrew without addon on older macOS images
+
+It may be convenient to invoke the `brew` command directly (e.g, you write
```
plaindocs: What do you think about something like

> If you're running the `brew` command directly in your build scripts, and you're using an older macOS image you'll need to update to Ruby 2.3 or newer.

And if we can tell them which is the macOS image that would be awesome. Ie macOS X or older...

PR #2101, opened 2018-11-01
Add CloudFoundry app_name

abdonrd:
Reference: https://github.com/travis-ci/dpl/pull/831


plaindocs:
Hi @abdonrd thanks for this. 

It does lead me to wonder why we have both though... And reckon we could probably fix that in this PR. What do you think?

abdonrd:
At https://github.com/travis-ci/dpl/pull/831#issuecomment-435150735, @BanzaiMan said that the CloudFoundry "children" are no longer descendants of it, so we have to duplicate it.

plaindocs:
Thanks!

PR #2102, opened 2018-11-02
Fix headings for in-page TOC

plaindocs:
Hey @mjm @Lyoness @cotsog @bogdanap 

Using these levels of heading displays slightly nicer with with the in page toc. 
Don't think these changes should break anything. :-)


plaindocs:
(beyond the trivial in page stuff I'm fixing that is :-P)

PR #2104, opened 2018-11-03
Add missing token scope to documentation

Malax:
I've had my build reported as failed for a week due to a, seemingly, failed gh-pages deploy. However, my branch was successfully updated on each build, but the build still showed up as errored. After some digging, I found out that the personal token also requires `user` scope. After creating a new token with both `user` and `repo` scope, my build is reported as succesful.

### DIFF
--- a/user/deployment/pages.md
+++ b/user/deployment/pages.md
@@ -34,7 +34,7 @@ deploy:
 
 You'll need to generate a [personal access
 token](https://help.github.com/articles/creating-an-access-token-for-command-line-use/)
-with the `public_repo` or `repo` scope (`repo` is required for private
+with the `user` as well as the `public_repo` or `repo` scope (`repo` is required for private
 repositories). Since the token should be private,
 you'll want to pass it to Travis securely in your [repository
 settings](/user/environment-variables#defining-variables-in-repository-settings)


plaindocs:
Hi @Malax 

Thanks for looking into this.

I can't reproduce the error, and wonder if it could possibly have been related to recent GH outages? https://status.github.com/messages

Malax:
I just recently fixed it and opened the issue afterwards. I think the outage had nothing to do with it. The affected repository was https://github.com/Malax/elmboy

Maybe it has something to do with the way I set it up? I can just say that the additional scope fixed it for me. :) I'm happy to share more details if required!

BanzaiMan:
@Malax Hello, there. I'm a little perplexed that the Pages provider requires the `user` scope. https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/#available-scopes These things are not necessary for the pages release.

In addition, I reviewed all the API calls that the pages provider makes, and none of them should raise `Octokit::Unauthorized`, which results in [the error you saw](https://travis-ci.org/Malax/elmboy/builds/449585757#L557). This exception is a clear indication that your GitHub token was invalid. Your deployment was fixed, not because of the scope you added, but because it is a valid one.

Please try dropping the `user` scope from your current token, and try again. It should still work.

I'm closing this now, but if you encounter the problem after the test, we are happy to revisit it.

Thanks.

PR #2106, opened 2018-11-05
Add  oraclejdk11 to Trusty docs

gliptak:


gliptak:
@BanzaiMan please review

PR #2110, opened 2018-11-07
Simplifty the Snaps documentation

joshk:


plaindocs:
Xenial only??

MariadeAnton:
Xenial and beyond, once we have those! 
Xenial is the first Ubuntu distribution that supports snaps :)

PR #2113, opened 2018-11-07
Fixed the actually install and test flags for go

Ilyes512:


plaindocs:
Thanks for noticing and fixing this!

Issue #2097, opened 2018-10-31
Error in Elastic Beanstalk doc


svianney:
Hi, 
I'm here to point out a little error in the documentation that got me some trouble. 
https://docs.travis-ci.com/user/deployment/elasticbeanstalk/

Here is the culprit
```
deploy:
  provider: elasticbeanstalk
  access_key_id: "Encrypted <access-key-id>="
  secret_access_key:
    secure: "Encypted <secret-access-key>="
```

In fact, in this state, the value of the node `access_key_id` has to be the unencrypted access_key_id like so `access_key_id: "<access_key_id>"`.

But If you want to also secure the access_key_id you have to put the encrypted access key id under a `secure` node like it's done with the `secret_access_key`

PR #2123, opened 2018-11-13
secure key is needed when using encrypted key id

Hjortsberg:
I was unable to use this config since the example missed the secure key for access key id.


plaindocs:
Thanks for fixing this @Hjortsberg !

Also fixes https://github.com/travis-ci/docs-travis-ci-com/issues/2097

PR #2128, opened 2018-11-14
Updates to the path on GitHub to see PAT's

logankilpatrick:
Added a path fix in order to eliminate any confusion and help users find the new way of creating a Personal Access Token. It originally took me 5-10 minutes to figure out how to generate the PAT and hopefully the new links will resolve this confusion.


user/private-dependencies.md
```diff
@@ -236,7 +236,7 @@ Assumptions:
 
 This approach works just like the [password](#Password) approach outlined above, except instead of the username/password pair, you use a GitHub API token.
 
-Under the GitHub account settings for the user you want to use, navigate to [Applications](https://github.com/settings/applications) and generate a "personal access tokens". Make sure the token has the "repo" scope.
+Under the GitHub account settings for the user you want to use, navigate to [Settings](https://github.com/settings/tokens), [Developer settings](https://github.com/settings/developers), and then generate a "Personal access tokens". Make sure the token has the "repo" scope.
```
plaindocs:
```suggestion
Under the GitHub account settings for the user you want to use, navigate to [Settings > Developer settings](https://github.com/settings/developers), and then generate a "Personal access tokens". Make sure the token has the "repo" scope.
```

This is a great idea, but I think the two links is confusing. What do you think?

logankilpatrick:
Updated per the recommendation of @plaindocs

user/private-dependencies.md
logankilpatrick: Yes, I agree! It should be intuitive once the user gets to that point.  No need to have 2 links in place. I updated it per that suggestion.

PR #2129, opened 2018-11-14
Add a reference that completes the information

native-api:
A lack of a reference creates an illusion that this is the entire info on this matter since this is reference documentation -- a section on subj is expected to contain full, authoritative info on subj.


plaindocs:
Thanks!

PR #2131, opened 2018-11-15
Rework Releases docoumentation

BanzaiMan:
To correspond to https://github.com/travis-ci/dpl/pull/923

Traditionally, we've said that Releases requires a tagged build (and restrict deployment with `on.tags: true`), but this is not necessarily true. Some workflows work better without a tag (when tag names are not known ahead of time, or when a draft release is desired). I tried to explain this in detail.

### DIFF
--- a/user/deployment/releases.md
+++ b/user/deployment/releases.md
@@ -21,10 +21,50 @@ deploy:
 
 > Make sure you have `skip_cleanup` set to `true`, otherwise Travis CI will delete all the files created during the build, which will probably delete what you are trying to upload.
 
-GitHub Releases uses git tags. If the build commit does not have any tags, one will be created in the form of `untagged-*`, where `*` is a random hex string.
+In general, GitHub Releases works with git tags, and it is important that
+you understand how tags affect GitHub Releases.
 
-If this is not what you want, either set your build to deploy only when the build already has a tag using `on.tags: true` as shown in the previous example `.travis.yml`, or tag the commit with `git tag` in `before_deploy`:
+## Deploying only on tagged builds
 
+With [`on.tags: true`](/user/deployment/#conditional-releases-with-on),
+your Releases deployment will trigger if and only if the build is a tagged
+build.
+
+## Regular releases
+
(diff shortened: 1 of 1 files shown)


user/deployment/releases.md
```diff
@@ -21,9 +21,53 @@ deploy:
 
 > Make sure you have `skip_cleanup` set to `true`, otherwise Travis CI will delete all the files created during the build, which will probably delete what you are trying to upload.
 
-GitHub Releases uses git tags. If the build commit does not have any tags, one will be created in the form of `untagged-*`, where `*` is a random hex string.
+In general, GitHub Releases works with git tags, and it is important that
```
plaindocs:
```suggestion
GitHub Releases works with git tags, so it is important that
```

user/deployment/releases.md
```diff
@@ -21,9 +21,53 @@ deploy:
...
+{: data-file=".travis.yml"}
+
+the resultant deployment is a draft Release that only repository collaborators
+can see.
+This gives you an opportunity to examine and edit the draft release.
+
+## Releases without `on.tags: true`
+
+It is desirable to use the Releases provider without `on.tags: true`,
+depending on the workflow.
```
plaindocs: This doesn't actually tell me what depends on what

BanzaiMan: The "workflows" I am thinking of are:

1. Setting the tag at the deployment time (explained in the section below)
1. Creating a release draft (which is explained above)

BanzaiMan: Whether the user wants to set `on.tags: true` or not depends on the workflow. I think the current document implicitly assumes that it is desirable, but this is not true.

plaindocs: So it would probably make sense to set this out up front. If the first workflow is about `on.tags: true` and says so, the second could start with something like

> If you're want to release without a known tag, or generate one during the build....

Which avoids the "depends" without a clear link to what depends on what

plaindocs: Works for me :-p

PR #2134, opened 2018-11-16
Make explicit that `file` is not available for GCS

PK1312:
The `file` option is not available for GCS deployment- you have to specify a directory. This is a small update that makes that explicit to avoid confusion.


BanzaiMan:
GCS is currently not functional at all. See https://github.com/travis-ci/dpl/issues/792

plaindocs:
@BanzaiMan does this want closing?

BanzaiMan:
Let's close it.

PR #2136, opened 2018-11-20
Warn about single encrypted file

DarthGandalf:
I lost a file because of this issue, and will need to recreate it. Let others to not suffer the same way until this is fixed.

Ref https://github.com/travis-ci/travis.rb/issues/627


user/encrypting-files.md
```diff
@@ -27,6 +27,7 @@ Assumptions:
 - You have version **1.7.0** or later of the Travis CI Command Line Client installed and set up (you are logged in)
 - You have a local copy of the repository and a terminal open where your current working directory is said copy
 - In the repository is a file, called super_secret.txt, that you need on Travis CI but you don't want to publish its content on GitHub.
+- You need to ever encrypt a single file. Encrypting another file later will [overwrite](https://github.com/travis-ci/travis.rb/issues/627) the secure variable, rendering the old file unusable. Look [below](#encrypting-multiple-files) for a workaround.
```
BanzaiMan: > You need to ever encrypt a single file.

Not sure if this conveys the idea well. Perhaps something more straightforward, like "You have only one file to encrypt."

DarthGandalf: What about "You have only one file to encrypt, ever."?

BanzaiMan: You deleted your original file and didn't have a backup?

DarthGandalf: Yeah. It's SSH key. I tend to not reuse private keys, so wouldn't need a backup of it. If the private key is leaked, I'd need to update too many places.

Creating new SSH key was not pleasant, but not too difficult.

DarthGandalf: To provide a timeline:
2 years ago I configured Travis to do something over SSH on server A.
Now I've configured it to do something else over SSH on server B, but that broke login to server A, because the key to decrypt SSH key was overwritten.

So I had to configure A again, but this time used the manual method for both A and B, so that they use different variables.

BanzaiMan: Hmmm. I feel that that is a different concern. With `travis encrypt-file`, the user will not see the variables they need to decrypt the file, so I am surprised that they might have an expectation of recovery (and remove the original file). We can discuss that, I suppose.

But "You have only one file to encrypt, ever." sounds definitely incorrect here.

DarthGandalf: The recovery is not a big concern. But the fact that new `travis encrypt-file` invalidates results of old `travis encrypt-file` is, which is what it did.

The problem is that server A stopped working even though I didn't touch it, but only because I used `travis encrypt-file` again. Even if I had a backup of the file, I'd need to re-submit that file to the git repo anyway, because old encrypted file became unreadable. Such resubmission could potentially follow the path of archiving those files into one, and encrypting them as one file... until I need to add yet another encrypted file, which would then invalidate both old files.

BanzaiMan: > But the fact that new `travis encrypt-file` invalidates results of old `travis encrypt-file` is, which is what it did.

I understand. And this is the very reason that trying to encrypt multiple files is problematic, and also the reason that we explain how to encrypt multiple files.

Stating "You need to ever encrypt a single file." or similar is misleading, in my opinion. So, what we can do instead is to state this fact, rather than to ask the users to look into the future and decide now whether or not this is the only file they are going to encrypt.

DarthGandalf: I agree that looking into the future is not very productive. By that sentence I meant to provide a summary (TL;DR) of this item, which explained the same thing you are saying, but in longer words. I've removed this sentence altogether now, and made the text shorter, so it doesn't need a summary now at all. PTAL.

plaindocs:
@DarthGandalf  @BanzaiMan this looks fine to me now.

PR #2141, opened 2018-11-23
windows.md:add supported languages - c and cpp

andy5995:
According to @joshk, these lanaguages are now supported. See this link
for a reference:
https://travis-ci.community/t/c-support-on-windows/282/10


plaindocs:
Thanks @andy5995 !

PR #2171, opened 2018-12-07
Julia: Document verbose build

fredrikekre:
Document verbose build, see https://github.com/travis-ci/travis-build/pull/1615.

cc @ararslan, @simonbyrne


simonbyrne:
The comment is a little confusing, since it won't work on versions before 1.1. Maybe something like
```julia
Pkg.build() # Pkg.build(; verbose = true) for Julia 1.1 and up
```

plaindocs:
Oh dang it, not yet merged right?

plaindocs:
Reverted, sorry for the noise folks!

PR #2172, opened 2018-12-07
Surface the `npm ci` alternative in the top table

compulim:
This will make sure if people are overriding `install`, they know if they should use `npm install` or `npm ci`.


plaindocs:
Hey @compulim thanks for this! 

I'd prefer to keep the explanation out of the sidebar, as it gets a little complicated. Adding the command to the table is great though! 

Do you think the "`npm install` or `npm ci`" line, plus the link to dependency management where we explain that is enough? I think it probably is.

user/languages/javascript-with-nodejs.md
```diff
@@ -10,11 +10,13 @@ layout: en
 
 | JavaScript and Node.js                      | Default                                   |
 |:--------------------------------------------|:------------------------------------------|
-| [Default `install`](#dependency-management) | `npm install`                             |
+| [Default `install`](#dependency-management) | `npm install` or `npm ci`                 |
 | [Default `script`](#default-build-script)   | `npm test`                                |
 | [Matrix keys](#build-matrix)                | `env`, `node_js`                          |
 | Support                                     | [Travis CI](mailto:support@travis-ci.com) |
 
+(If `package-lock.json` or `npm-shrinkwrap.json` exists and your npm version supports it, Travis CI will use `npm ci` instead of `npm install`.)
```
plaindocs:
```suggestion

```

Pezi777:
Thank you compulim :-)

PR #2177, opened 2018-12-10
Clarify the purpose of setting the CACHE_NAME var

danielcompton:
When I was reading these docs, I had to look quite closely to work out why it was recommended to create an environment variable for `CACHE_NAME`. This PR might make it a little bit clearer, though there are probably ways to improve the language further.


plaindocs:
Good call!

PR #2183, opened 2018-12-11
Re-order concepts to aid progressive understanding

muizzk:
@plaindocs Minor update to aid incremental understanding of core concepts.


plaindocs:
Nice.

PR #2187, opened 2018-12-14
Add another deployment example with conditions

iriberri:
From a support request, it seems we're lacking an example in this page for multiple deployments but using conditions as well.

What do you think, @plaindocs? :) Thanks!


plaindocs:
:100: !

iriberri:
Looks good in Netlify as well!

imagen

will merge as soon as the rebase passes the tests :)

PR #2189, opened 2018-12-14
Extend and update beta migration docs

iriberri:
Pinging @MariadeAnton, @clekstro and @drogus here :)


iriberri:
By the way, since I'm including the Crons change here as well, I'll wait to merge until that is actually on production which should be soonish :)

iriberri:
I'm gonna remove the cron jobs change from here and ship it :)

PR #2190, opened 2018-12-15
Adding External Storage to Android Emulator Creation command

GabrielBB:
Adding a 100 MB SDCard parameter so users can run automated tests that need to write to External Storage. This is to avoid suffering from the following issue: [Android Emulator throws “Permission Denied”](
https://travis-ci.community/t/android-emulator-throws-permission-denied/1229) which is difficult to figure out because the error message is not helpful at all and most people just create emulators using Android Studio AVD and don’t know that this parameter is even necessary.


plaindocs:
Thanks @GabrielBB

PR #2194, opened 2018-12-18
Fix command to encrypt for com projects

daviddavis:
The command here doesn't work and issues an error, "unknown command
--com". I believe the com option should come after encrypt.


plaindocs:
:100: correct! Thanks for fixing.

PR #2199, opened 2018-12-27
Updated csharp.md

jaaufauvre:
* Mono: "xbuild tool is deprecated and will be removed in future updates, use msbuild instead"
* The MSTest framework is actually supported when testing against .NET Core
* NUnit.Runners is deprecated (replaced by NUnit.Console)
* Fixed title capitalization


plaindocs:
Many thanks!

PR #2201, opened 2018-12-29
snap provider only works on Ubuntu 16.04+

abitrolly:
Because `snapcraft` is only available there
https://packages.ubuntu.com/search?keywords=snapcraft


plaindocs:
Thanks!

PR #2214, opened 2019-01-08
include when the git clone happens

MariadeAnton:
Hi @plaindocs!

What do you think about including this here? Would it make sense to also talk about the add-ons?


plaindocs:
:+1:  to including both.

To be nitpicky though you can't say 'repo is cloned and after that you can customize'... because the act of customization is when you edit your config. Which is before the cloning. 

So something like like 

> Travis CI provides a default build environment and a default set of phases for each programming language. A virtual machine is created with the build environment for your job, your repository is cloned into it, optional addons are installed and then your build phases are run.

?

plaindocs:
But I'm not mega keen on that suggestion either :-p

MariadeAnton:
Ended up going with a mix & match - could you take another look, @plaindocs? Thank you!

PR #2215, opened 2019-01-08
mention ruby pronto issues with shallow cloning

kubamahnert:
I have encountered an issue with ruby's pronto (automated code review), specifically it's rugged git wrapper. It is caused by the default --depth=50 switch on the git clone command, and can be fixed fully by overriding the depth switch, just as mentioned in the docs. However, I have found no actual mention of the caveats of this solution anywhere, and I think that this information would definitely help the next person having a similar problem, given that pronto is pretty much the standard in ruby CI solutions.

### DIFF
--- a/user/customizing-the-build.md
+++ b/user/customizing-the-build.md
@@ -85,6 +85,8 @@ git:
 ```
 {: data-file=".travis.yml"}
 
+> Some operations on the repository, such as common automated code review scripts (e.g. Pronto for Ruby), may fail due to the limited git clone depth, not being able to access all the objects in the repository. Removing the depth flag, or running `git fetch --unshallow` might solve the issue.
+
 ## Git Clone Quiet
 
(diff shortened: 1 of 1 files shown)


user/customizing-the-build.md
```diff
@@ -85,6 +85,8 @@ git:
 ```
 {: data-file=".travis.yml"}
 
+> Some operations on the repository, such as common automated code review scripts (pronto for ruby) may fail due to the limited git clone depth, not being able to access all the objects in the repository. Removing the depth flag, or running git fetch --unshallow might solve the issue.
```
BanzaiMan: Needs a comma after "…ruby)".

kubamahnert:
added the comma @BanzaiMan requested + put the code snippet into code tags

plaindocs:
We should probably capitalize Pronto and Ruby as well.

kubamahnert:
@plaindocs fixed

PR #2219, opened 2019-01-10
Artifacts addon also runs after failure

iriberri:
:-)


BanzaiMan:
It should be `after_script`. https://github.com/travis-ci/travis-build/blob/5a1fdb7c4ac8bacd3f3d438bf135023d0a1fe554/lib/travis/build/addons/artifacts.rb#L24-L28

iriberri:
Oooooh, thanks @BanzaiMan. I ran a test with job that I made fail, but hadn't checked the actual phase where it ran.

Then I think this might need some rewording. The PR comes from a Support question where someone was asking if it was possible to upload artifacts as well when the job failed, since docs mention `after_success` explicitly.

user/uploading-artifacts.md
```diff
@@ -5,7 +5,7 @@ layout: en
 ---
 
 Travis CI can automatically upload your build artifacts to Amazon S3, after the
-[`after success`](/user/job-lifecycle/) stage of the
+[`after success` or `after_failure`](/user/job-lifecycle/) phases of the
```
iriberri: What about keeping it a simple:

```suggestion
end of the job.
```

instead?

plaindocs: We're not entirely consistent, but we include the deploy steps in the [job lifecycle](https://docs.travis-ci.com/user/job-lifecycle/) which is probably why I was specific there.

plaindocs:
@iriberri what status is this in? Do you want me to look at it? Or are you rewriting?

PR #2221, opened 2019-01-10
Update ruby.md

ssabrii:
Added info regarding Bundler as some variations still fail, the best way to avoid failure is to ensure running the command added in the amended doc.


plaindocs:
@BanzaiMan would you mind taking a look at this?

PR #2225, opened 2019-01-11
Explain how to uninstall Bundler 2.x safely

BanzaiMan:


BanzaiMan:
This supersedes #2221.

plaindocs:
Awesome, thanks @BanzaiMan !

user/languages/ruby.md
```diff
@@ -166,6 +176,23 @@ If you find your builds are failing due to “bundler not installed” errors, t
...
+    ```
+    {: data-file=".travis.yml"}
+
+  The `gem uninstall` command above removes any Bundler 2.x installed in
+  RVM's "global" gemset during the Ruby run time installation, which would be
+  selected as the default `bundle` command.
+  We ignore the failure from that command, because the failure most likely
+  means that there was no matching Bundler version to uninstall.
+
+In your build configuration may require a combination of these workarounds.
```
plaindocs:
```suggestion
Your build configuration may require a combination of these workarounds.
```

PR #2229, opened 2019-01-14
Test feedback with Netlify forms

plaindocs:


plaindocs:
Nothing here worth keeping. Feedback mechanism still worth investigating and implementing someday.

PR #2249, opened 2019-01-25
Consolidating on ataglance side panel

PatMyron:
and linking Github profiles in Dart, Haxe, Julia, and Nix

No files left with both Matrix keys in the side panel and an explicit repetitive "Build Matrix" section:

    user/languages $ grep -l "ataglance" `grep -l "Build Matrix" *`

I removed some "Default Script" sections that were strictly repetitive as well.

I also added the env key to the Elm build matrix.

I created side panels for C#, Crystal, D, Dart, and Perl 6. These files are missing side panels after this:

	user/languages $ grep -L "ataglance" *

	android.md
	haxe.md
	julia.md
	nix.md
	objective-c.md
	perl.md
	r.md
	smalltalk.md


PatMyron:
CC'ing the Crystal, D, and Dart folks for the only substantial changes made (new side panels):

Crystal: @asterite, @jhass, @waj, and @will
D: @MartinNowak and @wilzbach
Dart: @nex3 and @a14n

Screenshots:
crystal
d
dart

wilzbach:
Thanks a lot for this @PatMyron. Looks great from the D side! Looking at the screenshot I realized though that there's a newline or at least comma missing after "GDC's issue tracker". While it wasn't introduced in this PR, any chance we could quickly fix it here?
Also:
- there's now `ldc-latest-ci` (so a nightly build for LDC)
- the example versions are more than four years old, 2.084.0 is the latest DMD and 1.13.0 for LDC (maybe use these instead)

You can find out more about the install script that's used here: https://dlang.org/install

wilzbach:
Oh and if you want to link the list of all versions for D (like other languages do) that would be:
- [DMD](https://dlang.org/changelog/index.html)
- [LDC](https://github.com/ldc-developers/ldc/releases)
- [GDC](https://gdcproject.org/downloads)

plaindocs:
Hey, this looks like some great work! I'll review in detail so we can get this merged ASAP.

I'm not sure if entirely deleting the "Build Matrix" section, like in `clojure.md`, and leaving it in the "ataglance" bar is the way to go, even the information is entirely duplicated.

plaindocs:
I'll probably try to add the remaining side panels in while we're here as well.

PatMyron:
@plaindocs I'll add some more side panels, but I struggled with pages that contextually decide defaults based on user configurations (Like Android based on whether they are using Maven or Gradle).

plaindocs:
Yeah, there is some of that in JS as well (https://docs.travis-ci.com/user/languages/javascript-with-nodejs/)

I can take another look at it and see if we can come up with a better to show it.

Maybe even something as simple as 

Default install: yarn: `....`
Default install: npm: `....` 

in the table?

PatMyron:
@plaindocs good catch on the existing JS one. I'll keep thinking about how to best display all of those, especially the ones with more details on how they choose defaults. What are your thoughts on merging these changes in the meantime?

plaindocs:
@PatMyron sure, sounds good to me. Just link the next PR to this one so we can see the thought process.

PatMyron:
@plaindocs These pages seem to solve complex defaults well enough:

https://docs.travis-ci.com/user/languages/groovy/
https://docs.travis-ci.com/user/languages/java/
https://docs.travis-ci.com/user/languages/scala/

plaindocs:
By removing the "ataglanceness" of it though. I guess it depends a little on how common each of the options is, and finding the right balance between overloading people, and showing enough detail.

PR #2253, opened 2019-01-28
Following the Haskell guide leads to ruby builds

sheyll:
When following the guide to setup a Haskell build, travis mistakes the project with a ruby project.
Setting the `language` field explicitly to `haskell` resolves this issue.


plaindocs:
Thanks!

PR #2254, opened 2019-01-28
naming minimal file examples in all ataglance side panels

PatMyron:
Ignore/squash the git history


plaindocs:
Hey @PatMyron do you fancy fixing the conflict here?  Otherwise looks :100:

PR #2255, opened 2019-01-29
Fix package name for Google Chrome in Xenial

fonglh:
The correct name for the package to be installed instead of causing an error is `google-chrome-stable`.


plaindocs:
Thank you!

PR #2262, opened 2019-02-02
Add extra hints for docker-compose integration

leandromoreira:
Add tip for docker-compose integration with external tools, like coveralls.

I spent a reasonable amount of time [trying to figure out](https://github.com/leandromoreira/nginx-lua-redis-rate-measuring/pull/22) why wasn't my [`.travis.yml`](https://github.com/leandromoreira/nginx-lua-redis-rate-measuring/blob/master/.travis.yml#L16) not sending data for coveralls.

I don't know if this is too specific or not useful but I would be glad to find this documentation close to the `docker-compose` section.

### DIFF
--- a/user/docker.md
+++ b/user/docker.md
@@ -189,7 +189,7 @@ before_install:
 ```
 {: data-file=".travis.yml"}
 
-If you're using `docker-compose` to integrate with externals tools, like coveralls, make sure you repass the required environment variables.
+If you're using `docker-compose` to integrate with externals tools, like coveralls, make sure you repass the required environment variables, here's a simple example of a docker-compose.yml file.
 
 ```yaml
   test:
--- a/user/docker.md
+++ b/user/docker.md
@@ -189,6 +189,26 @@ before_install:
 ```
 {: data-file=".travis.yml"}
 
+If you're using `docker-compose` to integrate with externals tools, like coveralls, make sure you repass the required environment variables.
+
+```yaml
+  test:
+    command: busted -c
+    environment:
+      - TRAVIS=true
+      - CI=true
+      - TRAVIS_JOB_ID=${TRAVIS_JOB_ID}
+      - TRAVIS_BRANCH=${TRAVIS_BRANCH}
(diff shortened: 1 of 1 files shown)


user/docker.md
```diff
@@ -189,6 +189,26 @@ before_install:
...
+      - TRAVIS_JOB_ID=${TRAVIS_JOB_ID}
+      - TRAVIS_BRANCH=${TRAVIS_BRANCH}
+      - TRAVIS_REPO_SLUG=${TRAVIS_REPO_SLUG}
+    build:
+      context: .
+      dockerfile: Dockerfile.test
+    volumes:
+      - ".:/lua/"
+    working_dir: "/lua"
+```
```
plaindocs:
```suggestion
```
{: data-file=".travis.yml"}
```

plaindocs:
This looks relevant to me, @BanzaiMan any thoughts?

BanzaiMan:
I understand the motivation, but I have to ask: Is this not basic `docker compose` usage? By this, I mean, "Everyone who uses `docker compose` should know how to pass environment variables to the container they run."

The example seems to me quite far from "simple". The context of the shared volume `./lua` is unclear, and, in fact, the snippet is incomplete. (`test` is not a valid top-level key in the `docker-compose.yml` file https://docs.docker.com/compose/compose-file/.)

It seems to me that pointing to a more complete command reference is a better idea than giving an example like this.

leandromoreira:
@BanzaiMan thanks for the review, I see your points, I didn't put [a complete example](https://github.com/leandromoreira/nginx-lua-redis-rate-measuring/blob/master/docker-compose.yml) because I thought it would be hard to read but it can be shrink.

And if we just use the textual part of PR? 

> If you're using `docker-compose` to integrate with externals tools, like coveralls, **make sure you repass the required [environment variables](https://docs.docker.com/compose/environment-variables/)**.

Anyway, the repo [`luacov-coveralls`](https://github.com/moteus/luacov-coveralls/commit/435177178b8c5e37fbbd15abe6ebb2a9a1128869) already accept the PR and it can help users that might pass through this "issue".

PR #2264, opened 2019-02-04
Grouping Python examples

PatMyron:


plaindocs:
As these two examples are specifically about the dependency environment variables, we should probably mention those two by name if we're including them into the full example list. As such I'm not sure if it that helpful. What do you think?

PR #2265, opened 2019-02-05
Ruby gem trav3 - API v3 client

danielpclark:
A lightweight abstraction layer for using Travis CI version 3 of their API.  This gem has no dependencies, fully documented API interface methods with examples, and 100% test coverage.

### DIFF
--- a/user/apps.md
+++ b/user/apps.md
@@ -425,6 +425,7 @@ By Sankarsan Kampa
 ## Ruby
 
 - [travis.rb](https://github.com/travis-ci/travis.rb) **(official)**
+- [trav3](https://github.com/danielpclark/trav3) by Daniel P. Clark
 - [TravisMiner](https://github.com/smcintosh/travisminer) by Shane McIntosh
 - [hoe-travis](https://github.com/drbrain/hoe-travis) by Eric Hodel
 - [Knapsack](https://github.com/ArturT/knapsack) by Artur Trzop


DrTorte:
Hey @danielpclark - this is very cool! Super appreciate all the effort put into it. From our end, we'd love to know if there's anything we can do that'll make it easier for you to make further improvements to the client?

danielpclark:
> From our end, we'd love to know if there's anything we can do that'll make it easier for you to make further improvements to the client?

That's a good question.  I've already implemented every publicly documented endpoint of the API with a successful response so the biggest part of the work is done.

## What I can do

Personally one of the things I was considering before releasing a 1.0 version was a more uniform way for the ruby methods to take parameters.  For example some take a parameter for `delete` which may be `true`, some take a keyword argument for deleting which would be `delete: true`, and some take a symbol `:delete`.  Each of these decisions were arrived to at the time based on arriving at the simplest implementation given all the other potential parameters given.  The keyword args are here to stay but I'm leaning towards having a more explicit internal API and changing all of the default `delete = nil` parameters to explicitly expect a `:delete` symbol since this make for much more readable code.

Another thing for me to consider is to examine the responses more closely and see what I missed.  For example I had missed the `@warning` part of the JSON response when using a follow feature on `@href` which simply occured because I have a default url option of limiting the response with `limit` which does not apply in this case.  I'll have that amended shortly.

I'll be reviewing the raw responses I have recorded to see if there are any more improvements I can make on my end.

## What you can do

First of all let me say thank you all for the API that you've built it was quite a nice experience for me to work with.  I appreciate all the work you've already put into it. :+1:  :sunglasses: :heart:

As for what can help me improve the client I'm not sure there's much left.  I think a few improvements to your own API documentation and I can generally improve my own  documentation and that would helpful.  Other than that I suppose using it and opening issues or feature suggestions would be the best way for better improvement.

A little more work on the documentation side would be helpful for others using the API.  There were a few resources the showed all the proper API endpoints but for the written descriptions it showed up like:

```
Document `resources/beta_migration_request/overview` not found.
```

:construction: But the details about those items were simple for me to write in myself so they're not a big issue.

Come to think of it I don't remember seeing the [Beta migration_request](https://developer.travis-ci.com/resource/beta_migration_request#beta%20migration_request) resource before… are you all still adding resources? :new: :question:   I need to add this one in.

Some resources say they have no endpoint _(which is a good thing)_ while others say nothing and provide no API examples (such as **message**).

On the **requests** resource it provides an example to “override a commit message” but when I tried providing an old commit number as one of the parameters it did not change the commit message in an old TravisCI test but rather overwrote the most recent with a new build with the commit message.  So it seems to just change the most recent commit name while leaving the previous build, with the same commit number, with the previous git name in the build history.

One feature I would really have liked to have is a dummy organization, user, repository to test every API endpoint against.  It should still require credentials when the resource itself would and it would be much nicer to have specific dummy credentials to use rather than me potentially **risking** putting my API TOKEN into my git repo.  I had to be extra cautious about that for the tests.

In your API documentation, many times, it is unclear whether or not an API token is required.  Most of the time I can simply be intuitive about it but without it written it takes more work for someone writing a client to discover what to note about token requirements in documentation.

One initial rookie mistake I made was when reading the documentation for **lint** where it says “The request body can contain…” I overlooked the fact that **body** was an actual entity I was supposed to use and thought it was just a general statement of fact.  I had been familiar with a response body because the response always has a `body` method on it from which I can get the text of content, but I didn't make the connection in my head.  Something as simple as putting the word “body” in bold would have helped, or a `curl` example showing usage would clear this up for people who aren't accustomed to writing raw REST requests.

I found the **enc_key** API endpoint not clear enough in how exactly to create a key that TravisCI API would accept.  It was at this point that I discovered you even had an official client and read the source code to see how to generate an acceptable output your server would accept.  If you could document how one might do this with `curl` & `gpg` that would be awesome.  Also I'm curious as to whether the keys have to be of `2048` strength or if `4096` would work?  That should be in the API documentation in my opinion as well.  Once I know this I can add it to my own projects info.

## TLDR; Summary answer

Your API is well designed and I like it very much.  Other than using the gem and opening issues/feature requests there isn't too much, that I can see, to improve upon at this time.  But one big area that would help is improving the online API documentation.

plaindocs:
Thanks @danielpclark and thanks also for the detailed reply. @DrTorte and I will process that lot and see what we can do. :-)

PR #2266, opened 2019-02-05
Update golang docs

meatballhat:
per PR https://github.com/travis-ci/travis-build/pull/1636

### DIFF
--- a/user/languages/go.md
+++ b/user/languages/go.md
@@ -12,31 +12,31 @@ swiftypetags:
 
 <aside markdown="block" class="ataglance">
 
-| Go                                          | Default                                   |
-|:--------------------------------------------|:------------------------------------------|
-| [Default `install`](#dependency-management) | `go get -t -v ./...`                         |
-| [Default `script`](#default-build-script)   | `make` or `go test`                       |
-| [Matrix keys](#build-matrix)                | `go`, `env`                               |
-| Support                                     | [Travis CI](mailto:support@travis-ci.com) |
+| Go                                          | Default                                                      |
+|:--------------------------------------------|:-------------------------------------------------------------|
+| [Default `install`](#dependency-management) | `travis_install_go_dependencies <go-version> [gobuild-args]` |
+| [Default `script`](#default-build-script)   | `travis_script_go {gobuild-args}`                            |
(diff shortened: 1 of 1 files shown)


user/languages/go.md
```diff
@@ -50,19 +50,28 @@ handled by [gimme](https://github.com/travis-ci/gimme).
...
+- 1.11.x
+- master
 ```
 {: data-file=".travis.yml"}
 
 
+## Go Modules
+
+If Go 1.11+ is specified and either `GO111MODULE=on` is set or a `go.mod` file
+is present, then it is assumed that [Go
```
BanzaiMan: `go.mod` should exist at the repository root, correct?

meatballhat: @BanzaiMan yes, but I didn't feel that explaining the rules here was appropriate given that they aren't _our_ rules

user/languages/go.md
```diff
@@ -72,26 +81,23 @@ go_import_path: example.org/pkg/foo
...
-  ```
-  go get -t -v ./...
-  ```
+The default `install` step of `travis_install_go_dependencies <go-version>
+[gobuild-args]` will behave differently depending on the Go version specified,
+as well as the presence of certain environment variables and file paths.
 
-* if go version is older than `1.2`
+If the Go version and environment are detected to support Go 1.11+ Modules or Go
+1.5+ vendoring, then we skip to Makefile detection. In all other cases, file
```
BanzaiMan: Should `Makefile` be stylized?

user/languages/go.md
```diff
@@ -72,26 +81,23 @@ go_import_path: example.org/pkg/foo
...
+paths are checked for use of [godep](https://github.com/tools/godep) (see below).
 
-  ```
-  go get ./...
-  ```
+If a Makefile is present by any of the following names, then no further actions
```
BanzaiMan: Ditto

user/languages/go.md
meatballhat: I'm good with either! I didn't in this case because we check for 4 different known file names.

BanzaiMan: Ah, I see. I suppose we can leave it as is.

PR #2273, opened 2019-02-07
making ataglance minimal examples as minimal as possible

PatMyron:
Python is the only language without a default script in its at-a-glance side panel. Should Python's script default to `pytest`?

### DIFF
--- a/user/languages/elixir.md
+++ b/user/languages/elixir.md
@@ -19,8 +19,6 @@ Minimal example:
 
 ```yaml
 language: elixir
-elixir: '1.5.2'
-otp_release: '19.0'
 ```
 {: data-file=".travis.yml"}
 
--- a/user/languages/elm.md
+++ b/user/languages/elm.md
@@ -24,8 +24,6 @@ Minimal example:
 
 ```yaml
   language: elm
-  elm:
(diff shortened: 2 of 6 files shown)


plaindocs:
Nope, python is special like that.

> If you do not provide a script key in a Python project, Travis CI prints a message (“Please override the script: key in your .travis.yml to run tests.”) and fails the build.

PatMyron:
@plaindocs Sorry, that was unclear. I meant should the actual behavior itself be changed to run `pytest` as the default script for Python?

plaindocs:
I know enough python projects using makefiles that ancecdatally it doesn't seem default enough. But no reason for it not to detect Makefile and pytest like other languages.

PatMyron:
@plaindocs makes sense. Fixed the merge conflicts

plaindocs:
Perfect! merging. 

I've opened the Python thing for wider discussion at https://travis-ci.community/t/detect-python-defaults-opinions-wanted/2236 . Cheers!

PatMyron:
@plaindocs could be worth analyzing public Github repositories to get a better understanding of the breakdown. I might look into it sometime

PR #2279, opened 2019-02-13
Mention snap addon in snap deploy instructions

abitrolly:


plaindocs:
Thanks for the PR. As we're specifically talking about deployments here I don't think adding this makes sense.

Were you looking for info on installing snaps and ended up on this page?

PR #2291, opened 2019-02-27
Equals comparison on tag is based on its name.

senakafdo:
As explained in [here](https://travis-ci.community/t/somewhat-misleading-documentation-on-conditions/2468), the equals comparison on `tag` checks if it had a name `true`. The term true also represents the boolean value `true` and easily confuses a user. Therefore it would be better to make the documentation unambigous by having an example such as `tag = bar`.


BanzaiMan:
Thank you!

PR #2292, opened 2019-03-06
Julia: document JULIA_PROJECT env var

fingolfin:
We had a hard time understanding what was going on until we realized that `language: julia` forces a specific value for `JULIA_PROJECT` to be set.


plaindocs:
@ararslan, @staticfloat, and @StefanKarpinski does this look right to you?

plaindocs:
https://discourse.julialang.org/t/handling-multiple-versions-of-julia/14035/10

plaindocs:
Thanks!

PR #2294, opened 2019-03-08
Add some extra language around forks

muizzk:
@cotsog this is in relation to ZD: 4906.
cc/ @BanzaiMan


plaindocs:
Nice one @muizzk !

PR #2312, opened 2019-03-23
Remove Python 3.4 from this example because it is EOL

cclauss:
Removes Python 3.4 from one example because that release has reached its end of life.
* https://devguide.python.org/devcycle/#end-of-life-branches

This PR does _not_ remove Python 3.4 from line 48 where it appears with other EOLed versions of Python (2.6 and 3.3) that Travis CI continues to make available.


plaindocs:
Thanks!

PR #2321, opened 2019-04-04
Fix over indentation on release example

foxyblue:
The indentation on the "release to a different pypi index"
example was over indented compared to the other examples.


plaindocs:
Thanks!

PR #2330, opened 2019-04-12
Update multi-os.md

aldot:
Remove invocations of legacy test(1) names;
[ "foo" = "bar ]
is the proper SUSv3 onward spellig in sh.
No double brackets, no double =.


plaindocs:
Thanks

Issue #2233, opened 2019-01-16
Adapt Xvfb instruction docs to different init systems in our images


iriberri:
At https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-xvfb-to-run-tests-that-require-a-gui, we're documenting the new `services: xvfb` method but we still mention starting xvfb using the `/etc/init.d/xvfb` file, which does no longer exist in the Xenial env that uses systemd.

I see that these two sections:

* https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-xvfb-directly
* https://docs.travis-ci.com/user/gui-and-headless-browsers/#starting-a-web-server

still mention `/etc/init.d/xvfb`. The "Starting a web server" one seems easily adaptable by adding to it to new `services: xvfb` approach. For "Using Xvfb directly", should we split the section with the `upstart`/`systemd` approaches? What do you think @joepvd?

Code context: https://github.com/travis-ci/travis-build/blob/664f39d1c7da26e145a85974d71d7972bf2588fc/lib/travis/build/appliances/services.rb#L82-L90

PR #2338, opened 2019-04-23
Add workaround for xvfb errors

muizzk:
@cotsog for ZD 6580.


lechiffre10:
I think the issue here is the gap between Enterprise and Hosted and not necessarily the documentation on the Hosted side ( Display is set properly on Xenial hosted) and we need to raise this in Enterprise Sync. Also maybe @cotsog might want to chime in but we need to find a way to communicate that the Hosted documentation doesn't necessarily apply to the Enterprise side considering the gap?

ssabrii:
I think the issue here is the gap between Enterprise and Hosted and not necessarily the documentation on the Hosted side ( Display is set properly on Xenial hosted) and we need to raise this in Enterprise Sync. Also maybe @cotsog might want to chime in but we need to find a way to communicate that the Hosted documentation doesn't necessarily apply to the Enterprise side considering the gap?

user/gui-and-headless-browsers.md
```diff
@@ -72,6 +72,14 @@ services:
 ```
 {: data-file=".travis.yml"}
 
+> Sometimes it may be necessary to explicitly set the DISPLAY on Travis CI Enterprise. You can achieve this with the following snippet:
+
+```yaml
+before_script:
+  - "export DISPLAY=:99.0"
```
cotsog: Do you know if it's specific to `dist: xenial`? If so, I would put the full recipe:

```yaml
dist: xenial
services:
  - xvfb

before_script:
  - "export DISPLAY=:99.0"
```

muizzk: No idea. @ssabrii please can you confirm? Thanks

muizzk:
@cotsog updating as per discussion. Thanks

plaindocs:
Also does this fully address #2233  ?

user/gui-and-headless-browsers.md
Litlangelok: qguz9ZVRYsdtfzwnm6qX

Litlangelok: > Do you know if it's specific to `dist: xenial`? If so, I would put the full recipe:
> 
> ```yaml
> dist: xenial
> services:
>   - xvfb
> 
> before_script:
>   - "export DISPLAY=:99.0"
> ```

Litlangelok: > Do you know if it's specific to `dist: xenial`? If so, I would put the full recipe:
> 
> ```yaml
> dist: xenial
> services:
>   - xvfb
> 
> before_script:
>   - "export DISPLAY=:99.0"
> ```

Litlangelok (changes requested):

PR #2340, opened 2019-04-25
Validate yaml examples

svenfuchs:
This is meant to make all `.travis.yml` example snippets parse and validate.

I have kept individual changes in separate commits, so it is easier to reject some of them. Some changes fix broken indentation or broken formats, but most are rather stylistic, e.g. enclosing `[string]` into quotes as `"[string]"` (because YAML parses the former into an actual array, not a string, as the example tries to suggest. These stylistic choices are certainly debatable.


plaindocs:
I'm gonna merge in, and we can fixup / discuss any stylistic issues another time if they crop up. Thanks!

PR #2350, opened 2019-05-11
Python tests on multiple OSes: Fix pip on Windows

cclauss:
Without this change, pip will fail with permission errors on Windows.


Pezi777:
Thank you cclauss :-)

PR #2351, opened 2019-05-11
Update deployment.md: fix regular expression condition

alanpoulain:
The condition was not working because of the missing parentheses.


Pezi777:
Thank you alanpoulain :-)

PR #2352, opened 2019-05-13
Update s3.md

nishimura-yuki:
I met an error same as this issue(https://github.com/travis-ci/dpl/issues/237) last week.
Finally, I edited S3 block public access then it solved.

I think that S3 block public access is all deny by default and this document only refer for Bucket Policy at the moment .
So I think it need to refer S3 block public access then I commit this document.

My English is not so good. So I'll leave it you revise my sentence.


user/deployment/s3.md
```diff
@@ -113,6 +113,9 @@ This bucket policy grants the public read permission:
 }
 ```
 
+Also, if your bucket enables [S3 Block Public Access](https://aws.amazon.com/jp/blogs/aws/amazon-s3-block-public-access-another-layer-of-protection-for-your-accounts-and-buckets/),
```
BanzaiMan: The URL appears to point to a `jp` doc. The document itself is written in English, but links from this page _may_ (unconfirmed) not be. It is probably best to point to what I believe is the equivalent: https://aws.amazon.com/blogs/aws/amazon-s3-block-public-access-another-layer-of-protection-for-your-accounts-and-buckets/

user/deployment/s3.md
```diff
@@ -113,6 +113,9 @@ This bucket policy grants the public read permission:
...
+you should disable some of these settings.
```
BanzaiMan: What are the settings, more specifically?

PR #2362, opened 2019-05-25
Update and rename user/common-build-problems.md to  пользователь/comm…

korrectiv:
…on-build-problems.md


BanzaiMan:
I don't think this is valid. If you have further thoughts, do get in touch. Thanks.

PR #2371, opened 2019-06-04
Update python.md for info on 2.6 for Xenial

ssabrii:


cotsog:
I think we want to keep language specific docs version independent.

Supported Python versions on Xenial are listed here: https://docs.travis-ci.com/user/reference/xenial/#python-support

WDYT?

PR #2377, opened 2019-06-12
Add missing Encryption Key steps

dannygriffin000:
This is copied over from the operations manual page to make sure that users include the encryption key in their backups as well.


Pezi777:
It is already in the documentation

PR #2383, opened 2019-06-17
Update build-stages.md

j0k3r:
I found a small typo in the build stage doc. I think `d` should be `deploy` in that paragraph.

### DIFF
--- a/user/build-stages.md
+++ b/user/build-stages.md
@@ -144,7 +144,7 @@ the test stage has completed successfully.
 
 > Each job included in `jobs.include` inherits the first value of the array
 that defines a matrix dimension.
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Hello, there. Thanks for the PR. The rendering is not great, but this is meant to be read as "…the included job inherits…". Would that make sense? We also appreciate your suggestions on how to improve the rendering and readability.

Thanks!

j0k3r:
Oh, include**d**, didn't read it that way :nerd_face:

BanzaiMan:
As I said, if you have suggestions on how to increase clarity here, we are happy to look at them. 😄

PR #2393, opened 2019-06-24
Explain in more detail how GitHub Releases expects tags

BanzaiMan:


BanzaiMan:
This seems to confuse some users, so let's try explaining it in more detail. https://travis-ci.community/t/tagging-a-github-release-results-in-an-error-and-the-release-publishes-without-the-file/3819

PR #2395, opened 2019-06-25
Add workspaces documentation

BanzaiMan:


user/using-workspaces.md
```diff
@@ -0,0 +1,152 @@
...
+
+# Workspaces (Beta)
+
+> Workspaces are currently in beta testing, and may change without a prior
+> notice.
+{: .beta}
+
+Workspaces allow jobs _within_ a build to share files.
+They are useful when you want to use build artifacts from a previous job;
+for example, you might create a cache that can be used in multiple jobs later.
```
Pezi777:
```suggestion
for example, you create a cache that can be used in multiple jobs later.
```

user/using-workspaces.md
```diff
@@ -0,0 +1,152 @@
...
+
+While any build can theoretically use workspaces, they are most useful when
```
Pezi777:
```suggestion
Theoretically, any build can use workspaces, but they are most useful when
```

user/using-workspaces.md
```diff
@@ -0,0 +1,152 @@
...
+there is a clear timeline of when they are used.
+For this reason, we recommend using workspaces when you use
+[build stages](/user/build-stages).
+
+Roughly speaking, you create a workspace in one stage, then use it in
```
Pezi777:
```suggestion
> Note that it is best to create a workspace in one stage and then use it in
```

user/using-workspaces.md
```diff
@@ -0,0 +1,152 @@
...
+
+In this example:
+  1. we run two jobs in the "Build binaries" stage,
+     each producing uploading binaries.
+  1. in the subsequent "Deploy" stage, we fetch the workspaces
+     produced by the jobs in the previous stage, and deploy it
+     using a custom script.
+
+## Workspaces and concurrency
+Note that workspaces work best if there is not race condition in
```
Pezi777:
```suggestion
Note that workspaces work best if there is no race condition set while
```

user/using-workspaces.md
```diff
@@ -0,0 +1,152 @@
...
+```
+{: data-file=".travis.yml"}
+
+even if the consumer is running on a Linux VM.
+
+## How workspaces differ from caches
+
+It is worth reiterating that the workspaces are meant for sharing files
+within a build.
+For files that you want to share across builds, use
```
Pezi777:
```suggestion
For files you want to share across builds, use
```

user/using-workspaces.md
```diff
@@ -0,0 +1,152 @@
...
+[caches](/user/caching).
+
+### Workspaces and caches are considered independently
+
+It is possible to include a single file in the cache and workspaces.
+
+## How restarting a portion of a build affects workspaces
+Note that workspaces are built and consumed by jobs in a builld.
```
Pezi777:
```suggestion
Note that workspaces are built and consumed by jobs in a build.
```

user/using-workspaces.md
```diff
@@ -0,0 +1,152 @@
...
+    name: $TRAVIS_OS_NAME
+    paths: …
+```
+{: data-file=".travis.yml"}
+
+it will have to be consumed by
+
+```yaml
+workspaces:
+  use: $TRAVIS_OS_NAME
```
plaindocs: Does this need escaping in that case? `\$TRAVIS_OS_NAME`?

user/using-workspaces.md
```diff
@@ -0,0 +1,152 @@
...
+  use: linux
+```
+{: data-file=".travis.yml"}
+
+even if the consumer is running on a Linux VM.
+
+## How workspaces differ from caches
+
+It is worth reiterating that the workspaces are meant for sharing files
+within a build.
```
plaindocs:
```suggestion
within the same build.
```

might be clearer? What do you think?

user/using-workspaces.md
```diff
@@ -0,0 +1,152 @@
...
+For files that you want to share across builds, use
+[caches](/user/caching).
+
+### Workspaces and caches are considered independently
+
+It is possible to include a single file in the cache and workspaces.
+
+## How restarting a portion of a build affects workspaces
+Note that workspaces are built and consumed by jobs in a builld.
+If you restart parts of the build in some arbitrary order, workspaces may
```
plaindocs: What about replacing this all with.

> If you restart some builds within a job, and your workspace isn't working like you expect it to, try restarting the the entire build.

user/using-workspaces.md
```diff
@@ -0,0 +1,152 @@
...
+`workspaces.create`, and by the value when it is used with
+`workspaces.use`.
+
+In the example above, the workspace named `ws1` is created by the only
+job in the `warm_cache` stage.
+The workspace is subsequently consumed in the `use_cache` stage.
+
+## Example configurations
+
+### Using multiple workspaces in a build
```
plaindocs: These headings feel off.

Maybe just 

## Minimal workspace example

## Multiple workspaces example

user/using-workspaces.md
plaindocs: How about

> So you can clearly see when a workspace is created and used, we recommend using workspaces with [build stages](/user/build-stages), as shown in the [following](#configuration) [examples](#other-confugurations).

(fixing links as needed to example headings below)

user/using-workspaces.md
Pezi777: :+1:

user/using-workspaces.md
BanzaiMan: No escaping is necessary here.

PR #2405, opened 2019-07-02
Update default ubuntu image to xenial in doc

bagerard:
Default image recently got switched from trusty to xenial


Pezi777:
Thank you bagerard :-)

PR #2412, opened 2019-07-05
npm  default caching

anarosas:


user/caching.md
```diff
@@ -135,7 +135,16 @@ podfile: path/to/Podfile
 
 ### npm cache
 
-For caching with `npm`, use:
+> Please note that npm is now cached by default on Travis CI
```
plaindocs: Can we say as of Xenial, or as of July 2019 or something more specific?

user/caching.md
```diff
@@ -135,7 +135,16 @@ podfile: path/to/Podfile
...
+
+To disable npm caching, use:
+
+```yaml
+cache:
+  npm: false
+```
+
+To explicitly cache `npm`, use:
```
plaindocs: Is it worth deleting this line and the following block? Or is there a use case for explicitly caching npm?

user/languages/javascript-with-nodejs.md
```diff
@@ -135,7 +135,14 @@ as specified in your lock file.
 
 #### Caching with `npm`
 
-You can cache your dependencies with
+`npm` is now cached by default, in case you want to disbale it, please add the following to your `.travis.yml`:
```
plaindocs:
```suggestion
`npm` is now cached by default, in case you want to disable it, please add the following to your `.travis.yml`:
```

user/languages/javascript-with-nodejs.md
```diff
@@ -135,7 +135,14 @@ as specified in your lock file.
...
+
+```yaml
+cache:
+  npm: false
+```
+
+To explicitly cache your dependecies:
```
plaindocs:
```suggestion
To explicitly cache your dependencies:
```

user/caching.md
anarosas: yes, we can totally say `as of July 2019`

user/caching.md
BanzaiMan: I think we can have this explicit statement here. It is unlikely, but we _could_ switch the default behavior back to not caching, in which case the explicit `cache` will save the users.

anarosas: Just had a chat with @BanzaiMan, we will leave it since it might be useful for some users 

From slack: 
>hasari 12:09 PM I think leaving it there is fine; doing things implicitly can be a trap, when that default changes for some reason again.
> (relying on Ruby being the default, Precise/Trusty/Xenial/Bionic being default, etc., etc.)
> when the default can (or did) change, the being explicit is a good thing.

PR #2416, opened 2019-07-08
Update Ruby to something current

plaindocs:
Testing testing. Will need to test Heroku also.

FAO @Pezi777


plaindocs:
Seems ok to me.

PR #2421, opened 2019-07-11
Clean up public key retrieval instructions

BanzaiMan:
Instead of a wall of text, use a table to visually break up and present information in a more concise way.

This is a follow-up to #2420.

### DIFF
--- a/user/encryption-keys.md
+++ b/user/encryption-keys.md
@@ -12,16 +12,16 @@ A repository's `.travis.yml` file can have "encrypted values", such as [environm
 
 ## Encryption scheme
 
-Travis CI uses asymmetric cryptography. For each registered repository, Travis CI generates an RSA keypair. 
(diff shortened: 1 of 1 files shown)


user/encryption-keys.md
```diff
@@ -12,42 +12,76 @@ A repository's `.travis.yml` file can have "encrypted values", such as [environm
...
+exists, and the API version you are using.
+
+Furthermore, the request may require authorization via the `Authorization: token`
+header depending on the repository's location and visibility, as well as
+the API version used.
+
+<table>
+  <caption><tt>Authorization</tt> header requirement</caption>
+  <tr>
+    <th rowspan="2">repository visibility and location</th>
```
Pezi777:
```suggestion
    <th rowspan="2">Repository visibility and location</th>
```

PR #2432, opened 2019-07-17
Note PHP 5.4 and 5.5 now require Precise

donatj:
I'm not 100% sure when the change happened but 5.4 and 5.5 now give the following:

```
$ phpenv global 5.4
rbenv: version `5.4' not installed
```

It can be fixed by using Precise so I've noted that in the docs.


Pezi777:
@donatj Thanks so much for sharing that with us. Do you maybe have a (Travis CI) link showing this behavior? This would help us a lot! :-)

donatj:
Yup! https://travis-ci.org/donatj/mock-webserver/builds/560196635 Shows it not working with the error above, and https://travis-ci.org/donatj/mock-webserver/builds/560200907 shows them working after I switched to Precise

MariadeAnton:
Thanks so much for your contribution and for the example links @donatj! 

I suspect this is due to your repository using the new Linux default, Xenial which doesn't have 5.4 or 5.5 available. https://travis-ci.org/donatj/mock-webserver/jobs/560196636#L9 

For more context on the default update, you can check this post: https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

To confirm, I've opened https://github.com/donatj/mock-webserver/pull/19 :) 

Could you update the pull request to mention that these versions require either Trusty or Precise? Thanks again!

donatj:
> Could you update the pull request to mention that these versions require either Trusty or Precise? Thanks again!

@MariadeAnton I wasn't exactly sure how to word it as I believe 5.2 and 5.3 still require Precise - I added another small section, I hope this works for you.

PR #2440, opened 2019-07-23
adapting provider page for dpl v2

Pezi777:


svenfuchs:
`netlify` is now documented in `v2`

PR #2446, opened 2019-07-26
Update the common-build-problems.md

tmkasun:
Include the information about unexpected cache missing due to the usage of exit command in the script.

### DIFF
--- a/user/common-build-problems.md
+++ b/user/common-build-problems.md
@@ -610,3 +610,36 @@ jobs:
 
 
 This creates only one job,  _Peanut Butter and Bread_ under the stage named _Breakfast_ as you have defined. It is important to note that in YAML, the `-` symbol is used to create a list of items and the earlier example creates a list of 2 items, while you actually wanted 1. You can read more on [How to define Build Stages](/user/build-stages/#how-to-define-build-stages) and YAML lists syntax in the official [documentation](https://yaml.org/spec/1.2/spec.html#id2759963).
+
+## Cahce files not getting preserverd after build success
+
+Check the `scrip` section in your .travis.yaml configuration file. You might have used `exit` command in the given script
+
+i:e
+
+```yaml
+language: java
+dist: trusty
+
+script: mvn clean install -Dmaven.test.skip -q -B -V | grep -v DEBUG; exit "${PIPESTATUS[0]}";
+
+cache:
+  directories:
+   - $HOME/.m2
+```
+
+User commands are executed in the worker bash process with eval, so this `exit` quits this process and terminates your build before it reaches the cache stage.
+To overcome this issue , use the `test` command as below , instead the `exit` command to propergate the exit code.
+
+i:e
+
+```yaml
+language: java
+dist: trusty
+
+script: mvn clean install -Dmaven.test.skip -q -B -V | grep -v DEBUG; test ${PIPESTATUS[0]} -eq 0;
+
+cache:
+  directories:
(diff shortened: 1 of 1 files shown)


native-api:
This is in relation to https://travis-ci.community/t/java-maven-build-completes-successfully-but-m2-local-repository-is-not-getting-cached/4380

user/common-build-problems.md
```diff
@@ -610,3 +610,36 @@ jobs:
 
 
 This creates only one job,  _Peanut Butter and Bread_ under the stage named _Breakfast_ as you have defined. It is important to note that in YAML, the `-` symbol is used to create a list of items and the earlier example creates a list of 2 items, while you actually wanted 1. You can read more on [How to define Build Stages](/user/build-stages/#how-to-define-build-stages) and YAML lists syntax in the official [documentation](https://yaml.org/spec/1.2/spec.html#id2759963).
+
+## Build terminates early if I use `exit` in my script
+
+Check the `scrip` section in your .travis.yaml configuration file. You might have used `exit` command in the given script
```
BanzaiMan: It should be ``` `script` ```.

Additionally, these two sentences are basically pointing out the same idea. Drop one.

user/common-build-problems.md
```diff
@@ -610,3 +610,36 @@ jobs:
...
+
+i:e
```
BanzaiMan: You probably meant "for example".

---
### Aside
The phrase "_i.e._" (not "_i:e_") is shorthand for _id est_, meaning "in other words" to introduce another equivalent idea afterwards, and it does not fit here.

If you do mean "for example", the abbreviation is "_e.g._", but it should not start the sentence. Just write out the words.

user/common-build-problems.md
```diff
@@ -610,3 +610,36 @@ jobs:
 
 
 This creates only one job,  _Peanut Butter and Bread_ under the stage named _Breakfast_ as you have defined. It is important to note that in YAML, the `-` symbol is used to create a list of items and the earlier example creates a list of 2 items, while you actually wanted 1. You can read more on [How to define Build Stages](/user/build-stages/#how-to-define-build-stages) and YAML lists syntax in the official [documentation](https://yaml.org/spec/1.2/spec.html#id2759963).
+
+## Build terminates early if I use `exit` in my script
```
BanzaiMan: This "problem" statement seems to anticipate the answer already. The user would be looking for the statement that matches the symptom they are seeing in the FAQ, so I suggest dropping `exit`.

## Job terminates prematurely

user/common-build-problems.md
```diff
@@ -610,3 +610,36 @@ jobs:
...
+
+i:e
+
+```yaml
+language: java
+dist: trusty
+
+script: mvn clean install -Dmaven.test.skip -q -B -V | grep -v DEBUG; exit "${PIPESTATUS[0]}";
+
+cache:
```
BanzaiMan: `cache` directive is not essential to this example, so I suggest dropping it.

user/common-build-problems.md
```diff
@@ -610,3 +610,36 @@ jobs:
...
+
+## Build terminates early if I use `exit` in my script
+
+Check the `scrip` section in your .travis.yaml configuration file. You might have used `exit` command in the given script
+
+i:e
+
+```yaml
+language: java
+dist: trusty
```
BanzaiMan: This is not necessary. If you insist, please use `xenial` instead.

PR #2448, opened 2019-07-26
Add a note about MySQL being EOL on Trusty

cotsog:


user/database-setup.md
```diff
@@ -102,30 +102,9 @@ before_install:
 
 ### MySQL 5.7
 
-MySQL 5.7 is the default on the Xenial and Bionic image.
-On Trusty, you can install MySQL 5.7 by adding the following lines to your `.travis.yml`:
+> Since July 21st 2019, MySQL 5.7 is not supported on Ubuntu Trusty (14.04) anymore. See [MySQL Product Support EOL Announcements](https://www.mysql.com/support/eol-notice.html) and [this post](https://forums.mysql.com/read.php?11,677237,677268#msg-677268) in the MySQL Forums.
```
MariadeAnton: Can we move the note about where it's not available after we say what's available? Other than that, this looks great, @cotsog - thanks for adding this so quickly

cotsog: Done! 😄

PR #2459, opened 2019-08-01
Add workaround if debug vm crashes

ssabrii:


BanzaiMan:
We can give a few more hints.

## Show what commands actually run

Since are bash functions, you can ask what they do:

```sh-session
$ type travis_run_script
travis_run_script is a function
travis_run_script ()
{
    travis_cmd wget\ https://github.com/sormuras/bach/raw/master/install-jdk.sh --echo --timing;
    travis_result $?;
    travis_cmd which\ install-jdk.sh --echo --timing;
    travis_result $?;
    travis_cmd set\ -x --echo --timing;
    travis_result $?;
    travis_cmd source\ install-jdk.sh --echo --timing;
    travis_result $?;
    :
}
```

This goes into some internals, but `travis_cmd` basically executes the string argument (with escaped white spaces in the example above) and adds some decorations so that the output looks nice. In the debug sessions, you can simply run the string argument (unescaped) instead.

## Check `bash` options
Another common cause of unexpected debug session termination is that at some point the [`errexit` option](https://www.tldp.org/LDP/abs/html/options.html#OPTIONSREF) is set (`set -e` or `set -o errexit`).  

You can confirm this with `echo $-` and check for `e` in the output:

```sh-session
$ echo $-
himBH
$ set -e
$ echo $-
ehimBH
```
With this option set, any command that exits with nonzero status will terminate the build (and the debug session, if it's running). You can clear this option with `set +e`; this may allow debug sessions to continue.

ssabrii:
oh wow, this is awesome. Thanks, @BanzaiMan I'll make some changes to the PR based on this info!

PR #2467, opened 2019-08-07
Update build-matrix.md

prokopsimek:
It didn't make sense before. It was the same as is written on line 144.

### DIFF
--- a/user/build-matrix.md
+++ b/user/build-matrix.md
@@ -148,7 +148,7 @@ env:
 matrix:
   exclude:
     - rvm: 1.9.3
-      env: DB=mongodb SUITE=all # not 'env: DB=mongodb  SUITE=all' or 'env: SUITE=all DB=mongodb'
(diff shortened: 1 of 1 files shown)


MariadeAnton:
Thanks for your PR @prokopsimek! 

I think we were trying to showcase that one should use just one space in between, not two, but clearly that's not well showcased, rather than removing the extra env var after the double spacing, we could remove that example entirely and write the one-blank-space rule. What do you think?

 @Pezi777  would you mind taking a look? Thank you!

prokopsimek:
@MariadeAnton Hi, I thought it but how about my change? Would it work or not? :)

MariadeAnton:
It does not convey that one should use a single space between env vars, but it does outline a similar example than the one documented right before, so I think it's an improvement, pending us fully reviewing this page :) 

Thank you again @prokopsimek !

PR #2472, opened 2019-08-08
Type in 'escaping certain symbols' example

danieljhegeman:
The string being encrypted had a `B` where there should have been a `b`


BanzaiMan:
A good catch. Thank you!

PR #2477, opened 2019-08-13
Warn about old YAML version right upfront in the tutorial.

mk-pmb:
It took me about half an hour to find out the reason for Travis's
parsing failure was my header line declaring YAML 1.2.

### DIFF
--- a/user/tutorial.md
+++ b/user/tutorial.md
@@ -41,6 +41,9 @@ To start using Travis CI, make sure you have:
    The defaults for Ruby projects are `bundle install` to [install dependencies](/user/job-lifecycle/#customizing-the-installation-phase),
    and `rake` to build the project.
 
+    * NB: If you want to use a YAML header line, make it `%YAML 1.1`.
+      Travis cannot currently parse newer versions of YAML.
+
 5. Add the `.travis.yml` file to git, commit and push to trigger a Travis CI build:
 
    > Travis only runs builds on the commits you push *after* you've added a `.travis.yml` file.


user/tutorial.md
```diff
@@ -41,6 +41,9 @@ To start using Travis CI, make sure you have:
    The defaults for Ruby projects are `bundle install` to [install dependencies](/user/job-lifecycle/#customizing-the-installation-phase),
    and `rake` to build the project.
 
+    * NB: If you want to use a YAML header line, make it `%YAML 1.1`.
```
BanzaiMan: This caution about the YAML version probably does not belong here. The tutorial is meant to be followed step by step, and the YAML document above is YAML 1.1-compliant.

mk-pmb: Of course you're technically right. It's a matter of how to deal with real users not obeying your expectation, scrolling down until they find something that seems to fit their search. Would you prefer some of their time wasted as the rightful punishment, or do you make your docs to help them get their job done?

BanzaiMan: By this reasoning, we would be including the complete reference of the `.travis.yml` structure.

This statement belongs in our documents, but not in the tutorial.

mk-pmb: Ok, thanks for the explanation then. It's a great project nonetheless!

mk-pmb: Actually, I just checked again to see how bad a reader I must have been. But in current master, even when reading from the top, I can't find information about what file format the `.travis.yml` is meant to be. The `.yml` obviously isn't enough of a hint if only certain versions of YAML are supported. Let's at least have a link to the reference there.

PR #2486, opened 2019-08-17
Fix documentation on using postgresql under Bionic

nanaya:
Fixing error mentioned here: https://travis-ci.community/t/postgresql-fails-to-start/4558

This one is just result of guesswork and trial and error though so I don't actually know how to fix this correctly.

Sorry, force-pushed because the patch looked a bit wrong (originally done using github editor).


user/database-setup.md
```diff
@@ -116,6 +116,14 @@ services:
 ```
 {: data-file=".travis.yml"}
 
+In Ubuntu Bionic, you also need to specify the version like so:
+
+```yaml
+addons:
+  postgresql: 10
```
Pezi777:
```suggestion
  postgresql: 11
```

nanaya:
I just checked again and it looks like this isn't need anymore? It sets up v10 by default. Of course other versions are available but it's already documented.

Pezi777:
as it is no longer necessary to set the postgres version explicitly, I will close this PR

PR #2496, opened 2019-08-26
Removed old parameters that are no longer used

gmegidish:


Pezi777:
thanks so much for updating, @gmegidish!

PR #2514, opened 2019-09-11
Use API token instead of user / password

zero323:
Since July PyPI supports [token based authentication](https://pypi.org/help/#apitoken), and it is now recommended, when possible.

This PR changes references to user and password with `__token__` and API token respectively, and moves user / password  specific notes at the end of the paragraph.


zero323:
Thank you @MariadeAnton @Pezi777!

PR #2523, opened 2019-09-18
Add Python (and potentially other) language archive versions data

BanzaiMan:


BanzaiMan:
Netlify preview site does not contain the resources necessary, so you can't see it. It's deployed to the staging site instead:

https://docs-staging-travis-ci-com.herokuapp.com/user/languages/python/#python-versions

BanzaiMan:
We can probably add some helpful text, and consider different placements for this piece of information, which could break up the flow of information in this document.

user/languages/python.md
```diff
@@ -241,3 +248,30 @@ For a real world example, see [getsentry/sentry](https://github.com/getsentry/se
...
+
+{% if site.data.language-details.python-versions.size > 0 %}
+## Python versions
+
+These archives are available for on-demand installation.
+
+{: #python-versions-table}
+| Release | Arch | Name |
+| :------------- | :------------- | :------- |{% for file in site.data.language-details.python-versions %}
+| {{ file.release }} | {{ file.arch }} | {{ file.name }} |{% endfor %}
```
native-api: "Archives" are implementation detail and would only sidetrack users. Better specify just version IDs as they should go into `python:`.

PR #2537, opened 2019-09-29
Add procedure for travis-ci.com

native-api:
Multiple users have had problems encrypting for travis-ci.com using this guide,

The latest in https://travis-ci.community/t/script-cant-find-secure-environment-variable

### DIFF
--- a/user/environment-variables.md
+++ b/user/environment-variables.md
@@ -108,11 +108,15 @@ Encrypt environment variables with the public key attached to your repository us
 
 1. If you do not have the `travis` gem installed, run `gem install travis`.
 
-2. In your repository directory, run:
+2. In your repository directory:
 
-   ```bash
-   travis encrypt MY_SECRET_ENV=super_secret --add env.matrix
-   ```
+   * If you are using https://travis-ci.org, run:
+
+       ```bash
+       travis encrypt MY_SECRET_ENV=super_secret --add env.matrix
+       ```
+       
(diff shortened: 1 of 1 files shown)


native-api:
Test failure unrelated, fixed in https://github.com/travis-ci/docs-travis-ci-com/pull/2538

CJDennis:
The default case should be mentioned first. Since beginners (like me!) will have just created a new (likely free) account, and all free accounts are now created on the professional server like another other account is, the `--com`/`--pro` option should be mentioned first as the default, and the `--org` option for older, free accounts should be mentioned as an alternative.

user/environment-variables.md
```diff
@@ -108,11 +108,15 @@ Encrypt environment variables with the public key attached to your repository us
...
-   ```bash
-   travis encrypt MY_SECRET_ENV=super_secret --add env.matrix
-   ```
+   * If you are using https://travis-ci.org, run:
+
+       ```bash
+       travis encrypt MY_SECRET_ENV=super_secret --add env.global
+       ```
+       
+   * If you are using https://travis-ci.com, see [Encryption keys -- Usage](https://docs.travis-ci.com/user/encryption-keys#usage).
```
MariadeAnton: Can we put .com first, and instead of linking to the docs page, mention an example? 
i.e.
`travis encrypt --com SOMEVAR="secretvalue"`

Thank you!

native-api: The procedure for .com is more complex and takes an entire section on the linked page. Are you sure an example would be an adequate replacement? I foresee users asking why they are getting an error message in response to that command...

Pezi777:
Thank you so much @native-api

PR #2544, opened 2019-10-05
Document the special case with no matrix expansion

native-api:
https://travis-ci.community/t/config-with-a-python-build-and-a-matrix-only-runs-the-matrix/5309

### DIFF
--- a/user/build-matrix.md
+++ b/user/build-matrix.md
@@ -216,6 +216,43 @@ script: env $EXTRA_TESTS ./test.py $TEST_SUITE
 ```
 {: data-file=".travis.yml"}
 
+### Explicitly included jobs with only one element in the build matrix
+
+As a special case, if your build matrix has only one element _and_ you specified
+explicitly included jobs, matrix expansion is not done and the explicit jobs
+_completely_ define your build. For example:
+
+```yaml
+language: python
+python:
+  - '3.5'
+matrix:
+  include:
+    - env: EXTRA_TESTS=true
+# only defines one job with `python: 3.5` and `env: EXTRA_TESTS=true`
+```
+{: data-file=".travis.yml"}
+
+If you need the (sole) job from the matrix in such a case, too,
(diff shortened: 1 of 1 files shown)


user/build-matrix.md
```diff
@@ -216,6 +216,43 @@ script: env $EXTRA_TESTS ./test.py $TEST_SUITE
...
+_completely_ define your build. For example:
+
+```yaml
+language: python
+python:
+  - '3.5'
+matrix:
+  include:
+    - env: EXTRA_TESTS=true
+# only defines one job with `python: 3.5` and `env: EXTRA_TESTS=true`
```
MariadeAnton:
```suggestion
# defines one job with `python: 3.5` and `env: EXTRA_TESTS=true`
```

user/build-matrix.md
```diff
@@ -216,6 +216,43 @@ script: env $EXTRA_TESTS ./test.py $TEST_SUITE
 ```
 {: data-file=".travis.yml"}
 
+### Explicitly included jobs with only one element in the build matrix
+
+As a special case, if your build matrix has only one element _and_ you specified
```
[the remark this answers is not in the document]
native-api: Committed a similar simpler wording.

The documentation seems to use "explicitly included jobs" as the official term for the `matrix: include:` construct. So I try to stick to that constant phrase rather than inflect it -- to make the meaning clear without the need for long explanations each time.

user/build-matrix.md
native-api: "Only" is here for a reason. This example is a "typical mistake" example -- its intention, implied from identifiers, is contrary to the result. And the next example is the fix for it.
So the phrasing emphasises that this is a counterintuitive result, drawing the reader's attention to this fact.

I'm not feeling too strongly about this though so I'm okay if you feel there's no need going the extra mile helping the reader with narrative devices.

PR #2561, opened 2019-10-16
Julia; document major version selector

visr:
This seems to work fine, but is not documented:

https://github.com/travis-ci/travis-build/blob/ead50a6790a3e48c75c3459a017faf743c6246be/lib/travis/build/script/julia.rb#L188-L190

I was already using this since https://github.com/JuliaCI/Appveyor.jl does the same.

cc @ararslan

### DIFF
--- a/user/languages/julia.md
+++ b/user/languages/julia.md
@@ -37,6 +37,7 @@ julia:
 Acceptable formats are:
  - `nightly` will test against the latest [nightly build](https://julialang.org/downloads/nightlies.html)
 of Julia.
+ - `X` will test against the latest release for that major version.
  - `X.Y` will test against the latest release for that minor version.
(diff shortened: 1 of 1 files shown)


user/languages/julia.md
```diff
@@ -37,6 +37,7 @@ julia:
 Acceptable formats are:
  - `nightly` will test against the latest [nightly build](https://julialang.org/downloads/nightlies.html)
 of Julia.
+ - `X` will test against the latest release for that major version.
```
ararslan: Technically only true for 1.x releases; just specifying "0" won't do anything, for example.
```suggestion
 - `X` will test against the latest release for that major version. (Applies only to major versions 1 and later.)
```

PR #2571, opened 2019-10-22
Worker information update

michal-at-travisci:
Added lxd-arm64 description under Worker information image

### DIFF
--- a/user/reference/overview.md
+++ b/user/reference/overview.md
@@ -78,6 +78,7 @@ if it contains:
 * `gce` → the build ran in a virtual machine on Google Compute Engine.
 * `wjb` → the build ran on macOS.
 * `1803-containers` → the build ran on Windows.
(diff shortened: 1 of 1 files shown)


user/reference/overview.md
```diff
@@ -78,6 +78,7 @@ if it contains:
 * `gce` → the build ran in a virtual machine on Google Compute Engine.
 * `wjb` → the build ran on macOS.
 * `1803-containers` → the build ran on Windows.
+* `lxd-arm64` → the build ran within LXD container on Arm64-based infrastructure (currently delivered by Packet)
```
Pezi777:
```suggestion
* `lxd-arm64` → the build ran within an LXD container on Arm64-based infrastructure (currently delivered by Packet)
```

michal-at-travisci: Thanks @Pezi777 !

michal-at-travisci:
Mentioned only because we mention Google Compute Engine two lines above and
whole thing is about what infrastructure it's run in.

PR #2582, opened 2019-11-07
Complete the shift to the "jobs:" key for matrix

cclauss:
__Matrix__ is the concept but the yaml key should be __jobs:__.

Also, update some Python version numbers.

### DIFF
--- a/user/build-matrix.md
+++ b/user/build-matrix.md
@@ -21,7 +21,7 @@ There are two ways to specify multiple parallel jobs (what we call the build mat
   ```
   {: data-file=".travis.yml"}
 
-* specify the exact combination of configurations you want in `matrix.include`. For example, if not all of those combinations are interesting, you can specify just the combinations you want:
+* specify the exact combination of configurations you want in `job.include`. For example, if not all of those combinations are interesting, you can specify just the combinations you want:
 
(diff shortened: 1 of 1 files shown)


cclauss:
@MariadeAnton

user/build-matrix.md
```diff
@@ -21,7 +21,7 @@ There are two ways to specify multiple parallel jobs (what we call the build mat
   ```
   {: data-file=".travis.yml"}
 
-* specify the exact combination of configurations you want in `matrix.include`. For example, if not all of those combinations are interesting, you can specify just the combinations you want:
+* specify the exact combination of configurations you want in `job.include`. For example, if not all of those combinations are interesting, you can specify just the combinations you want:
```
MariadeAnton:
```suggestion
* specify the exact combination of configurations you want in `jobs.include`. For example, if not all of those combinations are interesting, you can specify just the combinations you want:
```

cclauss: `jobs` (plural) matches the yml syntax two lines below.

Pezi777: @cclauss could you please 'commit Maria's suggestion' - then I can merge the PR :-)

PR #2603, opened 2019-12-04
PHP7.4-support added

AlonaHlobina:


adaamz:
can you also update dist pages?
https://docs.travis-ci.com/user/reference/bionic/
https://docs.travis-ci.com/user/reference/xenial/

PR #2613, opened 2019-12-18
Update 'Building Ruby' docs with another fix for Bundler 2.0

connorshea:
I discovered this was necessary after upgrading to Bundler 2.1.0 yesterday,
only for CI to break when 2.1.1 came out today.

See also: https://github.com/faker-ruby/faker/pull/1867

The docs should maybe also emphasize that the specific version being installed should be explicitly set (e.g. `gem install bundler:2.1.1`) to prevent builds from breaking when new versions come out, but I didn't do that here.

### DIFF
--- a/user/languages/ruby.md
+++ b/user/languages/ruby.md
@@ -168,12 +168,12 @@ want to use Bundler 2.0, try one of the following solutions:
     {: data-file=".travis.yml"}
 
 
-* If you're using a version of Ruby before 2.6 and want to use Bundler 2.x,
-  you'll also want to make sure to upgrade to a newer version of RubyGems.
+* If you're using a version of Ruby lower than 2.6 and want to use Bundler 2.x,
+  make sure to upgrade to a newer version of RubyGems.
   On the default version of RubyGems shipped with older versions of Ruby, the
   Bundler version in `Gemfile.lock` must match exactly the version being used,
-  or it will error. This would cause when a new version of Bundler comes out,
-  but it's fixed in more recent RubyGems releaeses.
+  or it will error.
+  It's fixed in more recent RubyGems releases.
 
     ```yaml
     before_install:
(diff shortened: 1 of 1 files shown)


user/languages/ruby.md
```diff
@@ -167,6 +167,26 @@ want to use Bundler 2.0, try one of the following solutions:
     ```
     {: data-file=".travis.yml"}
 
+
+* If you're using a version of Ruby before 2.6 and want to use Bundler 2.x,
+  you'll also want to make sure to upgrade to a newer version of RubyGems.
```
Pezi777:
```suggestion
  make sure to upgrade to a newer version of RubyGems.
```

user/languages/ruby.md
```diff
@@ -167,6 +167,26 @@ want to use Bundler 2.0, try one of the following solutions:
...
+  On the default version of RubyGems shipped with older versions of Ruby, the
+  Bundler version in `Gemfile.lock` must match exactly the version being used,
+  or it will error. This would cause when a new version of Bundler comes out,
```
Pezi777:
```suggestion
  or it will error.
```

user/languages/ruby.md
```diff
@@ -167,6 +167,26 @@ want to use Bundler 2.0, try one of the following solutions:
...
+  but it's fixed in more recent RubyGems releaeses.
```
Pezi777:
```suggestion
  It's fixed in more recent RubyGems releases.
```

Pezi777:
thank you @connorshea

Sa1Beer777:
> I discovered this was necessary after upgrading to Bundler 2.1.0 yesterday,
> only for CI to break when 2.1.1 came out today.
> 
> See also: [faker-ruby/faker#1867](https://github.com/faker-ruby/faker/pull/1867)
> 
> The docs should maybe also emphasize that the specific version being installed should be explicitly set (e.g. `gem install bundler:2.1.1`) to prevent builds from breaking when new versions come out, but I didn't do that here.

PR #2614, opened 2019-12-19
Update pre-installed Ruby versions in Bionic

rmzelle:
Appears out of date. See my recent build:

https://travis-ci.org/citation-style-language/styles/builds/626802555#L160

image

Not sure how to easily check the default ruby version.


Pezi777:
thank you @rmzelle and @DamianSzymanski

PR #2622, opened 2019-12-19
Because you should never do mvn install on any CI

alecharp:
Maven `install` phase is *not* about gathering the project dependencies and plugins binaries to store them locally. 
It's about generating the binary of the project and storing it in the `localRepository`, sort of local proxy for Maven artifacts.

The current documented command is in fact recommending to build the project in the TravisCI installation step.
Then build it again in the normal build step.

CI systems should *never ever* perform `mvn install` or any equivalent for any build and project configuration tool.
So an update on the Gradle command might be required as well.

### DIFF
--- a/user/languages/java.md
+++ b/user/languages/java.md
@@ -48,13 +48,13 @@ language: java
 Before running the build, Travis CI installs dependencies:
 
 ```bash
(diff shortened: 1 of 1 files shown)


alecharp:
It would be great if I could understand why you are closing this..

PR #2633, opened 2020-01-02
Updated Windows/MSYS2 documentation

Sapd:
* Replaced $* with "$@"
$* does not parse commands with spaces correctly. E.g. `$mingw64 cmake -G"MSYS Makefiles" ..` will fail because of the space after MSYS.

It will parse as just $mingw64 cmake -G"MSYS

* ~~Commented out MAKE=mingw32-make
This is inconsistent with the mingw64 installation described, and will fail on tools like cmake. As it is highly specific, I commented it out.~~

### DIFF
--- a/user/reference/windows.md
+++ b/user/reference/windows.md
@@ -134,4 +134,4 @@ MSYS2 contains two noteworthy [subsystems](https://github.com/msys2/msys2/wiki/M
 
 A point of caution: the pre-installed "mingw" Chocolatey package should **not** be used within any MSYS2 subsystem. (In fact, the above snippet uninstalls the "mingw" Chocolatey package to be safe.) Note that the [MSYS2 wiki](https://github.com/msys2/msys2/wiki/MSYS2-introduction#path) says:
 
->Beware that mixing in programs from other MSYS2 installations, Cygwin installations, compiler toolchains or even various other programs is not supported and will probably break things in unexpected ways.
+>Be aware that mixing in programs from other MSYS2 installations, Cygwin installations, compiler toolchains or even various other programs is not supported and will probably break things in unexpected ways.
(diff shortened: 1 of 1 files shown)


tanzislam:
> - Commented out MAKE=mingw32-make
>This is inconsistent with the mingw64 installation described

`mingw32-make.exe`, despite the "32" in its name, is the only "make" program included in [the toolchain package targeting the mingw64 environment](https://packages.msys2.org/package/mingw-w64-x86_64-make?repo=mingw64).

>and will fail on tools like cmake.

I haven't tried it with cmake. Do you have an example (e.g. a failing build log) showing problems with that use case?

Sapd:
> I haven't tried it with cmake. Do you have an example (e.g. a failing build log) showing problems with that use case?

You're right, I tested it again locally and via travis, and it doesn't matter in the case of cmake. It probably errored out for some other reason (the old travis log just shows that the command returns 1; instead of an error message).

However the other change is still important.

tanzislam:
FWIW I also have https://github.com/msys2/MSYS2-packages/pull/1795 to make it a bit more robust, but it seems like a losing battle with [shell metacharacters in general](https://travis-ci.community/t/how-do-i-use-mingw-and-msys2-shells/1787/3).

user/reference/windows.md
```diff
@@ -134,4 +134,4 @@ MSYS2 contains two noteworthy [subsystems](https://github.com/msys2/msys2/wiki/M
 
 A point of caution: the pre-installed "mingw" Chocolatey package should **not** be used within any MSYS2 subsystem. (In fact, the above snippet uninstalls the "mingw" Chocolatey package to be safe.) Note that the [MSYS2 wiki](https://github.com/msys2/msys2/wiki/MSYS2-introduction#path) says:
 
->Beware that mixing in programs from other MSYS2 installations, Cygwin installations, compiler toolchains or even various other programs is not supported and will probably break things in unexpected ways.
\ No newline at end of file
+>Be aware that mixing in programs from other MSYS2 installations, Cygwin installations, compiler toolchains or even various other programs is not supported and will probably break things in unexpected ways.
```
Pezi777:
```suggestion
> Be aware that mixing in programs from other MSYS2 installations, Cygwin installations, compiler toolchains or even various other programs is not supported and will probably break things in unexpected ways.
```

PR #2638, opened 2020-01-08
Specify the regex flavor used

native-api:
Subj, to give an exact idea of what regex syntax is valid.


Pezi777:
thank you @native-api and @Montana

Montana:
Of course Petra! @Pezi777

PR #2653, opened 2020-01-27
Add clarity on image deletion

muizzk:


SuperSandro2000:
I would suggest to rather use ``docker image prune -f``

PR #2654, opened 2020-01-28
Linux and Windows ip adresses in one row

r-arek:


Montana:
Looks awesome! Thank you Arek.

PR #2663, opened 2020-02-05
fix quotes in expires example

svenfuchs:


BanzaiMan:
We need a disclaimer that this doesn't work on macOS, which uses BSD `date`.

BanzaiMan:
A way to do this with BSD is:
```
date -j -f "%s" $(( $(date -u +"%s") + 365*24*60*60 )) +"%Y-%m-%d %H:%M:%S -0000"
```

This is not compatible with the Linux `date`.

user/deployment-v2/providers/s3.md
```diff
@@ -222,10 +222,17 @@ deploy:
   provider: s3
   # ⋮
   cache_control: "public, max-age=31536000: *.jpg, *.gif, *.png, *.webp, *.css, *.js, *.ico"
-  expires: '$(date -u +"%Y-%m-%d %H:%M:%S -0000" --date "next Year"): *.{jpg,gif,png,webp,css,js,ico}'
+  expires: "$(date -u +'%Y-%m-%d %H:%M:%S -0000' --date 'next Year'): *.{jpg,gif,png,webp,css,js,ico}"
```
BanzaiMan: Maybe add an online comment `# for Linux` at the end of this line.

Issue #2675, opened 2020-02-16
dpl v2 Github pages provider commit_messages missing variables


adrianbroher:
The current documentation of dpl notes a `commit_message` key, which allows to set the commit message.  The given example shows some variables in the format `%{identifier}`, but does not list which variables are available for substitution.

Reading the dpl code there seems also the possibility to use environment variables, but the syntax is never described anywhere.

svenfuchs:
Thanks for pointing this out @adrianbroher. We haven't had introspection for interpolation variables on dpl v2 so far, but now we do: https://github.com/travis-ci/dpl/pull/1165

This should address your issue: https://github.com/travis-ci/docs-travis-ci-com/pull/2677. Pending a review on the PR i'll close your issue here. Please feel free to reopen it if needed.

PR #2677, opened 2020-02-17
Document interpolation variables on dpl v2 pages

svenfuchs:
Addresses https://github.com/travis-ci/docs-travis-ci-com/issues/2675

### DIFF
--- a/_includes/deploy/providers/codedeploy.md
+++ b/_includes/deploy/providers/codedeploy.md
@@ -76,4 +76,14 @@ The following variable are availabe for interpolation on `description`:
 * `revision_type`
 * `build_number`
 
+Interpolation uses the syntax `%{variable-name}`. For example,
+`"Current commit sha: %{git_sha}"` would result in a string with the
+current Git sha embedded.
+
+Furthermore, environment variables present in the current build
+environment can be used through standard Bash variable interpolation.
+For example: "Current build number: ${TRAVIS_BUILD_NUMBER}".
+See [here](/user/environment-variables/#default-environment-variables)
+for a list of default environment variables set.
+
 {% include deploy/secrets.md name="access_key_id" env_name="AWS_ACCESS_KEY_ID" %}
--- a/_includes/deploy/providers/codedeploy.md
+++ b/_includes/deploy/providers/codedeploy.md
@@ -18,7 +18,7 @@ deploy:
 
 ## Status
 
-Support for deployments to AWS Code Deploy is in **alpha**. Please see [Maturity Levels](/user/deployment-v2#maturity-levels) for details.
+Support for deployments to AWS Code Deploy is in **beta**. Please see [Maturity Levels](/user/deployment-v2#maturity-levels) for details.
 ## Known options
 
 Use the following options to further configure the deployment.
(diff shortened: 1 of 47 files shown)


adrianbroher:
Hi @svenfuchs, thanks for the prompt response.  That change helps tremendously.  However I would like to suggest two improvements:

```
The following variable are availabe for interpolation on `commit_message`:
```

Does this mean Ruby string interpolation?   The [dpl code states](https://github.com/travis-ci/dpl/blob/91e405cdf2b45ee692787d81d09cb887c2d2ae19/lib/dpl/helper/interpolate.rb#L7-L9)

```
# Variables can be contained in scripts, shell commands, and messages.
# They have the syntax `%{name}` or `%s` (or any other identifier supported
# by [Kernel#sprintf](https://ruby-doc.org/core-2.6.3/Kernel.html#method-i-format)).
```

So the placeholders are sprintf format specifiers and not string interpolation?  The code itself is over my head as I don't write Ruby so I can't answer that myself.

There is commit https://github.com/travis-ci/dpl/commit/ea6bd6633056abb3cfe31ea7504babfd7374b27d that allows the interpolation of environment variables.  It would be nice to document the correct syntax to access those.  From what I understand the syntax is:

```
# assuming `export FOO=bar`
commit_message: "Doing %{$FOO}"
# Would result in "Doing bar"
```

svenfuchs:
@adrianbroher yes, hmm, and no.

The inline documentation you link to is meant for internal use mostly. The syntax `%s` would not be available to users.

The syntax is `%{variable-name}` though, yes.

While `"Doing %{$FOO}"` should work for users as well, one can simply use Bash interpolation instead, i.e. `"Doing ${FOO}"` or just `"Doing $FOO"`. These will be replaced when the `dpl` Bash command is executed that Travis CI will embed to your build script when the `deploy` section is present in your build configuration.

You are right that this should be elaborated more in the documentation. I have added https://github.com/travis-ci/docs-travis-ci-com/pull/2677/commits/2e144b493e599b959fba5a13e3b3330ba63a096d which I think should address this.

Thanks again!

adrianbroher:
@svenfuchs Thanks for clarifying the syntax and updating the docs.

_includes/deploy/providers/codedeploy.md
```diff
@@ -52,4 +52,38 @@ For example, `access_key_id` can be given as
 
 * `AWS_ACCESS_KEY_ID=<access_key_id>` or 
 * `CODEDEPLOY_ACCESS_KEY_ID=<access_key_id>`
+## Interpolation variables
+
+The following variable are availabe for interpolation on `description`:
```
Pezi777:
```suggestion
The following variables are available for interpolation on `description`:
```

_includes/deploy/providers/flynn.md
```diff
@@ -0,0 +1,31 @@
...
+
+
+
+{{ include.content }}
+{% endunless %}
+
+## Status
+
+Support for deployments to Flynn is in **development**. Please see [Maturity Levels](/user/deployment-v2#maturity-levels) for details.
+## Known options
```
Pezi777:
```suggestion

## Known options
```

_includes/deploy/providers/lambda.md
```diff
@@ -59,4 +59,38 @@ For example, `access_key_id` can be given as
 
 * `AWS_ACCESS_KEY_ID=<access_key_id>` or 
 * `LAMBDA_ACCESS_KEY_ID=<access_key_id>`
+## Interpolation variables
+
+The following variable are availabe for interpolation on `description`:
```
Pezi777:
```suggestion
The following variables are available for interpolation on `description`:
```

lib/dpl/docs.rb
```diff
@@ -303,6 +308,44 @@ def env # TODO
       end
     end
 
+    class Vars < Struct.new(:cmd)
+      STR = <<~str
+        ## Interpolation variables
+
+        The following variable are availabe for interpolation on %s:
```
Pezi777:
```suggestion
        The following variables are available for interpolation on %s:
```

svenfuchs: fixed. thanks 👍

PR #2685, opened 2020-02-19
mention conditional allow_failures, talk about jobs, not rows

svenfuchs:


user/customizing-the-build.md
```diff
@@ -523,12 +524,27 @@ jobs:
 ```
 {: data-file=".travis.yml"}
 
+#### Conditionally Allowing Jobs to Fail
+
+Allowed failures can include a [condition](https://docs.travis-ci.com/user/conditional-builds-stages-jobs#conditionally-allowing-jobs-to-fail) using the key `if`.
```
BanzaiMan: You want a relative URL here, right?

user/customizing-the-build.md
```diff
@@ -523,12 +524,27 @@ jobs:
...
+
+For example, the following would allow the job using `rvm: 1.9.3` to fail
```
BanzaiMan: Nothing technical, but mentioning 1.9.3 in 2020 seems really outdated.

user/customizing-the-build.md
```diff
@@ -523,12 +524,27 @@ jobs:
...
+```
+{: data-file=".travis.yml"}
+
 #### Matching Jobs with `allow_failures`
 
 When matching jobs against the definitions given in `allow_failures`, _all_
-conditions in `allow_failures` must be met exactly, and
-all the keys in `allow_failures` element must exist in the
-top level of the build matrix (i.e., not in `matrix.include`).
+attributes specified on a row in `allow_failures` must be met exactly, and all
```
BanzaiMan: Here, "on a row" is probably incorrect. If there are more than one attribute for a condition to allow failures, it would span multiple rows:

```yaml
jobs:
  allow_failures:
  - os: linux
    rvm: 2.7.0
  - os: linux
    rvm: 2.6.2
```

svenfuchs: i think talking about rows is kinda misleading anyway, if we have the much more fitting term "job". i'll change that.

user/customizing-the-build.md
svenfuchs: true, but i don't think i want to change all the examples everywhere with this change :)

user/customizing-the-build.md
svenfuchs: whoops. yes. thanks!

user/customizing-the-build.md
```diff
@@ -523,12 +524,27 @@ jobs:
...
+For example, the following would allow the job using `rvm: 1.9.3` to fail
+only on the `master` branch:
+
+```yaml
+jobs:
+  allow_failures:
+  - rvm: 1.9.3
+    if: branch = master
+```
+{: data-file=".travis.yml"}
```
native-api: Since this is already documented elsewhere, is it a good idea to duplicate the information here? (You know, stuff getting desynced with time, users getting confused, everyone's time and nerves getting wasted.)

I'd go with only a reference: "A job [can be allowed to fail conditionally](https://docs.travis-ci.com/user/conditional-builds-stages-jobs#conditionally-allowing-jobs-to-fail)"

PR #2698, opened 2020-02-25
new file for multiple platform login created

Pezi777:


user/multi-platform-login.md
```diff
@@ -0,0 +1,42 @@
...
+2. Re-login on Bitbucket with the desired account.
+3. Go to the Travis CI Sign-in page.
+4. Click **Sign In with Bitbucket** on the sign-in page. Your account is stored and you can now switch between the different accounts in Travis CI.
+
+### Adding accounts using the user menu
+
+To add more accounts:
+
+1. Click on your user icon in the top right corner of [Travis CI](travis-ci.com).
+2. Click the **Add an account** link at the bottom of the menu. The Sign-in page appears displaying all your accounts.
```
AlonaHlobina: Should we also add here `re-login on Bitbucket with the desired account`?

Pezi777: I don't think we need to mention this here as it is in the Note right on top

PR #2700, opened 2020-02-26
Suggest better wording for 'inside a custom script'

shaleh:
The origin text

> Using exit inside a custom script which will be invoked from during a build is fine.

is confusing to read because it appears to either be missing words or to be the left over from a previous edit. "invoke from during" is not standard English usage.

This PR updates the text to indicate that using exit is safe and that it will have the expected result of marking the task as failed.


muizzk:
Thanks @shaleh

udeathstarck:
Deleted

PR #2721, opened 2020-03-12
fix anchor name

native-api:
In https://github.com/travis-ci/docs-travis-ci-com/pull/2720, I've mixed up the name of the added anchor so the intra-page link didn't work.

Also put the anchor above the linked note -- this way, the note looks better when following the link to the anchor.


native-api:
That said, the PR build has a check for broken links in documentation which would have caught this -- but it's currently disabled.

Montana:
Good catch

PR #2731, opened 2020-03-25
rewrite sole job in matrix

Pezi777:


user/build-matrix.md
```diff
@@ -280,6 +278,7 @@ python:
 jobs:
   include:
     - {}
+    -
```
svenfuchs: i think this line is not needed?

PR #2746, opened 2020-04-08
Use "jobs" in favor of "matrix" in Global Vars description

thomasplevy:
The yaml example uses the key "jobs" in favor "matrix" (I believe from older versions of the specification). This change updates the language in the description text to match what's found in the example.


thomasplevy:
@Pezi777 you're welcome!

PR #2749, opened 2020-04-14
Environment Variables: matrix -> jobs

adamjstewart:
I believe this is a typo, or leftover from some old syntax. The description says to use "global" and "matrix" keys, but the example `.travis.yml` uses "global" and "jobs" keys.


Montana:
> thank you @adamjstewart and @Montana

No problem @Pezi777  :)

PR #2754, opened 2020-04-20
S3 docs - Fixing YAML syntax

yogingale:


BanzaiMan:
I don't think `file_glob` is a recognized option for the S3 provider. (It is for GitHub Releases in [v1](https://docs.travis-ci.com/user/deployment/releases/#uploading-multiple-files) and [v2](https://docs.travis-ci.com/user/deployment-v2/providers/releases/#known-options)). Do you have evidence to suggest otherwise?

yogingale:
@BanzaiMan Yes, you are right. Turns out `glob: *.txt` is not a valid YAML. Changed it to `glob: "*.txt"`

PR #2764, opened 2020-04-29
Fix link to RoR `.travis.yml`

rjollos:
The file was removed in the [next commit](https://github.com/rails/rails/commit/ff10ac5fcb926fd491ccab09e898cd4268281beb).


BanzaiMan:
Might be better to find a better example than Rails at this point.

PR #2808, opened 2020-05-27
Add Xcode 11.5 GM info and remove Xcode 6.4 from docs

cesar-cs:


_data/xcodes.yml
```diff
@@ -837,20 +837,4 @@ osx_images:
...
-      - macosx10.10
-      - iphoneos8.4
-      - iphonesimulator8.4
-    simulators:
-      - "iOS 7.1"
-      - "iOS 8.1"
-      - "iOS 8.2"
-      - "iOS 8.4"
-    jdk: "1.8.0_45-b14"
+    jdk: "1.8.0_92-b14"
```
cclauss, line 840: 🚫 means the file does not end with one and only one /n.

PR #2819, opened 2020-06-10
Emphasize conditions for before/after steps to run

muizzk:
@BanzaiMan , it turns out this was already stated. Just added some emphasis.


BanzaiMan:
Looks good so far, but I was also thinking that https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/job-lifecycle.md should explicitly mention this.

muizzk:
@BanzaiMan updated Job Lifecycle as well. Thanks

PR #2836, opened 2020-07-08
Update triggering-builds.md

hughevans:
This doco (or your API) is currently broken and the only way to trigger builds (for me) is to hit `https://api.travis-ci.com/repo/{repo_id}/requests`

https://docs.travis-ci.com/user/triggering-builds/


Pezi777:
thank you @hughevans

Montana:
Good catch

PR #2845, opened 2020-07-14
Update master

travis-dpl-test:


michal-at-travisci:
@AlonaHlobina Can you check if this one should be merged already?

Montana:
@michal-at-travisci, @AlonaHlobina -- I merged this, will be deploying in Slack via ```deploys``` channel using fluxbot

PR #2847, opened 2020-07-16
Update Erlang/OTP release examples

michaelklishin:
No one is testing on OTP 17.5 or 18.x any more. It would
be more useful to have examples that use
Erlang 23.0.x and 22.3.x.


michaelklishin:
@nrios14 any feedback on this?

PR #2861, opened 2020-08-13
SonarCloud - Switch Maven execution to JRE v11

axel3rd:
Because using a configuration which can throw a *The version of Java (1.8.0_252) you have used to run this analysis is deprecated and we will stop accepting it from October 2020. Please update to at least Java 11.* warning would be not really reasonable 😁. 

Detail of end of support: https://sonarcloud.io/documentation/appendices/end-of-support/


axel3rd:
@bellingard (VP of Products at SonarSource) for approuval ?

Mr1262:
👍

PR #2862, opened 2020-08-14
Updating info for Xcode 11.6 GM

bigboos999:
****


AlonaHlobina (approved):
Thank you @cesar-cs for taking care of it

PR #2868, opened 2020-08-21
Add working example to install newer Elasticsearch version

schmengler:
The `chown` command is necessary at least on Xenial


nrios14 (approved):
@schmengler thanks for the update

PR #2899, opened 2020-10-07
Updating Xcode information

cesar-cs:
Thanks :)


Montana:
Looks good, @cesar-cs!

PR #2984, opened 2021-03-16
Update java.md

leevida:
Make documentation for switching JDKs 10 and above a little clearer


leevida:
For more context, I added some links for the tools that the documentation is referring to. And then corrected the sample script to something that will work

nrios14 (approved):

PR #3004, opened 2021-04-23
Update and rename user/tutorial.md to Winnipoker

anggashinichi:


michal-at-travisci (commented):
This is spam-change. Rejected.

michal-at-travisci:
Rejected.

PR #3013, opened 2021-05-10
Update linux

dgdghff:
Leaving out the odd.
The even  have a catch to them
Somewhere along the line.

### DIFF
--- a/user/reference/linux.md
+++ b/user/reference/linux.md
@@ -9,11 +9,11 @@ layout: en
 This page gives an overview of the different Ubuntu Linux distributions you can use as your CI environment.
 
 You can choose one of the following distributions:
-* [Ubuntu Focal 20.04](/user/reference/focal/)
+* [Ubuntu Focal19.01](/user/reference/focal/)
 * [Ubuntu Bionic 18.04](/user/reference/bionic/)
-* [Ubuntu Xenial 16.04](/user/reference/xenial/) **default**
+* [Ubuntu Xenial 17.01](/user/reference/xenial/) default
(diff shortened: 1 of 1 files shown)


user/reference/linux.md
```diff
@@ -9,19 +9,19 @@ layout: en
 This page gives an overview of the different Ubuntu Linux distributions you can use as your CI environment.
 
 You can choose one of the following distributions:
-* [Ubuntu Focal 20.04](/user/reference/focal/)
+* [Ubuntu Focal19.01](/user/reference/focal/)
```
nrios14, line 12: Focal 19.01 (there is an extra space)

nrios14 (approved):

user/reference/linux.md
```diff
@@ -9,19 +9,19 @@ layout: en
...
+* [Ubuntu Precise 13.03](/user/reference/precise/)
 
 ## Using Ubuntu Linux distributions
 
 To use our Ubuntu Linux build infrastructure, you can choose between the distributions above.
 
 ## Default 
 
-We use Ubuntu Xenial 16.04 as default. You find more about packages, tools and settings in [Ubuntu Xenial 16.04](/user/reference/xenial/).
+We use Ubuntu Xenial 17.01 as default. You find more about packages, tools and settings in [Ubuntu Xenial 16.04](/user/reference/xenial/).
```
SIMULATAN, line 24: shouldn't the link also be titled 17.01 instead of 16.04?

SIMULATAN (changes requested):
There might still be a wrong version number, see my comment.

PR #3016, opened 2021-05-19
Remove mention of stage name capitalization

gregnavis:
As of now, the build stage name is NOT in capitalized form but instead is copied verbatim from `stage`.


nrios14:
@Montana  is this correct? should I merge this?

nrios14 (approved):

PR #3023, opened 2021-05-31
Fix

mustafa-travisci:
should be yml instead yaml.


Montana:
Yeah, obviously agreed with @mustafa-travisci. `.yml` instead of `yaml`. Other than that looks good.

Montana (approved):

PR #3062, opened 2021-07-27
Update triggering-builds.md

nrios14:
removed headline note, since no longer applies


Ao-2 (commented):
> `[pjgpetecodes:add-pete-codes-remote-debug]() `

michal-at-travisci (approved):

PR #3128, opened 2021-11-17
Add 'medium' VM size to VM Instance Size.

qasim-at-tci:
Please review.
Thanks.

### DIFF
--- a/user/reference/overview.md
+++ b/user/reference/overview.md
@@ -240,7 +240,7 @@ If you need more than the default 2vCPU and RAM, you can run a build on a larger
 
 ```yaml
 vm:
-  size: [large|x-large|2x-large]
+  size: [medium|large|x-large|2x-large]
 ```
 {: data-file=".travis.yml"}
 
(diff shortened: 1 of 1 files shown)


arturgrigorenko201:
gh pr checkout 3128

arturgrigorenko201 (approved):

user/reference/overview.md
```diff
@@ -240,7 +240,7 @@ If you need more than the default 2vCPU and RAM, you can run a build on a larger
 
 ```yaml
 vm:
-  size: [large|x-large|2x-large]
+  size: [medium|large|x-large|2x-large]
```
michal-at-travisci, line 243: actually if you use no `size` tag, the medium/standard is assumed by default and is not necessary to be explicitly called out (unless you want a full visibility in mixed size environment). Maybe there should be a note added about `medium` being assumed by default and thus not requiring explicit `sie: medium`?

user/reference/overview.md
```diff
@@ -257,9 +257,10 @@ Each tier of instance size delivers more vCPU and RAM resources available at you
 
 | size      | vCPU        | Memory GiB  | Comment                                                 |
 |:---------:|:-----------:|:-----------:|:-------------------------------------------------------:|
+| medium    | 2           | ~8          | default                                                 |
```
michal-at-travisci, line 260: if you're adding `medium` here as default, please extend comment: from `default` to `default and assumed automatically if no size: medium present in the .travis.yml` or something along those lines

michal-at-travisci (changes requested):
@qasim-at-tci  if we're adding the default assumed size into documentation, we need to be very explicit, that it doesn't have to be called out explicitly and is assumed automatically if size tag is not used in .travis.yml

PR #3165, opened 2022-02-08
fix: broken login link in OSS migration page

agilgur5:
## Summary

Found this glaring rendering issue while migrating my OSS repos to `.com` from `.org`:

Screen Shot 2022-02-08 at 6 45 45 PM

 Figured I'd fix it and other issues I found along the way (e.g. #3164 )

## Description

- [https://travis-ci.com] was rendering as is, with brackets and
  no link
  - see screenshot above
  - use [travis-ci.com] instead which works and is already used further
    up in this document


nrios14:
@agilgur5 
Thank you for fixing the link!

PR #3218, opened 2022-09-14
Hashicorop vault integration docs

nrios14:
Please review new page and new updated section


michal-at-travisci (changes requested):
@nrios14 can you add left hand menu entry as per original design doc, in integrations section? It may be a separate PR, as I am inclined to approve that one

michal-at-travisci:
Ok I added Left hand menu entry, approving

michal-at-travisci (approved):

PR #3236, opened 2022-11-14
Adding Xcode 14.1 info

cesar-cs:


Montana (approved):
Looks great @cesar-cs!

PR #3270, opened 2023-02-08
Document a known issue in the image

jbosse:
The documentation says that postgres 12 is installed by default and you just need to activate it within `services:` but that is insufficient to get it working.


michal-at-travisci:
@cesar-cs @smokodon please verify and either accept change to docs or reject if it's not valid anymore

smokodon:
Thanks @jbosse, this was indeed the case with Focal, it had mistakenly preinstalled version 13 instead of 12 of PostgreSQL - it's fixed now and it works as described in docs

PR #3322, opened 2023-06-30
Update private-dependencies.md

LordSloppy:
Added ### after comment on forks


michal-at-travisci:
Thank you, but I don't see the purpose for this change.

PR #3365, opened 2024-04-22
PR update gtag - update in `en.html` since it's Ruby the inheritance should be for all pages.

Montana:
update gtag in `en.html` so inheritance is for all pages.


Montana:
Yup, agreed.

PR #3388, opened 2024-08-07
Update user-role-management.md

nrios14:


stan-travis (approved):
Hope it is the reason of broken table.

PR #3465, opened 2025-03-17
Du 35/remove mac

nrios14:
Adding a notice to mention that macOS will be deprecated.


IderaJohn:
This is updating the docs to say we're no longer supporting macOS as of
March 31st....

Will we be removing the Mac specific portions after then, or leaving it
around in case we re-enable it?

Cheers
John

IderaJohn (approved):
All looks great!

PR #3468, opened 2025-03-26
Update open_source_license.md

nrios14:
Updating page according to Assembla ticket EWS-36, which requires a license update for FOSS - OpenSSH License. 

The license was added.


IderaJohn:
Can you remove the old version of the OpenSSH license please? I think you could have just updated the the version number...

nrios14:
no longer needed.

